PR #3, opened 2017-12-22
Add codes

TeXitoi:
On #2

### DIFF
--- a/src/ntfs.rs
+++ b/src/ntfs.rs
@@ -150,7 +150,7 @@ where
         Some(idx) => idx,
         None => {
             eprintln!(
-                "object_codes.txt: object {} {} not found",
+                "object_codes.txt: object_type={} object_id={} not found",
                 code.object_type, code.object_id
             );
             return;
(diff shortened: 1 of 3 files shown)


src/ntfs.rs
```diff
@@ -124,6 +126,55 @@ fn manage_stop_times(collections: &mut Collections, path: &path::Path) {
...
+}
+fn insert_code<T>(collection: &mut Collection<T>, code: Code)
+where
+    T: Codes + Id<T>,
+{
+    let idx = match collection.get_idx(&code.object_id) {
+        Some(idx) => idx,
+        None => {
+            eprintln!(
+                "object_codes.txt: object {} {} not found",
```
pbougue: maybe a `type:` and `id:` could help understand a bit mode

src/objects.rs
```diff
@@ -1,5 +1,24 @@
 use collection::{Id, Idx};
 
+pub type CodesT = Vec<(String, String)>;
```
pbougue: Could be nice to have a comment on the choice of the structure, so someone smart in the future doesn't do the same mistakes/tests with other structures

PR #6, opened 2017-12-22
Added more attributes in struct Line

datanel:


src/objects.rs
```diff
@@ -71,14 +71,67 @@ impl Id<Network> for Network {
...
+
+impl<'de> ::serde::Deserialize<'de> for Rgb {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: ::serde::Deserializer<'de>,
+    {
+        use serde::de::Error;
+
+        let color_hex = String::deserialize(deserializer)?;
+        let color_dec = u32::from_str_radix(&color_hex, 16).map_err(Error::custom)?;
```
TeXitoi: Just add a check that the string length is 6 to not wait for FFF to be white.

src/objects.rs
```diff
@@ -71,14 +71,67 @@ impl Id<Network> for Network {
...
+    #[serde(rename = "forward_line_name")] pub forward_name: Option<String>,
+    pub forward_direction: Option<String>,
+    #[serde(rename = "backward_line_name")] pub backward_name: Option<String>,
+    #[serde(rename = "line_color")] pub color: Option<Rgb>,
+    #[serde(rename = "line_text_color")] pub text_color: Option<Rgb>,
+    #[serde(rename = "line_sort_order")] pub sort_order: Option<u32>,
     pub network_id: String,
     pub commercial_mode_id: String,
+    pub geometry_id: Option<String>,
+    //TODO Type Datetime?
```
TeXitoi: Yeah, but we'll do that later when reading stop times. Something using a u32 for the number of seconds. It'll need to be compact for memory considerations.

src/objects.rs
datanel: added

PR #9, opened 2017-12-29
use RAII to check id integrity instead of passing closure

TeXitoi:


src/ntfs.rs
```diff
@@ -138,9 +140,10 @@ fn insert_code_with_idx<T>(collection: &mut Collection<T>, idx: Idx<T>, code: Co
 where
     T: Codes + Id<T>,
 {
-    collection.mut_elt(idx, move |obj| {
-        obj.codes_mut().push((code.object_system, code.object_code));
-    });
+    collection
+        .index_mut(idx)
+        .codes_mut()
+        .push((code.object_system, code.object_code));
```
datanel: :+1:  cleaner

PR #15, opened 2018-01-04
Added FeedInfo, Calendar and more atributes to Network

datanel:


src/ntfs.rs
```diff
@@ -205,6 +214,10 @@ pub fn read<P: AsRef<path::Path>>(path: P) -> PtObjects {
     collections.routes = make_collection(path, "routes.txt");
     collections.vehicle_journeys = make_collection(path, "trips.txt");
     collections.physical_modes = make_collection(path, "physical_modes.txt");
+    if path.join("feed_infos.txt").exists() {
+        collections.feed_infos = make_feed_infos(path, "feed_infos.txt");
+    }
```
TeXitoi: As feed_infos.txt is now needed, you can remove the check.

src/objects.rs
```diff
@@ -340,6 +345,32 @@ impl Id<StopArea> for StopPoint {
 }
 impl_codes!(StopPoint);
 
+#[derive(Serialize, Deserialize, Debug)]
+pub struct FeedInfo {
+    #[serde(rename = "feed_info_param")] pub info_param: String,
+    #[serde(rename = "feed_info_value")] pub info_value: String,
+}
```
TeXitoi: No pub as it's only used to decode the CSV. Thus please put it private in ntfs.rs.

src/ntfs.rs
```diff
@@ -21,20 +23,27 @@ fn default_visible() -> bool {
...
+            acc
+        })
 }
 
 #[derive(Serialize, Deserialize, Debug, Clone)]
 struct Stop {
     #[serde(rename = "stop_id")] id: String,
     #[serde(rename = "stop_name")] name: String,
-    #[serde(default = "default_visible", deserialize_with = "de_from_i32")] visible: bool,
+    #[serde(default = "default_visible", deserialize_with = "de_from_u8")] visible: bool,
```
TeXitoi: you can add serialization

src/ntfs.rs
```diff
@@ -107,7 +116,7 @@ struct StopTime {
     #[serde(default)] alighting_duration: u16,
     #[serde(default)] pickup_type: u8,
     #[serde(default)] dropoff_type: u8,
-    #[serde(default, deserialize_with = "de_from_i32")] datetime_estimated: bool,
+    #[serde(default, deserialize_with = "de_from_u8")] datetime_estimated: bool,
```
TeXitoi: you can add serialization

src/objects.rs
```diff
@@ -340,6 +345,32 @@ impl Id<StopArea> for StopPoint {
...
+    #[serde(rename = "calendar_id")] pub id: String,
+    #[serde(deserialize_with = "de_from_u8", serialize_with = "ser_from_bool")] pub monday: bool,
+    #[serde(deserialize_with = "de_from_u8", serialize_with = "ser_from_bool")] pub tuesday: bool,
+    #[serde(deserialize_with = "de_from_u8", serialize_with = "ser_from_bool")] pub wednesday: bool,
+    #[serde(deserialize_with = "de_from_u8", serialize_with = "ser_from_bool")] pub thursday: bool,
+    #[serde(deserialize_with = "de_from_u8", serialize_with = "ser_from_bool")] pub friday: bool,
+    #[serde(deserialize_with = "de_from_u8", serialize_with = "ser_from_bool")] pub saturday: bool,
+    #[serde(deserialize_with = "de_from_u8", serialize_with = "ser_from_bool")] pub sunday: bool,
+    pub start_date: String,
+    pub end_date: String,
```
TeXitoi: We'll need to type that.

src/objects.rs
datanel: Are you sure? I need to read the properties in ntfs.rs

src/objects.rs
datanel: this can be done in another PR?

datanel: I wrote an issue https://github.com/CanalTP/navitia_model/issues/17

src/objects.rs
TeXitoi: If you move it to ntfs.rs, pub is not needed, yes.

PR #16, opened 2018-01-04
Gtfs reader - reading agency for networks

prhod:


examples/gtfs_reader.rs
```diff
@@ -0,0 +1,7 @@
+extern crate csv;
+extern crate navitia_model;
+
+fn main() {
+    let objects = navitia_model::gtfs::read("./examples/data/");
+    println!("Count of networks loaded : {}", objects.networks.len());
```
TeXitoi: you can dump the model using serde_json

examples/gtfs_reader.rs
```diff
@@ -0,0 +1,7 @@
+extern crate csv;
+extern crate navitia_model;
+
+fn main() {
+    let objects = navitia_model::gtfs::read("./examples/data/");
```
TeXitoi: better to read current directory

src/gtfs.rs
```diff
@@ -0,0 +1,53 @@
...
+    "1".to_string()
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone)]
+struct Agency {
+    #[serde(rename = "agency_id", default = "default_agency_id")] id: String,
+    #[serde(rename = "agency_name")] name: String,
+    #[serde(rename = "agency_url")] url: String,
+    #[serde(rename = "agency_timezone")] timezone: Option<String>,
+    #[serde(rename = "agency_lang", default)] lang: Option<String>,
```
TeXitoi: I don't think you need default for option fields.

src/gtfs.rs
```diff
@@ -0,0 +1,53 @@
...
+    #[serde(rename = "agency_phone", default)] phone: Option<String>,
+    #[serde(rename = "agency_fare_url", default)] fare_url: Option<String>,
```
TeXitoi: If there is field you don't use, just don't put them.

tests/gtfs_reader.rs
```diff
@@ -0,0 +1,35 @@
...
+use std::io::prelude::*;
+
+#[test]
+fn load_minimal_agency() {
+    let agency_content = "agency_name,agency_url,agency_timezone\n
+    My agency,http://my-agency_url.com,Europe/London";
+    let tmp_dir = TempDir::new("osm_transit_extractor").expect("create temp dir");
+    let file_path = tmp_dir.path().join("agency.txt");
+    let mut f = File::create(&file_path).unwrap();
+    f.write_all(agency_content.as_bytes()).unwrap();
```
TeXitoi: better to add the files in the repo and read them.

tests/gtfs_reader.rs
```diff
@@ -0,0 +1,35 @@
...
+use navitia_model::gtfs;
+use tempdir::TempDir;
+use std::fs::File;
+use std::io::prelude::*;
+
+#[test]
+fn load_minimal_agency() {
+    let agency_content = "agency_name,agency_url,agency_timezone\n
+    My agency,http://my-agency_url.com,Europe/London";
+    let tmp_dir = TempDir::new("osm_transit_extractor").expect("create temp dir");
```
TeXitoi: That's not osm_transit_extractor ;-)

prhod: dammit, I've been spotted :p

tests/gtfs_reader.rs
prhod: I started with a fixture, but it led to an awful list of subdirectories. I thought it best to create the file one the fly for those simple tests.

examples/gtfs_reader.rs
prhod: I don't understand what you want me to do here. Print out a JSON output for the first network ?

TeXitoi: That's for you. I don't really care. You can easilly dump the model as json, interesting for testing.

prhod: ok, good idea. I'll try to use it in the tests. If you're ok with it, I'll let the example like that for the moment.

TeXitoi: In test, that's easier to directly check the structure.

As you want, that's just for you ;-)

tests/gtfs_reader.rs
```diff
@@ -0,0 +1,49 @@
...
+    f.write_all(agency_content.as_bytes()).unwrap();
+
+    let networks = gtfs::read_agency(tmp_dir.path());
+    tmp_dir.close().expect("delete temp dir");
+    assert_eq!(1, networks.len());
+}
+
+#[test]
+fn load_complete_agency() {
+    let agency_content = "agency_id,agency_name,agency_url,agency_timezone,agency_lang,agency_phone,agency_fare_url,agency_email\n
```
pbougue: Not sure (and for knowledge only), but you might use https://doc.rust-lang.org/reference/tokens.html#raw-string-literals

PR #18, opened 2018-01-04
Added calendar dates

datanel:
- [x]  TODO: use https://github.com/CanalTP/navitia_model/pull/9 if merged before this PR

### DIFF
--- a/src/ntfs.rs
+++ b/src/ntfs.rs
@@ -199,10 +199,7 @@ struct CalendarDate {
     exception_type: ExceptionType,
 }
 
-fn insert_calendar_date<T>(collection: &mut Collection<T>, calendar_date: CalendarDate)
-where
-    T: CalendarDates + Id<T>,
-{
+fn insert_calendar_date(collection: &mut Collection<Calendar>, calendar_date: CalendarDate) {
     let idx = match collection.get_idx(&calendar_date.calendar_id) {
         Some(idx) => idx,
         None => {
@@ -213,10 +210,10 @@ where
             return;
         }
     };
-    collection.mut_elt(idx, move |obj| {
-        obj.calendar_dates_mut()
-            .push((calendar_date.date, calendar_date.exception_type));
-    });
+    collection
+        .index_mut(idx)
+        .calendar_dates
+        .push((calendar_date.date, calendar_date.exception_type))
 }
 
 fn manage_calendars(collections: &mut Collections, path: &path::Path) {
--- a/src/ntfs.rs
+++ b/src/ntfs.rs
@@ -193,6 +193,46 @@ fn manage_codes(collections: &mut Collections, path: &path::Path) {
 }
 
 #[derive(Serialize, Deserialize, Debug)]
+struct CalendarDate {
+    calendar_id: String,
+    date: Date,
+    exception_type: ExceptionType,
+}
+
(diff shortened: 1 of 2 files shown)


src/objects.rs
```diff
@@ -345,6 +345,21 @@ impl Id<StopArea> for StopPoint {
 }
 impl_codes!(StopPoint);
 
+pub type Date = String;
+
+#[derive(Serialize, Deserialize, Debug)]
+pub enum ExceptionType {
+    #[serde(rename = "1")] Add,
+    #[serde(rename = "2")] Remove,
```
TeXitoi: 👍

src/objects.rs
```diff
@@ -345,6 +345,21 @@ impl Id<StopArea> for StopPoint {
...
+}
+
+pub type CalendarDatesT = Vec<(Date, ExceptionType)>;
+
+pub trait CalendarDates {
+    fn calendar_dates(&self) -> &CalendarDatesT;
+    fn calendar_dates_mut(&mut self) -> &mut CalendarDatesT;
+}
```
TeXitoi: You don't need this trait machinery. It is needed for codes because several objects have codes, but for callendar_dates, only calendars will be impacted.

Thus you can just have `pub type CalendarDates = Vec<(Date, ExceptionType)>;` and then just directly take a Calendar object.

src/ntfs.rs
```diff
@@ -189,6 +189,46 @@ fn manage_codes(collections: &mut Collections, path: &path::Path) {
...
+#[derive(Serialize, Deserialize, Debug)]
+struct CalendarDate {
+    calendar_id: String,
+    date: Date,
+    exception_type: ExceptionType,
+}
+
+fn insert_calendar_date<T>(collection: &mut Collection<T>, calendar_date: CalendarDate)
+where
+    T: CalendarDates + Id<T>,
```
TeXitoi: This generic is not needed, just take `Collection<Calendar>`.

src/objects.rs
datanel: i'm a robot :)

src/objects.rs
```diff
@@ -355,8 +365,15 @@ pub struct Calendar {
...
-    pub end_date: String,
+    pub start_date: Date,
+    pub end_date: Date,
+    #[serde(skip)] pub calendar_dates: CalendarDatesT,
+}
+
+impl Calendar {
+    pub fn calendar_dates_mut(&mut self) -> &mut CalendarDatesT {
+        &mut self.calendar_dates
+    }
```
TeXitoi: You don't really need this method as the field is public.

src/objects.rs
```diff
@@ -345,6 +345,16 @@ impl Id<StopArea> for StopPoint {
...
 
+pub type Date = String;
+
+#[derive(Serialize, Deserialize, Debug)]
+pub enum ExceptionType {
+    #[serde(rename = "1")] Add,
+    #[serde(rename = "2")] Remove,
+}
+
+pub type CalendarDatesT = Vec<(Date, ExceptionType)>;
```
TeXitoi: You can remove the final T

datanel: T is not for Type?

TeXitoi: That was because it was clashing with the trait name.

PR #20, opened 2018-01-05
gtfs_reader - add tests on agencies and fix agency_id default

prhod:


src/gtfs.rs
```diff
@@ -20,8 +20,13 @@ struct Agency {
 }
 impl From<Agency> for objects::Network {
     fn from(agency: Agency) -> objects::Network {
+        let agency_id: String;
+        match agency.id {
+            None => agency_id = default_agency_id(),
+            Some(id) => agency_id = id,
+        }
```
TeXitoi:
```rust
let agency_id = match agency.id {
    None => default_agency_id(),
    Some(id) => id,
};
```

tests/gtfs_reader.rs
```diff
@@ -18,6 +19,8 @@ fn load_minimal_agency() {
     let networks = gtfs::read_agency(tmp_dir.path());
     tmp_dir.close().expect("delete temp dir");
     assert_eq!(1, networks.len());
+    let agency = &networks.into_vec()[0];
```
TeXitoi:
```rust
let agency = networks.iter().next().unwrap().1;
```
Minor for this case, but if you want to do something else with networks, `into_vec` will be problematic.

tests/gtfs_reader.rs
```diff
@@ -46,4 +49,20 @@ id_1,My agency,http://my-agency_url.com,Europe/London,EN,0123456789,http://my-ag
...
+fn load_2_agencies_with_no_id() {
+    let agency_content = "agency_name,agency_url,agency_timezone\n
+My agency 1,http://my-agency_url.com,Europe/London
+My agency 2,http://my-agency_url.com,Europe/London";
+    let tmp_dir = TempDir::new("navitia_model_tests").expect("create temp dir");
+    let file_path = tmp_dir.path().join("agency.txt");
+    let mut f = File::create(&file_path).unwrap();
+    f.write_all(agency_content.as_bytes()).unwrap();
+    gtfs::read_agency(tmp_dir.path());
+    tmp_dir.close().expect("delete temp dir");
```
TeXitoi: 👍

src/gtfs.rs
```diff
@@ -20,8 +20,13 @@ struct Agency {
...
         objects::Network {
-            id: agency.id,
+            id: agency_id,
```
datanel: I think this could work
```rust
id: agency.id.unwrap_or(default_agency_id())
```

TeXitoi:
```rust
id: agency.id.unwrap_or_else(default_agency_id),
```

tests/gtfs_reader.rs
```diff
@@ -1,4 +1,5 @@
 extern crate navitia_model;
+extern crate serde_json;
```
TeXitoi: I don't think that's needed.

PR #21, opened 2018-01-05
Added company and completed some models

datanel:
- Added Company (related to https://github.com/CanalTP/navitia/pull/2282)
- Updated PhysicalMode
- Updated Route
- Added NaiveDate for dates fix https://github.com/CanalTP/navitia_model/issues/17

### DIFF
--- a/src/ntfs.rs
+++ b/src/ntfs.rs
@@ -7,7 +7,6 @@ use objects::*;
 use collection::{Collection, Id, Idx};
 use {Collections, PtObjects};
 use utils::*;
-use chrono::NaiveDate;
 
 fn make_collection<T>(path: &path::Path, file: &str) -> Collection<T>
 where
@@ -197,7 +196,7 @@ fn manage_codes(collections: &mut Collections, path: &path::Path) {
 struct CalendarDate {
     calendar_id: String,
     #[serde(deserialize_with = "de_from_date_string", serialize_with = "ser_from_naive_date")]
-    date: NaiveDate,
+    date: Date,
     exception_type: ExceptionType,
 }
 
--- a/src/ntfs.rs
+++ b/src/ntfs.rs
@@ -265,6 +265,7 @@ pub fn read<P: AsRef<path::Path>>(path: P) -> PtObjects {
     collections.vehicle_journeys = make_collection(path, "trips.txt");
     collections.physical_modes = make_collection(path, "physical_modes.txt");
     manage_calendars(&mut collections, path);
+    collections.companies = make_collection(path, "companies.txt");
     manage_feed_infos(&mut collections, path);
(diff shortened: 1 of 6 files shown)


src/ntfs.rs
```diff
@@ -105,6 +105,8 @@ struct StopTime {
     #[serde(default, deserialize_with = "de_from_u8", serialize_with = "ser_from_bool")]
     datetime_estimated: bool,
     local_zone_id: Option<u16>,
+    headsign: Option<String>,
+    stop_time_id: Option<String>,
```
TeXitoi: StopTime is the object that use the RAM. Thus, what's the impact of this change on fr-idf?

We may need to manage these data from outside (with something like a `stop_time_ids: HashMap<(Idx<VehicleJourney>, u32), String>`) as these fields are rarely used.

datanel: without stop_time_id and headsign
1.03 GB

without stop_time_id
1.40 GB

with stop_time_id and headsign
2.17 GB :open_mouth:

I remove these 2 lines, I will do something better as you suggest in another PR

TeXitoi: That's almost always None, thus an external storage will be less costy

prhod: It's not very used for the moment, but I think it will be more and more used in the future.

src/objects.rs
```diff
@@ -376,6 +380,25 @@ impl Id<Calendar> for Calendar {
...
+pub struct Company {
+    #[serde(rename = "company_id")] pub id: String,
+    #[serde(rename = "company_name")] pub name: String,
+    #[serde(rename = "company_address_name")] pub address_name: Option<String>,
+    #[serde(rename = "company_address_number")] pub address_number: Option<String>,
+    #[serde(rename = "company_address_type")] pub address_type: Option<String>,
+    #[serde(rename = "company_url")] pub url: Option<String>,
+    #[serde(rename = "company_mail")] pub mail: Option<String>,
+    #[serde(rename = "company_phone")] pub phone: Option<String>,
+    #[serde(rename = "company_fax")] pub fax: Option<String>,
```
TeXitoi: fax? really? :-P

prhod: removed in the navitia PR on NTFS

src/objects.rs
```diff
@@ -345,15 +351,13 @@ impl Id<StopArea> for StopPoint {
 }
 impl_codes!(StopPoint);
 
-pub type Date = String;
```
TeXitoi: Better to keep this Date type and use it everywhere, it's simpler if we want to change.

Thus, something like `pub type Date = chrono::NaiveDate;`

datanel: fixed

PR #22, opened 2018-01-09
Added Comment and CommentLinks

datanel:


src/utils.rs
```diff
@@ -34,3 +34,18 @@ where
...
+{
+    use serde::Deserialize;
+    let s = String::deserialize(deserializer)?;
+    if s.is_empty() {
+        return Ok(Default::default());
+    }
+
+    ::serde_json::from_value(::serde_json::value::Value::String(s))
+        .map_err(::serde::de::Error::custom)
+}
```
TeXitoi: Inspired by https://github.com/BurntSushi/rust-csv/blob/master/src/lib.rs#L347

```rust
pub fn de_with_empty_default<'de, T: Default, D>(de: D) -> Result<T, D::Error>
where
    D: ::serde::Deserializer<'de>,
    for<'d> T: serde::Deserialize<'d>,
{
    use serde::Deserialize;
    Option::<T>::deserialize(de).map(|opt| opt.unwrap_or_else(Default::default))
}
```

Cargo.toml
```diff
@@ -14,7 +14,7 @@ log = "0.4"
 get-corresponding-derive = { path = "get-corresponding-derive" }
 tempdir = "0.3.5"
 chrono = "0.4.0"
+serde_json = "1"
```
TeXitoi: You can revert that with the new de_with_empty_default

src/lib.rs
```diff
@@ -8,6 +8,7 @@ extern crate log;
 extern crate serde;
 #[macro_use]
 extern crate serde_derive;
+extern crate serde_json;
```
TeXitoi: You can revert that too

src/ntfs.rs
```diff
@@ -253,6 +256,60 @@ fn manage_feed_infos(collections: &mut Collections, path: &path::Path) {
...
+    collection
+        .index_mut(idx)
+        .comment_links_mut()
+        .push(comment_link.comment_id);
+}
+
+fn manage_comments(collections: &mut Collections, path: &path::Path) {
+    if csv::Reader::from_path(path.join("comments.txt")).is_ok() {
+        info!("Reading comments.txt");
+        collections.comments = make_collection(path, "comments.txt");
```
TeXitoi: better to have an optional version of make_collection, that try to create a Read, and then, if successful, create a csv::Reader::from_reader. Creating a csv::Reader for cheecking the presence of the file is a bit overkill.

src/ntfs.rs
```diff
@@ -253,6 +256,60 @@ fn manage_feed_infos(collections: &mut Collections, path: &path::Path) {
...
+                let comment_link: CommentLink = comment_link;
+                match comment_link.object_type.as_str() {
+                    "stop_area" => insert_comment_link(&mut collections.stop_areas, comment_link),
+                    "stop_point" => insert_comment_link(&mut collections.stop_points, comment_link),
+                    "line" => insert_comment_link(&mut collections.lines, comment_link),
+                    "route" => insert_comment_link(&mut collections.routes, comment_link),
+                    "trip" => insert_comment_link(&mut collections.vehicle_journeys, comment_link),
+                    "stop_time" => {
+                        insert_comment_link(&mut collections.vehicle_journeys, comment_link)
+                    }
```
TeXitoi: stop_time should be on a stop_time, not a vehicle journey. You'll have id mismatch like that.

src/objects.rs
```diff
@@ -306,7 +331,9 @@ pub struct StopTime {
     pub dropoff_type: u8,
     pub datetime_estimated: bool,
     pub local_zone_id: Option<u16>,
+    pub comment_links: CommentLinksT,
```
TeXitoi: Still RAM problem with that.

src/objects.rs
```diff
@@ -400,6 +431,33 @@ impl Id<Company> for Company {
...
+pub enum CommentType {
+    Information,
+    OnDemandTransport,
+}
+
+impl Default for CommentType {
+    fn default() -> Self {
+        CommentType::Information
+    }
+}
```
TeXitoi: you can use https://crates.io/crates/derivative

src/objects.rs
datanel: absolutely I remove it?

src/ntfs.rs
datanel: oups

src/objects.rs
TeXitoi: Yeah, we'll handle all of that later

src/ntfs.rs
TeXitoi: a make_collection_from_reader can be great

datanel: or I can do `if path.join("comments.txt").exists() {`

TeXitoi: looks good, yes.

TeXitoi: even better: `is_file`

TeXitoi: maybe not, as you wish.

src/ntfs.rs
```diff
@@ -253,6 +255,61 @@ fn manage_feed_infos(collections: &mut Collections, path: &path::Path) {
...
+                    "stop_point" => insert_comment_link(&mut collections.stop_points, comment_link),
+                    "line" => insert_comment_link(&mut collections.lines, comment_link),
+                    "route" => insert_comment_link(&mut collections.routes, comment_link),
+                    "trip" => insert_comment_link(&mut collections.vehicle_journeys, comment_link),
+                    // TODO find something better
+                    // "stop_time" => {
+                    //     insert_comment_link(&mut collections.stop_times, comment_link)
+                    // }
+                    // TODO
+                    // "line_group" => insert_comment_link(&mut collections.line_groups, code),
```
TeXitoi: you may want to handle the case with a `warn!()` message to not crash if you have data with stop_time or line_group comment.

PR #25, opened 2018-01-17
Add license headers

TeXitoi:
The LICENSE file was added in another commit thanks to the github "add license" feature


datanel:
we don't need to add LICENSE or COPYING file?

TeXitoi:
LICENSE is already here. Added yesterday with a github button.

PR #28, opened 2018-01-22
GTFS - read stops file

prhod:
There is still work on equipments, but that can be done later.


src/gtfs.rs
```diff
@@ -53,12 +53,89 @@ impl From<Agency> for objects::Company {
...
+    #[serde(rename = "zone_id")]
+    fare_zone_id: Option<String>,
+    #[serde(rename = "stop_url")]
+    url: Option<String>,
+    #[serde(default)]
+    location_type: i32,
+    parent_station: Option<String>,
+    #[serde(rename = "stop_timezone")]
+    timezone: Option<String>,
+    #[serde(default, rename = "wheelchair_boarding")]
```
datanel: do you really need to rename `wheelchair_boarding` to `wheelchair_boarding`?

prhod: you're right ! fixed

src/gtfs.rs
```diff
@@ -53,12 +53,89 @@ impl From<Agency> for objects::Company {
...
+            visible: true,
+            geometry_id: None,
+            equipment_id: None,
+        }
+    }
+}
+impl From<Stop> for objects::StopPoint {
+    fn from(stop: Stop) -> objects::StopPoint {
+        let id = stop.id;
+        let stop_area_id = stop.parent_station.unwrap_or_else(|| format!("SA{}", id));
```
TeXitoi: Also here.

TeXitoi: With it's test.

PR #33, opened 2018-02-01
Gtfs - read routes.txt and load mdoes and lines

prhod:
Work is partial, but I don't want to wait and have a too big PR.
Yet to be done : 
* creating lines from gtfs routes when gtfs contains routes
* moving/creating methods around physical_modes in the model to mutualize their use in connectors

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,5 @@
 **/*.rs.bk
 Cargo.lock
 *~
+/*.txt
+/*.zip
--- a/.gitignore
+++ b/.gitignore
@@ -2,5 +2,3 @@
 **/*.rs.bk
 Cargo.lock
 *~
-/*.txt
-/*.zip
--- a/src/gtfs/read.rs
+++ b/src/gtfs/read.rs
@@ -18,7 +18,7 @@ use std::path;
 use csv;
 use collection::CollectionWithId;
 use Collections;
-use objects::{self, KeysValues, Coord};
+use objects::{self, Coord, KeysValues};
 use std::collections::HashSet;
 
 fn default_agency_id() -> String {
@@ -182,7 +182,7 @@ impl RouteType {
 impl<'de> ::serde::Deserialize<'de> for RouteType {
     fn deserialize<D>(deserializer: D) -> Result<RouteType, D::Error>
     where
-        D: ::serde::Deserializer<'de>
+        D: ::serde::Deserializer<'de>,
     {
         let mut i = u16::deserialize(deserializer)?;
         if i > 7 && i < 99 {
@@ -240,7 +240,6 @@ impl Route {
         return false;
     }
 }
-
 
 pub fn read_agency<P: AsRef<path::Path>>(
     path: P,
@@ -497,8 +496,10 @@ mod tests {
     #[test]
     fn load_complete_agency() {
         let agency_content =
-            "agency_id,agency_name,agency_url,agency_timezone,agency_lang,agency_phone,agency_fare_url,agency_email\n\
-             id_1,My agency,http://my-agency_url.com,Europe/London,EN,0123456789,http://my-agency_fare_url.com,my-mail@example.com";
+            "agency_id,agency_name,agency_url,agency_timezone,agency_lang,agency_phone,\
+             agency_fare_url,agency_email\n\
+             id_1,My agency,http://my-agency_url.com,Europe/London,EN,0123456789,\
+             http://my-agency_fare_url.com,my-mail@example.com";
         let tmp_dir = TempDir::new("navitia_model_tests").expect("create temp dir");
         let file_path = tmp_dir.path().join("agency.txt");
         let mut f = File::create(&file_path).unwrap();
--- a/src/gtfs/read.rs
+++ b/src/gtfs/read.rs
@@ -194,13 +194,11 @@ struct Route {
     route_type: RouteType,
     #[serde(rename = "route_url")]
     url: Option<String>,
-    #[serde(rename = "route_color", default,
-        deserialize_with = "string_to_color",
-        serialize_with = "ser_from_color")]
+    #[serde(rename = "route_color", default, deserialize_with = "string_to_color",
+            serialize_with = "ser_from_color")]
     color: Option<objects::Rgb>,
-    #[serde(rename = "route_text_color", default,
-        deserialize_with = "string_to_color",
-        serialize_with = "ser_from_color")]
+    #[serde(rename = "route_text_color", default, deserialize_with = "string_to_color",
+            serialize_with = "ser_from_color")]
     text_color: Option<objects::Rgb>,
 }
 
@@ -221,8 +219,7 @@ impl Route {
     }
 }
 
-
-fn de_route_type<'de, D>(deserializer: D) -> Result<RouteType, D::Error> 
+fn de_route_type<'de, D>(deserializer: D) -> Result<RouteType, D::Error>
 where
     D: ::serde::Deserializer<'de>,
 {
@@ -256,7 +253,7 @@ where
     serializer.serialize_str(&s)
 }
 
-fn string_to_color<'de, D>(deserializer: D) -> Result<Option<objects::Rgb>, D::Error> 
+fn string_to_color<'de, D>(deserializer: D) -> Result<Option<objects::Rgb>, D::Error>
 where
     D: ::serde::Deserializer<'de>,
 {
@@ -265,7 +262,9 @@ where
     if s.is_empty() {
         Ok(None)
     } else {
-        s.parse().map(|c| Some(c)).map_err(::serde::de::Error::custom)
+        s.parse()
+            .map(|c| Some(c))
+            .map_err(::serde::de::Error::custom)
     }
 }
 
@@ -362,7 +361,7 @@ fn get_commercial_mode(route_type: &RouteType) -> objects::CommercialMode {
     objects::CommercialMode {
         id: route_type.to_gtfs_value(),
         name: get_commercial_mode_label(&route_type),
-    }    
+    }
 }
 
 fn get_physical_mode(route_type: &RouteType) -> objects::PhysicalMode {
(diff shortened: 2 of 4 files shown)


src/gtfs/read.rs
```diff
@@ -199,12 +220,191 @@ pub fn read_stops<P: AsRef<path::Path>>(
...
+                    } else {
+                        if r1.long_name == r2.long_name {
+                            continue;
+                        } else {
+                            return RouteReadType::RouteAsNtmRoute;
+                        }
+                    }
+                }
+            }
+        }
```
datanel: a lot of if else, I think you can do something shorter

```rust
for (i1, r1) in gtfs_routes.iter().enumerate() {
    for (i2, r2) in gtfs_routes.iter().enumerate() {
        if i2 <= i1 {
            continue;
        } 

        if r1.agency_id == r2.agency_id {
            if r1.short_name != "" {
                if r1.short_name == r2.short_name {
                    continue;
                } 
            } else {
                if r1.long_name == r2.long_name {
                    continue;
                }
            }
            
            return RouteReadType::RouteAsNtmRoute;
        }
    }
}
```

src/objects.rs
```diff
@@ -176,6 +177,58 @@ pub struct Rgb {
...
+        }
+    }
+}
+
+impl Error for RgbError {
+    fn description(&self) -> &str {
+        match *self {
+            RgbError::NotHexa => "String is not a valid Hexadecimal value",
+            RgbError::TooLongHexa => "String is too long (6 char expected)",
+            RgbError::TooShortHexa => "String is too shot (4 char expected)",
```
datanel: too sho**r**t 
why 4?

.gitignore
```diff
@@ -2,3 +2,5 @@
 **/*.rs.bk
 Cargo.lock
 *~
+/*.txt
+/*.zip
```
datanel: this can be added in your global .gitignore

prhod: we don't have any txt or zip for the moment, I don't see what is the problem

datanel: 1) only you have .txt and .zip in navitia_model directory.
All your "personal" files can be added in ~/.gitignore

2) later we will have .txt files for tests. We will need to rollback this change.

3) I also have ".out" and ".orig" and more files but I don't add them in this .gitignore 

but anyway it is not critical

prhod: ok, I've reverted the modification

src/gtfs/read.rs
```diff
@@ -199,12 +220,186 @@ pub fn read_stops<P: AsRef<path::Path>>(
...
+            id: "Bus".to_string(),
+            name: "Bus".to_string(),
+            co2_emission: None,
+        },
+    };
+    mode
+}
+
+fn get_modes_from_gtfs(
+    gtfs_routes: &Vec<Route>,
```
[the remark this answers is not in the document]
datanel: why?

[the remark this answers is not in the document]
TeXitoi: because all you can do with `&Vec<T>`, you can do it with `&[T]`, and there is cases you can use `&[T]` but not `&Vec<T>` (subslice, array on stack...).

prhod:
I think I made all the requested changes. I had to add serialize/deserialize functions or Traits implementations. Please tell me if there are further improvements to do ^^

src/gtfs/read.rs
```diff
@@ -339,4 +625,38 @@ mod tests {
...
+        assert_eq!(3, collections.lines.len());
+        assert_eq!(2, collections.commercial_modes.len());
+
+        let commercial_modes: Vec<String> = collections
+            .commercial_modes
+            .iter()
+            .map(|(_, ref cm)| cm.name.to_string())
+            .collect();
+        assert!(commercial_modes.contains(&"Bus".to_string()));
+        assert!(commercial_modes.contains(&"Rail".to_string()));
```
TeXitoi: You can compare vector quite easily:
```rust
    let v = vec!["toto".to_string(), "tata".to_string()];
    assert_eq!(v, &["toto", "tata"]);
```

src/gtfs/read.rs
```diff
@@ -199,12 +321,176 @@ pub fn read_stops<P: AsRef<path::Path>>(
...
+        &RouteType::Gondola_SuspendedCableCar => objects::PhysicalMode {
+            id: "Funicular".to_string(),
+            name: "Funicular".to_string(),
+            co2_emission: None,
+        },
+        &RouteType::Funicular => objects::PhysicalMode {
+            id: "Funicular".to_string(),
+            name: "Funicular".to_string(),
+            co2_emission: None,
+        },
```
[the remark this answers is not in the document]
prhod: I have trouble with this one. The borrow checker raise an error on `match *route_type` and if I try to remove `RouteType::` using `use RouteType::*` I have an other weird error `pattern doesn't bind 'Funicular'` like the compiler understand the match arms like variables ... 
I ended combining match arms the long way...
Can you give me a hand on this one ?

src/gtfs/read.rs
```diff
@@ -199,12 +321,176 @@ pub fn read_stops<P: AsRef<path::Path>>(
...
+        .map(|mt| get_physical_mode(mt))
+        .collect();
+    (commercial_modes, physical_modes)
+}
+
+fn get_lines_from_gtfs(gtfs_routes: &Vec<Route>, read_mode: RouteReadType) -> Vec<objects::Line> {
+    let mut lines = vec![];
+    match read_mode {
+        RouteReadType::RouteAsNtmLine => for r in gtfs_routes {
+            let line_code = match r.short_name.as_ref() {
```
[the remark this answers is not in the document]
prhod: thanks for the hint, but I like it this way :)

src/gtfs/read.rs
```diff
@@ -199,12 +321,176 @@ pub fn read_stops<P: AsRef<path::Path>>(
...
+                codes: vec![],
+                comment_links: vec![],
+                name: r.long_name.to_string(),
+                forward_name: None,
+                forward_direction: None,
+                backward_name: None,
+                backward_direction: None,
+                color: r.color.clone(),
+                text_color: r.text_color.clone(),
+                sort_order: None,
```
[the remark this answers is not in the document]
prhod: I saw it with you, but I don't seem to be able to find it ! 
I checked https://developers.google.com/transit/gtfs/reference/ and http://gtfs.org/reference/ => no reference to it.
There is a mention here https://mbta.com/developers/gtfs-documentation but with an experimental status. I'm adding it just in case ...

src/gtfs/read.rs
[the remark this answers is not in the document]
prhod: I found the solution : 
```rust
use self::RouteType::*;
```
I thought using it inside the two functions would be best to avoid any conflict with the `Other(u16)` of the enum.

src/gtfs/read.rs
```diff
@@ -191,12 +191,12 @@ struct Route {
     route_type: RouteType,
     #[serde(rename = "route_url")]
     url: Option<String>,
-    #[serde(rename = "route_color", default, deserialize_with = "string_to_color",
-            serialize_with = "ser_from_color")]
+    #[serde(rename = "route_color", default, deserialize_with = "string_to_color")]
```
TeXitoi: I think you don't need a custom deserializer, it should work out of the box.

src/gtfs/read.rs
```diff
@@ -239,17 +237,6 @@ where
...
-    S: ::serde::Serializer,
-{
-    let s = match color {
-        &Some(ref c) => format!("{}", c).to_string(),
-        &None => "".to_string(),
-    };
-    serializer.serialize_str(&s)
-}
-
 fn string_to_color<'de, D>(deserializer: D) -> Result<Option<objects::Rgb>, D::Error>
```
TeXitoi: This code should be useless

src/objects.rs
```diff
@@ -176,6 +180,65 @@ pub struct Rgb {
...
+        }
+    }
+}
+
+impl Error for RgbError {
+    fn description(&self) -> &str {
+        match *self {
+            RgbError::NotHexa => "String is not a valid Hexadecimal value",
+            RgbError::TooLongHexa => "String is too long (6 char expected)",
+            RgbError::TooShortHexa => "String is too short (6 char expected)",
```
datanel: 6 char**S** ?

src/gtfs/read.rs
```diff
@@ -143,6 +145,113 @@ impl From<Stop> for objects::StopPoint {
...
+                }
+            } else if self.long_name == other.long_name {
+                return true;
+            }
+        }
+        return false;
+    }
+}
+
+fn de_route_type<'de, D>(deserializer: D) -> Result<RouteType, D::Error>
```
TeXitoi: Better to direcly impl Deserialize

src/gtfs/read.rs
```diff
@@ -143,6 +145,113 @@ impl From<Stop> for objects::StopPoint {
...
+where
+    D: ::serde::Deserializer<'de>,
+{
+    use serde::Deserialize;
+    let mut i = u16::deserialize(deserializer)?;
+    if i > 7 && i < 99 {
+        i = 3;
```
TeXitoi: Shouldn't you error in this case?

src/gtfs/read.rs
```diff
@@ -199,12 +308,168 @@ pub fn read_stops<P: AsRef<path::Path>>(
...
+            co2_emission: None,
+        },
+        Ferry => objects::PhysicalMode {
+            id: "Ferry".to_string(),
+            name: "Ferry".to_string(),
+            co2_emission: None,
+        },
+        CableCar | Gondola_SuspendedCableCar | Funicular => objects::PhysicalMode {
+            id: "Funicular".to_string(),
+            name: "Funicular".to_string(),
```
TeXitoi: Sometime in French, sometime in English

src/gtfs/read.rs
prhod: you're right, it's automagic ! :D

src/gtfs/read.rs
prhod: ok, I'll do it asap

src/gtfs/read.rs
prhod: I didn't want to just panic with an invalid gtfs, but make a fallback on bus mode.

src/gtfs/read.rs
prhod: you're right, it should be consistent. I'll make it all in English, sounds better for this piece of soft ;)

src/gtfs/read.rs
TeXitoi: At least `err!("illegal route_type: {}, using 3 as fallback")`

PR #36, opened 2018-02-02
Serialize comments and codes

datanel:


src/ntfs/mod.rs
```diff
@@ -73,6 +73,21 @@ struct Stop {
     equipment_id: Option<String>,
 }
 
+#[derive(Serialize, Deserialize, Debug, Clone)]
+struct CommentLink {
+    object_id: String,
+    object_type: String,
```
TeXitoi: Maybe we should type that a bit more. I think of something like
```rust
enum ObjectType {
    StopArea,
    StopPoint,
    Network,
    ...
}
impl ObjectType {
    fn as_str(&self) -> &'static str {...}
}
impl FromStr for ObjectType {...}
trait GetObjectType {
    fn get_object_type() -> ObjectType;
}
impl GetObjectType for StopArea {
    fn get_object_type() -> ObjectType { ObjectType::StopArea }
}
...
```

src/ntfs/write.rs
```diff
@@ -158,3 +159,80 @@ pub fn write_stops(
...
 }
+
+fn write_comment_links_from_collection_with_id<W, T>(
+    wtr: &mut csv::Writer<W>,
+    collections: &CollectionWithId<T>,
+    object_type: &str,
+) where
+    T: Id<T> + CommentLinks,
+    W: ::std::io::Write,
+{
```
TeXitoi: Then, you can have something like
```rust
fn write_comment_links_from_collection_with_id<W, T>(wtr: &mut csv::Writer<W>, collections: &CollectionWithId<T>)
where
    T: Id<T> + CommentLinks + GetObjectType,
    W: ::std::io::Write,
```

src/ntfs/read.rs
```diff
@@ -186,14 +179,14 @@ pub fn manage_codes(collections: &mut Collections, path: &path::Path) {
...
-            "trip" => insert_code(&mut collections.vehicle_journeys, code),
-            _ => panic!("{} is not a valid object_type", code.object_type),
+        match code.object_type {
+            ObjectType::StopArea => insert_code(&mut collections.stop_areas, code),
+            ObjectType::StopPoint => insert_code(&mut collections.stop_points, code),
+            ObjectType::Network => insert_code(&mut collections.networks, code),
+            ObjectType::Line => insert_code(&mut collections.lines, code),
+            ObjectType::Route => insert_code(&mut collections.routes, code),
+            ObjectType::VehicleJourney => insert_code(&mut collections.vehicle_journeys, code),
+            _ => (),
```
TeXitoi: It should panic in this case as `panic!("code does not support {}", code.object_type)`

src/ntfs/read.rs
```diff
@@ -287,15 +274,21 @@ pub fn manage_comments(collections: &mut Collections, path: &path::Path) {
...
+                        insert_comment_link(&mut collections.stop_points, comment_link)
+                    }
+                    ObjectType::Line => insert_comment_link(&mut collections.lines, comment_link),
+                    ObjectType::Route => insert_comment_link(&mut collections.routes, comment_link),
+                    ObjectType::VehicleJourney => {
+                        insert_comment_link(&mut collections.vehicle_journeys, comment_link)
+                    }
+                    ObjectType::StopTime => warn!("comments are not added to StopTime yet"),
+                    ObjectType::LineGroup => warn!("line_groups.txt is not parsed yet"),
+                    _ => (),
```
TeXitoi: Same here, or don't do a catch all if all enum variants are checked.

src/objects.rs
```diff
@@ -18,6 +18,45 @@ use collection::{Id, Idx};
 use utils::*;
 use chrono;
 
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub enum ObjectType {
+    #[serde(rename = "stop_area")]
```
TeXitoi: You can do a `#[serde(rename_all = "snake_case")]` to do almost everything (except trip)

datanel:
```rust
#[derive(Serialize, Deserialize, Debug, Clone)]
#[serde(rename_all = "snake_case")]
pub enum ObjectType {
    StopArea,
    StopPoint,
    Network,
    Line,
    Route,
    #[serde(rename = "trip")]
    VehicleJourney,
    StopTime,
    LineGroup,
}
```

PR #37, opened 2018-02-06
Added trip properties

datanel:


src/objects.rs
```diff
@@ -704,6 +704,46 @@ pub struct Transfer {
...
+#[derive(Serialize, Deserialize, Debug, Derivative, PartialEq)]
+#[derivative(Default)]
+pub enum TransportType {
+    #[derivative(Default)]
+    #[serde(rename = "0")]
+    Regular,
+    #[serde(rename = "1")]
+    ExclusiveSchool,
+    #[serde(rename = "2")]
+    Mixed,
```
datanel: better names? @prhod @TeXitoi

TeXitoi: I'm not found of these names, but I can't find anything better now.

prhod: For the `Mixed` value, it's meaning is not clear when read alone. What about `RegularAndSchool` ?

datanel: ok for RegularAndSchool

PR #41, opened 2018-02-07
(de)serialize geometries

nberard:


src/objects.rs
```diff
@@ -776,6 +776,20 @@ impl Id<TripProperty> for TripProperty {
     }
 }
 
+#[derive(Serialize, Deserialize, Debug, PartialEq)]
+pub struct Geometry {
+    #[serde(rename = "geometry_id")]
+    pub id: String,
+    #[serde(deserialize_with = "de_with_empty_default")]
+    pub geometry_wkt: String,
```
datanel: you can name it `wkt` and add `rename="geometry_wkt"`

src/ntfs/mod.rs
```diff
@@ -123,7 +123,7 @@ pub fn read<P: AsRef<path::Path>>(path: P) -> PtObjects {
     read::manage_codes(&mut collections, path);
     read::manage_comments(&mut collections, path);
     read::manage_object_properties(&mut collections, path);
-
+    collections.geometries = read::make_collection_with_id(path, "geometries.txt");
```
TeXitoi: make_collection_with_id are grouped a bit before

src/ntfs/read.rs
```diff
@@ -165,7 +165,7 @@ where
     let idx = match collection.get_idx(&code.object_id) {
         Some(idx) => idx,
         None => {
-            error!(
+            warn!(
```
TeXitoi: I prefer error, that's a violation of the format.

src/ntfs/read.rs
```diff
@@ -197,7 +197,7 @@ fn insert_calendar_date(collection: &mut CollectionWithId<Calendar>, calendar_da
     let idx = match collection.get_idx(&calendar_date.service_id) {
         Some(idx) => idx,
         None => {
-            error!(
+            warn!(
```
TeXitoi: Same here

src/ntfs/read.rs
```diff
@@ -254,7 +254,7 @@ where
     let idx = match collection.get_idx(&comment_link.object_id) {
         Some(idx) => idx,
         None => {
-            error!(
+            warn!(
```
TeXitoi: Same

src/ntfs/read.rs
```diff
@@ -307,7 +307,7 @@ where
     let idx = match collection.get_idx(&obj_prop.object_id) {
         Some(idx) => idx,
         None => {
-            error!(
+            warn!(
```
TeXitoi: Same

src/objects.rs
```diff
@@ -776,6 +776,20 @@ impl Id<TripProperty> for TripProperty {
     }
 }
 
+#[derive(Serialize, Deserialize, Debug, PartialEq)]
+pub struct Geometry {
+    #[serde(rename = "geometry_id")]
+    pub id: String,
+    #[serde(deserialize_with = "de_with_empty_default")]
```
TeXitoi: You can just remove this line. An empty field is not a valid geometry.

src/ntfs/read.rs
```diff
@@ -288,8 +288,8 @@ pub fn manage_comments(collections: &mut Collections, path: &path::Path) {
                     ObjectType::VehicleJourney => {
                         insert_comment_link(&mut collections.vehicle_journeys, comment_link)
                     }
-                    ObjectType::StopTime => warn!("comments are not added to StopTime yet"),
-                    ObjectType::LineGroup => warn!("line_groups.txt is not parsed yet"),
+                    ObjectType::StopTime => error!("comments are not added to StopTime yet"),
+                    ObjectType::LineGroup => error!("line_groups.txt is not parsed yet"),
```
TeXitoi: Sorry, these ones are warn because it should work but we warn that it's not supported yet.

nberard: omg my bad >_<

PR #48, opened 2018-02-13
implement serialization/deserialization of admin_stations

nberard:


TeXitoi:
@pbougue in reallity, you want the other way (give me all the stop areas from an admin).

As we will not do anything interested now with it, we just do it the easy way, and will refactor when we'll have a use case.

PR #50, opened 2018-02-14
read gtfs config from json file

nberard:
json file expected should look like:

```json
{
  "contributor": {
    "contributor_id": "OIF",
    "contributor_name": "OIF - STIF Ile de France",
    "contributor_website": "http://www.vianavigo.com",
    "contributor_license": "ODbL"
  },
  "dataset": {
    "contributor_id": "OIF",
    "dataset_id": "OIF:0",
    "dataset_start_date": "20170607",
    "dataset_end_date": "20170702",
    "dataset_type": "0",
    "dataset_desc": "NIO_auto_20170618",
    "dataset_system": "GTFS V2 Ligne",
    "dataset_extrapolation": 0
  }
}
```

to test:
```
cargo run -- -i /path/to/input -o /path/to/output -c /path/to/config
```

### DIFF
--- a/examples/gtfs_reader.rs
+++ b/examples/gtfs_reader.rs
@@ -15,12 +15,25 @@
 // <http://www.gnu.org/licenses/>.
 
 extern crate csv;
+extern crate env_logger;
 extern crate navitia_model;
 #[macro_use]
 extern crate serde_json;
+use navitia_model::Result;
 
-fn main() {
-    let objects = navitia_model::gtfs::read(".", None);
+fn run() -> Result<()> {
+    let objects = navitia_model::gtfs::read(".", None)?;
     let json_objs = json!(objects);
     println!("{:?}", json_objs.to_string());
+    Ok(())
+}
+
+fn main() {
+    env_logger::init();
+    if let Err(err) = run() {
+        for cause in err.causes() {
+            eprintln!("{}", cause);
+        }
+        std::process::exit(1);
+    }
 }
--- a/examples/gtfs_reader.rs
+++ b/examples/gtfs_reader.rs
@@ -20,7 +20,7 @@ extern crate navitia_model;
 extern crate serde_json;
 
 fn main() {
-    let objects = navitia_model::gtfs::read(".");
+    let objects = navitia_model::gtfs::read(".", None);
     let json_objs = json!(objects);
     println!("{:?}", json_objs.to_string());
 }
--- a/src/gtfs/mod.rs
+++ b/src/gtfs/mod.rs
@@ -20,15 +20,16 @@ use std::path;
 use {Collections, PtObjects};
 use std::fs::File;
 use gtfs::read::Config;
+use Result;
 
 extern crate serde_json;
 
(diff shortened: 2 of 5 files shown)


src/gtfs/mod.rs
```diff
@@ -18,10 +18,24 @@ mod read;
...
     let mut collections = Collections::default();
+
+    if let Some(config_path) = config_path {
+        let json_config_file = File::open(config_path).unwrap();
+        let config: Config = serde_json::from_reader(json_config_file).unwrap();
+        info!("config loaded: {:#?}", config);
+        let (contributors, datasets) = read::read_config(config);
+        collections.contributors = contributors;
+        collections.datasets = datasets;
+    }
```
TeXitoi: We don't create a default one if that's not given?

datanel: @TeXitoi  I don't think so

datanel: contributors.txt and datasets.txt are created but empty

src/gtfs/mod.rs
```diff
@@ -18,10 +18,24 @@ mod read;
...
 use std::path;
 use {Collections, PtObjects};
+use std::fs::File;
+use gtfs::read::Config;
 
-pub fn read<P: AsRef<path::Path>>(path: P) -> PtObjects {
-    let path = path.as_ref();
+extern crate serde_json;
+
+pub fn read<P: AsRef<path::Path>>(path: P, config_path: Option<P>) -> PtObjects {
```
datanel: we should instead now return a `Result<PtObjects>` (see https://github.com/CanalTP/navitia_model/blob/master/src/lib.rs#L48)

src/gtfs/mod.rs
```diff
@@ -18,10 +18,24 @@ mod read;
...
     let mut collections = Collections::default();
+
+    if let Some(config_path) = config_path {
+        let json_config_file = File::open(config_path).unwrap();
```
datanel: replace .unwrap by `?`

src/gtfs/mod.rs
```diff
@@ -18,10 +18,24 @@ mod read;
...
+        let config: Config = serde_json::from_reader(json_config_file).unwrap();
```
datanel: replace .unwrap by `?`

src/gtfs/mod.rs
```diff
@@ -18,10 +18,24 @@ mod read;
...
+        info!("config loaded: {:#?}", config);
+        let (contributors, datasets) = read::read_config(config);
```
datanel: read_config should return a Result I think

examples/gtfs_reader.rs
```diff
@@ -20,7 +20,7 @@ extern crate navitia_model;
 extern crate serde_json;
 
 fn main() {
-    let objects = navitia_model::gtfs::read(".");
+    let objects = navitia_model::gtfs::read(".", None);
```
datanel: you can do the same as `example/test.rs` with `run` function

And we should do the same in `bin/gtfs2ntfs.rs`

PR #54, opened 2018-02-22
Add contributor prefix from config

nberard:


src/gtfs/read.rs
```diff
@@ -25,6 +25,10 @@ fn default_agency_id() -> String {
     "default_agency_id".to_string()
 }
 
+trait PrefixId {
+    fn alter_id_with_prefix(&mut self, prefix: Option<String>);
```
TeXitoi: You want `prefix: &Option<String>` or `prefix: Option<&str>`

src/gtfs/read.rs
```diff
@@ -25,6 +25,10 @@ fn default_agency_id() -> String {
     "default_agency_id".to_string()
 }
 
+trait PrefixId {
```
TeXitoi: I don't get the idea of the trait.

nberard: me neither :p, at first code was common between the 2 struct

PR #59, opened 2018-03-01
refactoring prefix

datanel:
- refactoring add prefix
- create contributor+ dataset related to https://github.com/CanalTP/navitia_model/pull/57/files#diff-a3fce5151b3ec922f7e1e2ed742d4ea2R28

fix https://github.com/CanalTP/navitia_model/issues/61

### DIFF
--- a/src/bin/gtfs2ntfs.rs
+++ b/src/bin/gtfs2ntfs.rs
@@ -40,6 +40,10 @@ struct Opt {
     /// config file
     #[structopt(short = "c", long = "config", parse(from_os_str))]
     config_path: Option<PathBuf>,
+
+    /// prefix
+    #[structopt(short = "p", long = "prefix")]
+    prefix: Option<String>,
 }
 
 fn run() -> Result<()> {
@@ -47,7 +51,8 @@ fn run() -> Result<()> {
 
     let opt = Opt::from_args();
 
-    let objects = navitia_model::gtfs::read(opt.input, opt.config_path)?;
+    let objects = navitia_model::gtfs::read(opt.input, opt.config_path, opt.prefix)?;
+
     navitia_model::ntfs::write(opt.output, &objects)?;
     Ok(())
 }
--- a/src/collection.rs
+++ b/src/collection.rs
@@ -303,21 +303,16 @@ where
     Ok(Collection::new(vec))
 }
 
-pub fn add_prefix<T>(
-    collection: &mut CollectionWithId<T>,
-    prefix: &str,
-) -> Result<CollectionWithId<T>>
+pub fn add_prefix<T>(collection: &mut CollectionWithId<T>, prefix: &str) -> Result<()>
 where
     T: AddPrefix + Id<T>,
 {
-    let objects = collection.take();
-    let objs_with_prefix = objects
-        .into_iter()
-        .map(|mut obj| {
-            obj.add_prefix(prefix);
-            obj
-        })
-        .collect();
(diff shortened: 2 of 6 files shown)


src/bin/gtfs2ntfs.rs
```diff
@@ -39,7 +39,7 @@ struct Opt {
 
     /// config file
     #[structopt(short = "c", long = "config", parse(from_os_str))]
-    config_path: Option<PathBuf>,
+    config_path: PathBuf,
```
TeXitoi: We need to be able to use navitia_model without prefix and without doing that, creating some default objects.

prhod: For the GTFS->NTFS conversion, I asked for a systematic prefix with a pattern <prefix>:<object_id>. If I understand, you ask for an optional config file that would create default contributor and dataset. How about a predefined prefix like "default_prefix" ?
An empty <prefix> would imply also removing the ":" separator, and would add quite complexity to the code. 
If you're all ok with that, I'll specify the default values to avoid refacto :D

src/gtfs/mod.rs
```diff
@@ -17,36 +17,35 @@
...
-    let (stopareas, stoppoints) = read::read_stops(path, &contributor_as_prefix)?;
-    collections.stop_areas = stopareas;
-    collections.stop_points = stoppoints;
+    let (mut networks, mut companies) = read::read_agency(path)?;
+    let (mut stopareas, mut stoppoints) = read::read_stops(path)?;
     manage_calendars(&mut collections, path)?;
-    read::read_routes(path, &contributor_as_prefix, &mut collections)?;
+    read::read_routes(path, &mut collections)?;
+
+    //add prefixes
```
TeXitoi:
```rust
if let Some(ref prefix) = prefix {
    // ...
```

src/gtfs/read.rs
```diff
@@ -373,22 +337,56 @@ pub fn read_stops<P: AsRef<path::Path>>(
...
+        info!("config loaded: {:#?}", config);
+
+        contributor = config.contributor;
+
+        use chrono::{Duration, Utc};
+        let duration = Duration::days(15);
+        let today = Utc::today();
+        let start_date = today - duration;
+        let end_date = today + duration;
+        dataset = objects::Dataset {
```
TeXitoi: I don't get it, dataset is not given in the config?

datanel: dataset in the config only have dataset_id and optional properties
see https://github.com/CanalTP/navitia_model/pull/57/files#diff-a3fce5151b3ec922f7e1e2ed742d4ea2R28

src/collection.rs
```diff
@@ -302,3 +302,22 @@ where
...
+    T: AddPrefix + Id<T>,
+{
+    let objects = collection.take();
+    let objs_with_prefix = objects
+        .into_iter()
+        .map(|mut obj| {
+            obj.add_prefix(prefix);
+            obj
+        })
+        .collect();
```
TeXitoi: You destroy a vector and recreate it. Better to just modifying it in place:
```rust
pub fn add_prefix<T>(
    collection: &mut CollectionWithId<T>,
    prefix: &str,
) -> Result<()>
where
    T: AddPrefix + Id<T>,
{
    let objects = collection.take();
    for obj in &mut objects {
        obj.add_prefix(prefix);
    }
    *collection = CollectionWithId::new(objects);
    Ok(())
}
```

PR #65, opened 2018-03-08
Comments : add in read/write func + idx refactoring

benoit-bst:


benoit-bst:
Update Comment.value into Comment.name also.

datanel:
welcome new rustacean

datanel:
you need to run `rustmft` as configured in travis.yml

benoit-bst:
I created a new data set (fixtures/ntfs) to add comments.txt and comment_links.txt because fixtures/minimal does not contain the files

src/ntfs/read.rs
```diff
@@ -227,8 +227,11 @@ pub fn manage_feed_infos(collections: &mut Collections, path: &path::Path) -> Re
     Ok(())
 }
 
-fn insert_comment_link<T>(collection: &mut CollectionWithId<T>, comment_link: CommentLink)
-where
+fn insert_comment_link<T>(
+    collection: &mut CollectionWithId<T>,
+    comments: &CollectionWithId<Comment>,
+    comment_link: CommentLink,
```
datanel: my friend clippy says;
> warning: this argument is passed by value, but not consumed in the function body
   --> src/ntfs/read.rs:233:19
    |
233 |     comment_link: CommentLink,
    |                   ^^^^^^^^^^^ help: consider taking a reference instead: `&CommentLink`

src/lib.rs
```diff
@@ -127,6 +157,17 @@ impl PtObjects {
                 Ok((idx, stop_points))
             })
             .collect::<StdResult<BTreeMap<_, _>, Error>>()?;
+        let lines_to_comments =
+            ManyToMany::from_forward(create_relation_with_comments::<Line>(&c.lines));
```
datanel: @TeXitoi is it necessary to write `::<Line>` ?

TeXitoi: It shouldn't, but that's easier to just test.

src/ntfs/read.rs
benoit-bst: :white_check_mark:

src/lib.rs
benoit-bst: :white_check_mark:

src/ntfs/read.rs
```diff
@@ -245,7 +248,7 @@ where
     collection
         .index_mut(idx)
         .comment_links_mut()
-        .push(comment_link.comment_id);
+        .push(comments.get_idx(&comment_link.comment_id).unwrap());
```
TeXitoi: This `unwrap` might fail if the data is bad, thus you should handle error. @prhod log or fail?

src/lib.rs
```diff
@@ -89,6 +89,24 @@ pub struct PtObjects {
...
+    #[get_corresponding(weight = "100")]
+    lines_to_comments: ManyToMany<Line, Comment>,
+    #[get_corresponding(weight = "100")]
+    routes_to_comments: ManyToMany<Route, Comment>,
+    #[get_corresponding(weight = "100")]
+    vehicle_journeys_to_comments: ManyToMany<VehicleJourney, Comment>,
+    #[get_corresponding(weight = "100")]
+    stop_areas_to_comments: ManyToMany<StopArea, Comment>,
+    #[get_corresponding(weight = "100")]
+    stop_points_to_comments: ManyToMany<StopPoint, Comment>,
```
TeXitoi: This may not be really interesting as it will only be used to convert `Vec<Idx<Comment>>` to `IdxSet<Comment>`.

But we will need something like that on `Collection<T>` (not sure of the name, @datanel any idea?):
```rust
fn iter_from<I>(&self, iter: I) -> impl Iterator<Item = &Comment>
where
    I: IntoIterator,
    I::Item: Into<Idx<Comment>>,
```
for doing something like that:
```rust
for comment in pt_objects.comments.iter_from(&stop_area.comment_links) {
    println!("{}", comment.name);
}
```
And `comment_links` may be a bad name, not representing what it is. Maybe `comments` would be better? @datanel what do you think?

src/ntfs/read.rs
prhod: As discussed together, we can fail for the moment. We will be permissive only if/when needed.

src/collection.rs
```diff
@@ -69,6 +70,18 @@ impl<T> Collection<T> {
...
+    where
+        I: IntoIterator + 'a,
+        I::Item: Borrow<Idx<T>>,
+    {
+        Box::new(
+            indexes
+                .into_iter()
+                .map(move |item| &self.objects[item.borrow().get()]),
+        )
+    }
```
datanel: :scream:

src/ntfs/read.rs
TeXitoi: @benoit-bst You must then return a `Result` in this function

src/ntfs/write.rs
```diff
@@ -197,19 +197,20 @@ pub fn write_stops(
...
     path: &path::Path,
 ) -> Result<()>
 where
     T: Id<T> + CommentLinks + GetObjectType,
     W: ::std::io::Write,
 {
-    for (_, obj) in collections.iter() {
-        for c_id in obj.comment_links() {
+    for (_, obj) in collection.iter() {
+        for &c_id in obj.comment_links() {
```
TeXitoi:
```rust
for comment in comments.iter_from(&obj.comment_links()) {
```

benoit-bst: :white_check_mark:

PR #75, opened 2018-04-04
Forgot some prefixes

datanel:
Added some prefixes in:
- trip_property_id in VehicleJourney object
- commercial_mode_id in Line and CommercialMode

### DIFF
--- a/src/objects.rs
+++ b/src/objects.rs
@@ -227,6 +227,11 @@ impl Id<CommercialMode> for CommercialMode {
         &self.id
     }
 }
+impl AddPrefix for CommercialMode {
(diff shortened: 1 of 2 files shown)


src/objects.rs
```diff
@@ -533,6 +539,9 @@ impl AddPrefix for VehicleJourney {
         self.route_id = prefix.to_string() + &self.route_id;
         self.dataset_id = prefix.to_string() + &self.dataset_id;
         self.company_id = prefix.to_string() + &self.company_id;
+        self.trip_property_id = self.trip_property_id
+            .clone()
```
TeXitoi:
```rust
.as_ref()
```

PR #77, opened 2018-04-04
do not add trip properties if there is no accessibility information in the trips

datanel:


src/gtfs/read.rs
```diff
@@ -792,31 +791,28 @@ fn make_ntfs_vehicle_journeys(
...
-                id: id.to_string(),
+    // If there is no accessibility information in the trips
+    if map_tps_trips.len() == 1 && map_tps_trips.keys().next() == Some(&(0, 0)) {
+        for trips in map_tps_trips.values() {
+            for t in trips {
+                vehicle_journeys.push(t.to_ntfs_vehicle_journey(routes, dataset, None));
+            }
+        }
+    } else {
+        for (&(weelchair_id, bike_id), trips) in &map_tps_trips {
```
TeXitoi: wheelchair with an h

src/gtfs/read.rs
```diff
@@ -792,31 +791,28 @@ fn make_ntfs_vehicle_journeys(
...
+            .push(t);
     }
 
-    let trip_properties: Result<Vec<objects::TripProperty>> = gtfs_tps_ids
-        .iter()
-        .map(|(&(weelchair_id, bike_id), id)| {
-            let trip_property = objects::TripProperty {
-                id: id.to_string(),
+    // If there is no accessibility information in the trips
+    if map_tps_trips.len() == 1 && map_tps_trips.keys().next() == Some(&(0, 0)) {
```
TeXitoi: Is this really what you want? Maybe you just want to never create the empty property case?

PR #79, opened 2018-04-05
[stop times] fixed typo dropoff_type => drop_off_type

datanel:


src/gtfs/read.rs
```diff
@@ -1461,7 +1461,7 @@ mod tests {
 
         let stop_times_content = "trip_id,arrival_time,departure_time,stop_id,stop_sequence,stop_headsign,pickup_type,drop_off_type,shape_dist_traveled\n\
                                   1,06:00:00,06:00:00,sp:01,1,,,,\n\
-                                  1,06:06:27,06:06:27,sp:02,2,,,,";
+                                  1,06:06:27,06:06:27,sp:02,2,,,1,";
```
TeXitoi: You should test that this drop_off_type is present after import.

PR #83, opened 2018-04-26
Added note on transfers

datanel:


prhod:
Would you mind adding something like `(speed value is lowered because effective transit is not straight` ?

datanel:
@prhod yes :)

PR #85, opened 2018-05-03
Add approx distance

TeXitoi:


src/objects.rs
```diff
@@ -674,8 +674,65 @@ impl Coord {
...
     }
+
+    /// Returns a proxy object allowing to compute approximate
+    /// distances for cheap computation.
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// # use navitia_model::objects::Coord;
+    /// let v: Vec<Coord> = vec![];
```
datanel: why v is empty in your example?

TeXitoi: because I'm lazy off adding a lot of coords, and that's not so useful for the example. I'll mask that with a function.

PR #86, opened 2018-05-04
Manage calendars as set of Dates

datanel:


src/collection.rs
```diff
@@ -311,6 +311,12 @@ impl<T: Id<T>> CollectionWithId<T> {
         }
     }
 
+    /// Returns a reference to the object corresponding to the
+    /// identifier.
```
datanel: this comment is wrong.
@TeXitoi can you help me to find a description and a good example please?

src/common_format.rs
```diff
@@ -31,34 +34,133 @@ pub struct CalendarDate {
...
+                    }
+                    ExceptionType::Remove => {
+                        calendar.dates.remove(&calendar_date.date);
+                    }
+                });
+            is_inserted.unwrap_or_else(|| {
+                if calendar_date.exception_type == ExceptionType::Add {
+                    let mut dates = BTreeSet::new();
+                    dates.insert(calendar_date.date);
+                    calendars.push(objects::Calendar {
```
datanel: here I have a warning
```
warning: unused `std::result::Result` which must be used
   --> src/common_format.rs:128:21
    |
128 | /                     calendars.push(objects::Calendar {
129 | |                         id: calendar_date.service_id,
130 | |                         dates,
131 | |                     });
    | |_______________________^
    |
    = note: #[warn(unused_must_use)] on by default
```
I don't know how to remove it

TeXitoi: adding a `?` at the end.

src/collection.rs
TeXitoi: I think you should take inspiration if index_mut. Thus, something like 

> Access to a mutable reference of the object corresponding to the identifier.
>
> The `drop` of the proxy object panic if the identifier is modified to an identifier already on the collection.

and an example close to index_mut

src/objects.rs
```diff
@@ -775,32 +775,14 @@ pub enum ExceptionType {
     Remove,
 }
 
-pub type CalendarDates = Vec<(Date, ExceptionType)>;
+use std::collections::BTreeSet;
 
 #[derive(Serialize, Deserialize, Debug, PartialEq)]
 pub struct Calendar {
-    #[serde(rename = "service_id")]
+    #[serde(skip)]
```
TeXitoi: please remove these skip. It's useless for gtfs/nffs, and may be useful someday if we want to support serde on the ntm::Model.

src/common_format.rs
datanel:
```
--> src/common_format.rs:128:21
    |
128 | /                     calendars.push(objects::Calendar {
129 | |                         id: calendar_date.service_id,
130 | |                         dates,
131 | |                     })?;
    | |_______________________^ cannot use the `?` operator in a function that returns `(
```

datanel: works with unwrap as suggested by @prhod

src/common_format.rs
```diff
@@ -31,34 +34,135 @@ pub struct CalendarDate {
...
+            valid_days.push(chrono::Weekday::Sun);
+        }
+
+        valid_days
+    }
+
+    fn get_valid_dates(&self) -> BTreeSet<Date> {
+        let valid_days = self.get_valid_days();
+        let duration = self.end_date - self.start_date;
+        (0..duration.num_days() + 1)
```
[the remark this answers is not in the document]
datanel: yes because the upper bound of a range in rust is exclusive

PR #87, opened 2018-05-06
Add some extentions to model

prhod:
Adding `push` fn  to `Collection` and `FromStr` trait on `Time` struct.

### DIFF
--- a/src/collection.rs
+++ b/src/collection.rs
@@ -188,10 +188,11 @@ impl<T> Collection<T> {
     /// # fn run() -> navitia_model::Result<()> {
     /// # #[derive(PartialEq, Debug)] struct Obj(&'static str);
     /// # impl Id<Obj> for Obj { fn id(&self) -> &str { self.0 } }
-    /// let mut c = Collection::new(vec![])?;
+    /// let mut c = Collection::new(vec![]);
     /// let foo_idx = c.push(Obj("foo"))?;
     /// let bar_idx = c.push(Obj("bar"))?;
-    /// assert_ne!(foo_idx, &Obj("foo")?);
+    /// assert_eq!(&c[foo_idx], &Obj("foo"));
+    /// assert_ne!(&c[foo_idx], &Obj("bar"));
     /// # Ok(())
     /// # }
     /// # fn main() { run().unwrap() }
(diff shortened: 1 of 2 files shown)


src/collection.rs
```diff
@@ -178,6 +178,31 @@ impl<T> Collection<T> {
...
+    /// let mut c = Collection::new(vec![]);
+    /// let foo_idx = c.push(Obj("foo"))?;
+    /// let bar_idx = c.push(Obj("bar"))?;
+    /// assert_eq!(&c[foo_idx], &Obj("foo"));
+    /// assert_ne!(&c[foo_idx], &Obj("bar"));
+    /// # Ok(())
+    /// # }
+    /// # fn main() { run().unwrap() }
+    /// ```
+    pub fn push(&mut self, item: T) -> Result<Idx<T>> {
```
TeXitoi: This can't fail, thus just return `Idx<T>`

src/objects.rs
```diff
@@ -568,6 +568,28 @@ impl GetObjectType for VehicleJourney {
...
+#[derive(Debug)]
+pub enum TimeError {
+    WrongFormat,
+    WrongValue,
+}
+impl std::fmt::Display for TimeError {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        match *self {
+            TimeError::WrongFormat => f.write_str("TimeError::WrongFormat"),
+            TimeError::WrongValue => f.write_str("TimeError::WrongValue"),
```
TeXitoi: Display is for humans, thus "wrong format" and "wrong value" will do the job.

src/objects.rs
```diff
@@ -584,6 +606,23 @@ impl Time {
...
 }
+impl FromStr for Time {
+    type Err = TimeError;
+    fn from_str(time: &str) -> Result<Self, Self::Err> {
+        let mut t = time.split(':');
+        let (hours, minutes, seconds) = match (t.next(), t.next(), t.next(), t.next()) {
+            (Some(h), Some(m), Some(s), None) => (h, m, s),
+            _ => return Err(TimeError::WrongFormat),
+        };
+        let hours: u32 = hours.parse().map_err(|_err| TimeError::WrongFormat)?;
```
TeXitoi: You can `impl From<std::num::ParseFloatError> for TimeError` to automate these conversions.

prhod: thanks for the hint, I added it using From<ParseIntError>

PR #88, opened 2018-05-14
Compute Validity period

datanel:


src/gtfs/read.rs
```diff
@@ -957,6 +941,42 @@ pub fn read_routes<P: AsRef<path::Path>>(path: P, collections: &mut Collections)
...
+        acc.union(&c.dates).cloned().collect()
+    });
+
+    if dates.is_empty() {
+        return None;
+    }
+
+    Some(objects::ValidityPeriod {
+        start_date: *dates.iter().next().unwrap(),
+        end_date: *dates.iter().next_back().unwrap(),
```
TeXitoi: It will fail with only one day.

datanel: no it won't. I added a test for this case

TeXitoi: ha, my bad, was thinking it was the same iterator, but there is one for each.

Thus, what's wrong with front() and back()?

TeXitoi: OK, my bad.

PR #89, opened 2018-05-16
adding append method for collections

prhod:
These methods are used in netex reading, collecting data file after file.


prhod:
following a discussion with @datanel, not having the use of the method is troublesome. I'll include those modifications in the global PR.

PR #99, opened 2018-06-14
improve contributors compatibility

### LINKED ISSUE #97: gtfs2ntfs - cohérence controle make conversion failing
Trying to convert real GTFS files (Israel or Swiss data for exemple), conversion fails with severral reasons, like transfers referencing stop_id that doesn't exists.
I think the conversion should be permissive, returning warnings for inconsistencies

nberard:
Fix #97 and compatibility with several contributors

### DIFF
--- a/src/gtfs/read.rs
+++ b/src/gtfs/read.rs
@@ -22,7 +22,7 @@ use objects::{
     self, Availability, CommentLinksT, Contributor, Coord, KeysValues, Time, TransportType,
 };
 use read_utils;
-use std::collections::{HashMap, HashSet};
+use std::collections::{BTreeSet, HashSet, HashMap};
 use std::fs::File;
 use std::path;
 use std::result::Result as StdResult;
@@ -79,8 +79,8 @@ impl From<Agency> for objects::Company {
     }
 }
 
-#[derive(Derivative)]
-#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq, Hash)]
+#[derivative(Default)]
+#[derive(Derivative, Serialize, Deserialize, Debug, Clone, PartialEq, Eq, Hash)]
 pub enum StopLocationType {
     #[derivative(Default)]
     #[serde(rename = "0")]
@@ -89,10 +89,6 @@ pub enum StopLocationType {
     StopArea,
     #[serde(rename = "2")]
     StopEntrace,
-}
-
-impl Default for StopLocationType {
-    fn default() -> StopLocationType {StopLocationType::StopPoint}
 }
 
 #[derive(Serialize, Deserialize, Debug, Clone)]
@@ -113,7 +109,7 @@ struct Stop {
     fare_zone_id: Option<String>,
     #[serde(rename = "stop_url")]
     url: Option<String>,
-    #[serde(default,deserialize_with = "de_with_empty_default")]
+    #[serde(default, deserialize_with = "de_with_empty_default")]
     location_type: StopLocationType,
     parent_station: Option<String>,
     #[serde(rename = "stop_timezone")]
@@ -204,8 +200,8 @@ impl RouteType {
 
 impl<'de> ::serde::Deserialize<'de> for RouteType {
     fn deserialize<D>(deserializer: D) -> StdResult<RouteType, D::Error>
-        where
-            D: ::serde::Deserializer<'de>,
+    where
+        D: ::serde::Deserializer<'de>,
     {
         let mut i = u16::deserialize(deserializer)?;
         if i > 7 && i < 99 {
@@ -296,7 +292,7 @@ struct Trip {
     headsign: Option<String>,
     #[serde(rename = "trip_short_name")]
     short_name: Option<String>,
-    #[serde(deserialize_with = "de_with_empty_default", rename = "direction_id")]
+    #[serde(default, deserialize_with = "de_with_empty_default", rename = "direction_id")]
     direction: DirectionType,
     block_id: Option<String>,
     shape_id: Option<String>,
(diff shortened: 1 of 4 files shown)


src/gtfs/read.rs
```diff
@@ -744,10 +750,11 @@ fn get_modes_from_gtfs(
         .iter()
         .map(|mt| get_commercial_mode(mt))
         .collect();
-    let physical_modes = gtfs_mode_types
+    let physical_modes_set: HashSet<objects::PhysicalMode> = gtfs_mode_types
```
prhod: I think you should remove the explicit type

src/gtfs/read.rs
```diff
@@ -744,10 +750,11 @@ fn get_modes_from_gtfs(
...
         .iter()
         .map(|mt| get_physical_mode(mt))
         .collect();
+    let physical_modes: Vec<objects::PhysicalMode> = physical_modes_set.into_iter().collect();
```
prhod: same here

src/objects.rs
```diff
@@ -296,6 +297,20 @@ impl Id<PhysicalMode> for PhysicalMode {
...
+    }
+}
+
+impl PartialEq for PhysicalMode {
+    fn eq(&self, other: &PhysicalMode) -> bool {
+        self.id == other.id && self.name == other.name
+    }
+}
+
+impl Eq for PhysicalMode {}
```
prhod: Is it ok and useful to declare `impl` without any function ?

src/gtfs/read.rs
nberard: it has to be there otherwise it gets a vector, we use hashset first to remove duplicate and then transform to a vector for function return

src/objects.rs
nberard: it has to be there but i guess it uses the declaration from the trait PartialEq
```
Eq has no extra methods, it is only informing the compiler that this is an equivalence relation rather than a partial equivalence relation.
```
see https://doc.rust-lang.org/std/cmp/trait.Eq.html

src/gtfs/read.rs
```diff
@@ -1446,6 +1454,41 @@ mod tests {
         });
     }
 
+    #[test]
+    fn gtfs_trips_no_direction_id() {
```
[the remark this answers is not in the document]
nberard: as the title says it tests the trips with no direction_id

src/gtfs/read.rs
```diff
@@ -79,19 +79,23 @@ impl From<Agency> for objects::Company {
...
-    StopArea,
-    #[serde(rename = "1")]
     StopPoint,
+    #[serde(rename = "1")]
+    StopArea,
     #[serde(rename = "2")]
     StopEntrace,
 }
 
+impl Default for StopLocationType {
```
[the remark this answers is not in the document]
nberard: 👍

src/gtfs/read.rs
```diff
@@ -79,19 +79,23 @@ impl From<Agency> for objects::Company {
     }
 }
 
-#[derive(Derivative)]
-#[derivative(Default(bound = ""))]
-#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq, Hash)]
+#[derive(Derivative, Serialize, Deserialize, Debug, Clone, PartialEq, Eq, Hash)]
```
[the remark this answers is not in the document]
nberard: 👍

src/gtfs/read.rs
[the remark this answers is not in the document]
datanel: ok but in `gtfs_trips()` there is the same fixture with no direction id.
I thought you might add an assertion.

src/gtfs/read.rs
```diff
@@ -744,10 +745,11 @@ fn get_modes_from_gtfs(
         .iter()
         .map(|mt| get_commercial_mode(mt))
         .collect();
-    let physical_modes = gtfs_mode_types
+    let physical_modes_set: HashSet<objects::PhysicalMode> = gtfs_mode_types
         .iter()
         .map(|mt| get_physical_mode(mt))
         .collect();
+    let physical_modes = physical_modes_set.into_iter().collect();
```
TeXitoi:
```rust
let physical_modes = gtfs_mode_types
    .iter()
    .map(|mt| get_physical_mode(mt))
    .collect::<BTreeSet<_>>()
    .into_iter()
    .collect();
```

BTreeSet because it will have a deterministic order.

src/gtfs/read.rs
```diff
@@ -523,7 +522,7 @@ pub fn read_stops<P: AsRef<path::Path>>(
         let comment_links = manage_comment_from_stop(comments, &stop);
         let equipment_id = get_equipment_id_and_populate_equipments(equipments, &stop);
         match stop.location_type {
-            StopLocationType::StopArea => {
+            StopLocationType::StopPoint => {
```
TeXitoi: :hankey:

src/utils.rs
```diff
@@ -173,3 +173,15 @@ where
         obj.add_prefix(prefix);
     }
 }
+
+macro_rules! skip_fail {
```
TeXitoi: I don't like this macro: it's too dependent of the context.

src/gtfs/read.rs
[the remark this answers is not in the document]
nberard: the trips file with no direction_id for a row does not make the same test as the missing column, that's why i added a new test with a missing column

src/utils.rs
nberard: should i change the name or just remove the macro?

src/gtfs/read.rs
nberard: order by id? or co2_emission?

TeXitoi: order by `Ord`

PR #102, opened 2018-06-20
Netex - read companies

prhod:
This PR is a start for reading a netex dataset, reading only companies.
I'm following @TeXitoi  's proposal to use a NetexReader object and object functions  to avoid passing the `collections` and the `context` objects.


src/netex/read.rs
```diff
@@ -15,9 +15,203 @@
...
+            context: NetexContext::default(),
+            collections: Collections::default(),
+        }
+    }
+
+    pub fn read_netex_file<R: Read>(&mut self, mut file: R) -> Result<()> {
+        let mut file_content = "".to_string();
+        file.read_to_string(&mut file_content)?;
+        let mut reader = Reader::from_str(&file_content);
+        let root = Element::from_reader(&mut reader)?;
```
datanel: you can remove 
- `let mut reader = Reader::from_str(&file_content);`
- `extern crate quick_xml;`
- `use self::quick_xml::Reader;`
- `quick_xml` in Cargo.toml

and use `let root: Element = file_content.parse()?;`

src/netex/read.rs
```diff
@@ -15,9 +15,194 @@
...
+                            self.read_composite_data_frame(frame).map_err(|_| {
+                                format_err!(
+                                    "Reading Frame id={:?}",
+                                    frame.attr("id").unwrap_or("undefined")
+                                )
+                            })
+                        })
+                        .collect()
+                },
+            )
```
TeXitoi: Wouhaou! I take some time to understand that...

Better to do that simpler:

```rust
for frame in root
    .get_child("dataObjects", self.context.namespace.as_str())
    .ok_or_else(|| bail!("Netex file does't contain a 'dataObjects' node"))?
    .filter(|frame| frame.name() == "CompositeFrame")
{
    self.read_composite_data_frame(frame)
        .with_context(|_| format!("Reading Frame id={:?}", frame.attr("id").unwrap_or("undefined")))?
}
Ok(())
```

src/netex/read.rs
```diff
@@ -15,9 +15,194 @@
...
+                            // "SiteFrame" => self.read_site_frame(&frame),
+                            // "ServiceFrame" => self.read_service_frame(&frame),
+                            // "ServiceCalendarFrame" => self.read_service_calendar_frame(&frame),
+                            // "TimetableFrame" => self.read_time_table_frame(&frame),
+                            "ResourceFrame" => self.read_resource_frame(&frame),
+                            _ => Ok(()),
+                        })
+                        .collect()
+                },
+            )
```
TeXitoi: same transformation

src/netex/read.rs
```diff
@@ -15,9 +15,194 @@
...
+                    id: id.to_string(),
+                    name: node
+                        .get_child("Name", &self.context.namespace)
+                        .map_or("".to_string(), |n| n.text().to_string()),
+                    ..Default::default()
+                }),
+                _ => bail!("An 'Operator' node doesn't have an 'id' property."),
+            })
+            .collect();
+        let companies = companies?;
```
TeXitoi: You can do these 2 let statment in one time:
```rust
let companies = organisations
    // ...
    .collect::<Result<Vec<_>>>()?;

File: src/netex/read.rs [hunk 5]
```diff
@@ -15,9 +15,194 @@
...
+        if !companies.is_empty() {
+            self.context.first_operator_id = companies[0].id.to_string();
+            let mut companies: Vec<objects::Company> = companies
+                .into_iter()
+                .filter(|c| self.collections.companies.get_idx(&c.id).is_none())
+                .collect();
+            self.collections.companies.append(&mut companies)?;
```
TeXitoi: Can simply be written like that (no need to append)
```rust
for c in companies {
    if self.collections.companies.get_idx(&c.id).is_none() {
        self.collections.companies.push(c)?;
    }
}
```

src/netex/read.rs
datanel: @TeXitoi I take some time to understand your code and I still don't understand

TeXitoi: @datanel you prefer the original version? you have a better idea?

datanel: I understand now :wink: 
@prhod don't forget `use failure::ResultExt;`

datanel:
@TeXitoi done

prhod:
thanks :+1:

PR #103, opened 2018-07-16
fix fmt again and again

datanel:


TeXitoi:
You should change the fmt to use stable.

PR #109, opened 2018-08-24
Basic generation of transfers

datanel:

### DIFF
--- a/src/transfers.rs
+++ b/src/transfers.rs
@@ -17,14 +17,23 @@
 //! [NTFS](https://github.com/CanalTP/navitia/blob/dev/documentation/ntfs/ntfs_fr.md)
 //! format management.
 
-use collection::{Collection, CollectionWithId};
+use collection::{Collection, CollectionWithId, Idx};
 use objects::{StopPoint, Transfer};
+use std::collections::HashSet;
 
-/// Checks if a transfer already exist between 2 stop points
-fn transfer_exists(transfers: &Collection<Transfer>, sp1: &StopPoint, sp2: &StopPoint) -> bool {
+fn make_transfers_set(
+    transfers: &Collection<Transfer>,
+    sp: &CollectionWithId<StopPoint>,
+) -> HashSet<(Idx<StopPoint>, Idx<StopPoint>)> {
     transfers
         .values()
-        .any(|t| t.from_stop_id == sp1.id && t.to_stop_id == sp2.id)
+        .map(|t| {
+            (
+                sp.get_idx(&t.from_stop_id).unwrap(),
+                sp.get_idx(&t.to_stop_id).unwrap(),
+            )
+        })
+        .collect()
 }
 
 /// Generates missing transfers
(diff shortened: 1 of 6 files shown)


prhod:
I tried to run it against Marseilles RTM (not in release mode), but I ended to kill it after 30 min. Is there room for performance improvements ?

datanel:
@prhod  rebased

TeXitoi:
@prhod use release mode, there is in general a 100x speedup.

src/transfers.rs
```diff
@@ -0,0 +1,164 @@
...
+                        from_stop_id: sp1.id.clone(),
+                        to_stop_id: sp2.id.clone(),
+                        min_transfer_time: Some(transfer_time),
+                        real_min_transfer_time: Some(transfer_time),
+                        equipment_id: None,
+                    });
+                }
+            }
+        }
+    }
```
TeXitoi: Minimizing distance computations.

```rust
let transfers_set = make_transfers_set(&transfers, &stop_points);
let sq_max_distance = max_distance * max_distance;
for (idx1, sp1) in &stop_points {
    let approx = sp1.coord.approx();
    for (_, sp2) in stop_points.values().filter(|&(idx2, _)| !transfers_set.contains((idx1, idx2))) {
        let sq_distance = approx.sq_distance_to(&sp2.coord);
        if sq_distance > sq_max_distance {
            continue;
        }
        let transfer_time = (sq_distance.sqrt() / walking_speed) as u32;
        transfers.push(Transfer {
            from_stop_id: sp1.id.clone(),
            to_stop_id: sp2.id.clone(),
            min_transfer_time: Some(transfer_time),
            real_min_transfer_time: Some(transfer_time),
            equipment_id: None,
        });
    }
}
```

src/transfers.rs
```diff
@@ -0,0 +1,164 @@
...
+        //we keep the existing transfers
+        assert_eq!(transfers[0].min_transfer_time, Some(50));
+        assert_eq!(transfers[0].real_min_transfer_time, Some(60));
+        assert_eq!(transfers[1].min_transfer_time, Some(200));
+        assert_eq!(transfers[1].real_min_transfer_time, Some(210));
+
+        // transfers sp_2 -> sp_3, sp_3 -> sp_2, sp_3 -> sp_1 are not added,
+        // because distances between them are > 100m
+        // sp_1 -> sp_3 is kept because it is an existing transfer.
+        assert_eq!(6, transfers.len());
```
TeXitoi: You should also check the durations of the generated transfers.

src/transfers.rs
```diff
@@ -0,0 +1,164 @@
...
+        for sp2 in stop_points.values() {
+            if !transfer_exists(&transfers, &sp1, &sp2) {
+                let distance = sp1.coord.distance_to(&sp2.coord);
+                let transfer_time = (distance / walking_speed) as u32;
+                if distance <= max_distance {
+                    transfers.push(Transfer {
+                        from_stop_id: sp1.id.clone(),
+                        to_stop_id: sp2.id.clone(),
+                        min_transfer_time: Some(transfer_time),
+                        real_min_transfer_time: Some(transfer_time),
```
TeXitoi: Don't we want different values here?

src/transfers.rs
TeXitoi: maybe just filtering by distance first before filtering with existing transfers would be enough.

PR #110, opened 2018-08-26
Fix depreciation: Error:causes

Tristramg:
https://docs.rs/failure/0.1.2/failure/struct.Error.html#method.causes


prhod:
thanks :+1:

PR #111, opened 2018-08-27
Use the gtfs-structures crate for Stops

Tristramg:


datanel:
you need to apply `cargo fmt`

```
rustu update
rustup component add rustfmt-preview
cargo fmt --all
```

prhod:
@Tristramg can you apply fmt to get CI passing ? I to am in favour or a GTFS external crate, but we may need to improve it by ourselves. Any idea on how we can go forward together ?

Tristramg:
Hello, sorry for being slow to reply. The formatting is done.

We could create an organization where you would have writing rights, or I could just give you rights to the existing repository?

datanel:
@Tristramg an organization seems good

prhod:
@Tristramg we discussed the subject with @TeXitoi @kinnou02 and @datanel. We need to agree with the crate scope, who can merge PR and how we want to go forward on the subject. @TeXitoi will lead the subject our side.
This PR way stay like this a little, sorry for asking a fix too soon :p

Tristramg:
https://gist.github.com/Tristramg/c52dd25ae5d23b19c6744fa33aa6c5da here is a suggestion how to handle this whole thing

datanel:
your tests fail because we remove "/" in some ids when deserializing but in your crate, you surely don't do that.
We have problems with object containing "/" in ids when loading in navitia.
We know this is not the correct way to do that and we have to discuss for a better solution

Tristramg:
yep, we could add it as an option. Otherwise it would be a surprise for a third party user.

But for now I have problems with

```
   Compiling navitia_model v0.1.0 (/home/tristram/beta.gouv.fr/navitia_model)                     
error: proc-macro derive panicked                                                                 
  --> src/bin/merge-ntfs.rs:33:17                                                                 
   |                                                                                              
33 | #[derive(Debug, StructOpt)]                                                                  
   |                 ^^^^^^^^^                                                                    
   |                                                                                              
   = help: message: unsupported option: short  
```

So I’ll be back some time later when I’ll take the time to figure it out…

TeXitoi:
@Tristramg `cargo update`

antoine-de:
changing the id at deserialization does not seems very neat.

I see 2 way of doing it:

1. we can do it while transforming a GTFS to a `Model` here: https://github.com/CanalTP/navitia_model/blob/master/src/gtfs/read.rs#L89
2. we can add a `sanitize` step in `gtfs2ntfs`

1. is the easiest and most straightforward solution. It makes it however a bit difficult to configure the `sanitize`.

I think 1. is fine for the moment.
What do you think ? (maybe we can open a separate issue on this? )

Tristramg:
I’m closing this for now. It pulls too many changes. We will try to make smaller incremental changes

PR #113, opened 2018-09-11
add merge binary for ntfs

nberard:
to test, run: 
```
cargo run --bin merge fixtures/minimal_ntfs fixtures/merge -o output
```

### DIFF
--- a/src/collection.rs
+++ b/src/collection.rs
@@ -200,6 +200,30 @@ impl<T> Collection<T> {
         self.objects.push(item);
         Idx::new(next_index)
     }
+
+    /// Merge a `Collection` parameter into the current one.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// # use navitia_model::collection::*;
+    /// # fn run() -> navitia_model::Result<()> {
+    /// # #[derive(PartialEq, Debug)] struct Obj(&'static str);
+    /// # impl Id<Obj> for Obj { fn id(&self) -> &str { self.0 } }
+    /// let mut c1 = Collection::new(vec![Obj("foo")]);
+    /// let c2 = Collection::new(vec![Obj("bar")]);
+    /// c1.merge(c2);
+    /// assert_eq!(c1.len(), 2);
+    /// # Ok(())
+    /// # }
+    /// # fn main() { run().unwrap() }
+    /// ```
+    pub fn merge(&mut self, other: Self) -> Result<()> {
+        for item in other {
+            self.push(item);
+        }
+        Ok(())
+    }
 }
 
 /// The type returned by `Collection::iter`.
@@ -212,6 +236,15 @@ impl<'a, T> IntoIterator for &'a Collection<T> {
 
     fn into_iter(self) -> Iter<'a, T> {
         self.iter()
+    }
+}
+
+impl<T> IntoIterator for Collection<T> {
+    type Item = T;
+    type IntoIter = ::std::vec::IntoIter<T>;
+
+    fn into_iter(self) -> Self::IntoIter {
+        self.objects.into_iter()
     }
 }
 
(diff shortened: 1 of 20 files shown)


src/collection.rs
```diff
@@ -525,11 +585,20 @@ impl<'a, T> IntoIterator for &'a CollectionWithId<T> {
...
         self.iter()
     }
 }
 
+impl<T> IntoIterator for CollectionWithId<T> {
+    type Item = T;
+    type IntoIter = ::std::vec::IntoIter<T>;
+
+    fn into_iter(self) -> Self::IntoIter {
+        self.collection.objects.into_iter()
```
datanel: `self.collection.into_iter()` ?

src/collection.rs
```diff
@@ -387,6 +420,33 @@ impl<T: Id<T>> CollectionWithId<T> {
...
+    /// assert_eq!(c1.len(), 4);
+    /// # Ok(())
+    /// # }
+    /// # fn main() { run().unwrap() }
+    /// ```
+    pub fn merge(&mut self, other: Self) -> Result<()> {
+        for item in other {
+            self.push(item)?;
+        }
+        Ok(())
```
TeXitoi:
```rust
pub fn merge(&mut self, other: Self) -> Result<()> {
    other.into_iter().map(|item| self.push(item)).collect()
}
```

src/model.rs
```diff
@@ -50,6 +50,33 @@ pub struct Collections {
...
+        self.calendars.merge(c.calendars)?;
+        self.companies.merge(c.companies)?;
+        self.comments.merge(c.comments)?;
+        self.equipments.merge(c.equipments)?;
+        self.transfers.merge(c.transfers)?;
+        self.trip_properties.merge(c.trip_properties)?;
+        self.geometries.merge(c.geometries)?;
+        self.admin_stations.merge(c.admin_stations)?;
+        Ok(())
+    }
```
TeXitoi: I'm afraid we forget a field. A way to disable it, but really verbose:
```rust
pub fn merge(&mut self, c: Collections) -> Result<()> { 
    let Collections {
        contributors,
        datasets,
        // ...
    } = c;
    self.contributors.merge(contributors)?; 
    self.datasets.merge(datasets)?; 
    // ...
    Ok(())
}
```

src/collection.rs
nberard: hum i have this error when i try it:
```
   --> src/collection.rs:445:55
    |
445 |         other.into_iter().map(|item| self.push(item)).collect()
    |                                                       ^^^^^^^ a collection of type `()` cannot be built from an iterator over elements of type `collection::Idx<T>`
```

datanel: self.push(item) returns an Idx not ()

TeXitoi: proposed code modified

src/model.rs
```diff
@@ -54,6 +54,27 @@ impl Collections {
...
+            feed_infos: _,
+            calendars: _,
+            companies: _,
+            comments: _,
+            equipments: _,
+            transfers: _,
+            trip_properties: _,
+            geometries: _,
+            admin_stations: _,
+        } = c;
```
TeXitoi: No, the idea is to destructure `c` into variables, so no `: _`. It will warning on unused variables. Then you use these variables (removing `c.`). If you don't match all the field, it will not compile, if you don't use all the fields, it will warn

src/collection.rs
datanel: `for` is easier to understand

TeXitoi: As you wish, that's a non blocking comment.

src/model.rs
TeXitoi: That's the hole point, after the let, you don't use `c` anymore, you use `calendars` that is defined by the let.

nberard: aaaaaaaaaaah ok

PR #119, opened 2018-09-26
Manage transfer exceptions

datanel:
Add and/or remove transfer from a csv files that contain

from_stop_id|to_stop_id|transfer_time| |
--|--|--|--
SP1|SP2| | no time specified, will be removed
SP3|SP2|111 | added
UNKNOWN|SP2|454 | ignored as UNKNOWN is an unknown stop   
UNKNOWN|SP2|| ignored as UNKNOWN is an unknown stop

### DIFF
--- a/src/transfers.rs
+++ b/src/transfers.rs
@@ -18,21 +18,182 @@
 //! format management.
 
 use collection::{Collection, CollectionWithId, Idx};
+use csv;
+use failure::ResultExt;
 use objects::{StopPoint, Transfer};
-use std::collections::HashSet;
+use std::collections::{HashMap, HashSet};
+use std::path::Path;
+use Result;
 
-fn make_transfers_set(
-    transfers: &Collection<Transfer>,
-    sp: &CollectionWithId<StopPoint>,
-) -> HashSet<(Idx<StopPoint>, Idx<StopPoint>)> {
+#[derive(Deserialize, Debug)]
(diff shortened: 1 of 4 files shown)


src/transfers.rs
```diff
@@ -18,62 +18,215 @@
...
+}
+
+/// Generates missing transfers
+///
+/// The `max_distance` argument allows you to specify the max distance
+/// in meters to compute the tranfer.
+///
+/// The `walking_speed` argument is the walking speed in meters per second.
+///
+/// The `waiting_time` argument is the waiting transfer_time in seconds at stop.
```
datanel: missing doc for rules

datanel:
I wonder if I'd rather make integration tests with fixtures instead of unit tests

src/transfers.rs
```diff
@@ -18,62 +18,230 @@
...
+        walking_speed,
+        waiting_time,
+    );
+
+    if !rules.is_empty() {
+        remove_unwanted_transfers(&mut transfers_map, stop_points, rules);
+        add_missing_transfers(&mut transfers_map, stop_points, rules, waiting_time);
+    }
+
+    let mut transfers: Vec<Transfer> = transfers_map.values().cloned().collect();
```
TeXitoi: You can `into_iter` and avoid the `cloned`

PR #122, opened 2018-09-28
Fix shape without coordinates

datanel:
Sometimes a GTFS contains shapes without coordinates  so we have the foloowinf error `"gtfs2ntfs: Error reading \"/tmp/tmpq_rtu4_v/shapes.txt\"\nCSV deserialize error: record 43 (line: 43, byte: 1304): field 1: cannot parse float from empty string"`

This PR fix that ignoring shapes with empty points but  breaksthe GTFS model

I might match the corresponding error when deserializing a shape.
If you have an idea.

### DIFF
--- a/src/gtfs/read.rs
+++ b/src/gtfs/read.rs
@@ -350,7 +350,7 @@ struct StopTime {
 }
 
 #[derive(Deserialize, Debug)]
-pub struct Shape {
+struct Shape {
     #[serde(rename = "shape_id")]
     id: String,
     #[serde(rename = "shape_pt_lat")]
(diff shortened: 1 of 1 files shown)


datanel:
the rust error is 
```
Error(Deserialize {
    pos: Some(Position {
        byte: 89,
        line: 5,
        record: 4
    }),
    err: DeserializeError {
        field: Some(1),
        kind: ParseFloat(ParseFloatError {
            kind: Empty
        })
    }
})
```
I can say `if (field == Some(1) or Some(2)) and ParseFloatError == kind Empty`
but I can't the the `kind` value of `ParseFloatError`
https://doc.rust-lang.org/nightly/core/num/struct.ParseFloatError.html

PR #124, opened 2018-10-02
Do not create files when optional collection is empty

datanel:


src/ntfs/write.rs
```diff
@@ -285,6 +294,22 @@ where
 }
 
 pub fn write_codes(path: &path::Path, collections: &Collections) -> Result<()> {
+    fn collection_has_codes<T: Codes>(collection: &CollectionWithId<T>) -> bool {
```
TeXitoi: bad name, it should be `collection_has_no_codes`, or negate the function.

src/ntfs/write.rs
```diff
@@ -326,6 +351,18 @@ where
 }
 
 pub fn write_object_properties(path: &path::Path, collections: &Collections) -> Result<()> {
+    fn collection_has_object_properties<T: Properties>(collection: &CollectionWithId<T>) -> bool {
```
TeXitoi: Same here

TeXitoi:
You now need to rebase :stuck_out_tongue:

datanel:
arf

PR #126, opened 2018-10-08
Fix relations networks<->lines and companies<->vj with routes without agency id

datanel:
fix `gtfs2ntfs: Error indexing networks_to_lines\nid=\"KSM:default_agency_id\" not found`

and same error with `companies_to_vehicle_journeys`

### DIFF
--- a/src/gtfs/read.rs
+++ b/src/gtfs/read.rs
@@ -35,6 +35,17 @@ fn default_agency_id() -> String {
     "default_agency_id".to_string()
 }
 
+fn get_agency_id(route: &Route, networks: &CollectionWithId<objects::Network>) -> Result<String> {
+    route
+        .agency_id
+        .as_ref()
+        .map(|id| Ok(id.to_string()))
+        .unwrap_or_else(|| match networks.values().next() {
+            Some(n) => Ok(n.id.clone()),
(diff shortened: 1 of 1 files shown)


src/gtfs/read.rs
```diff
@@ -35,6 +35,17 @@ fn default_agency_id() -> String {
     "default_agency_id".to_string()
 }
 
+fn get_agency_id(route: &Route, networks: &CollectionWithId<objects::Network>) -> Result<String> {
+    route
+        .agency_id
+        .as_ref()
+        .map(|id| Ok(id.to_string()))
+        .unwrap_or_else(|| match networks.values().next() {
```
TeXitoi: https://doc.rust-lang.org/stable/std/option/enum.Option.html#method.ok_or_else

src/gtfs/read.rs
```diff
@@ -35,6 +35,17 @@ fn default_agency_id() -> String {
...
+            Some(n) => Ok(n.id.clone()),
```
TeXitoi: I don't like that: it means "if there is no agency_id, use the first one", but I think it should be "if there is no agency_id and we have only one agency, use the first one".

@prhod ?

prhod: Agreed with @TeXitoi about the extact rule

PR #127, opened 2018-10-09
[ntfs2gtfs] write agency.txt

datanel:


src/gtfs/mod.rs
```diff
@@ -23,11 +23,46 @@ use collection::CollectionWithId;
...
+impl From<objects::Network> for Agency {
+    fn from(n: objects::Network) -> Agency {
+        Agency {
+            id: Some(n.id.clone()),
+            name: n.name.clone(),
+            url: n
+                .url
+                .clone()
+                .unwrap_or("http://www.navitia.io/".to_string()),
+            timezone: n.timezone.clone().or(Some("Europe/Paris".to_string())),
```
TeXitoi: Are you sure of these defaults? @prhod ?

src/gtfs/mod.rs
```diff
@@ -23,11 +23,46 @@ use collection::CollectionWithId;
...
+    phone: Option<String>,
+    #[serde(rename = "agency_email")]
+    email: Option<String>,
+}
+
+impl From<objects::Network> for Agency {
+    fn from(n: objects::Network) -> Agency {
+        Agency {
+            id: Some(n.id.clone()),
+            name: n.name.clone(),
```
TeXitoi: All these clones should be useless as you have the ownership of a Network.

src/gtfs/write.rs
```diff
@@ -13,3 +13,95 @@
...
+            id: Some(n.id.clone()),
+            name: n.name.clone(),
+            url: n
+                .url
+                .clone()
+                .unwrap_or("http://www.navitia.io/".to_string()),
+            timezone: n.timezone.clone().or(Some("Europe/Paris".to_string())),
+            lang: n.lang.clone(),
+            phone: n.phone.clone(),
+            email: None,
```
TeXitoi: You should use the from implementation here.

src/gtfs/mod.rs
datanel: I have 
```
error[E0507]: cannot move out of borrowed content
  --> src/gtfs/write.rs:31:22
   |
31 |             id: Some(n.id),
   |                      ^ cannot move out of borrowed content
```

src/gtfs/mod.rs
prhod: These are the actual defaults for Navitia.io, but they shouldn't be used in most cases.
Do you have a better idea ?

src/gtfs/mod.rs
```diff
@@ -23,11 +23,46 @@ use collection::CollectionWithId;
...
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
+struct Agency {
+    #[serde(rename = "agency_id")]
+    id: Option<String>,
+    #[serde(rename = "agency_name")]
+    name: String,
+    #[serde(rename = "agency_url")]
+    url: String,
+    #[serde(rename = "agency_timezone")]
+    timezone: Option<String>,
```
TeXitoi: Must be a String, not an option according to the spec.

PR #128, opened 2018-10-11
[ntfs2gtfs] write stops.txt

datanel:
see https://github.com/CanalTP/navitia_model/blob/master/src/documentation/ntfs2gtfs.md#stopstxt

### DIFF
--- a/src/gtfs/mod.rs
+++ b/src/gtfs/mod.rs
@@ -120,7 +120,7 @@ impl<'a> From<&'a objects::StopPoint> for Stop {
             desc: "".to_string(),
             wheelchair_boarding: None,
             url: None,
-            timezone: None,
+            timezone: obj.timezone.clone(),
         }
     }
 }
@@ -139,7 +139,7 @@ impl<'a> From<&'a objects::StopArea> for Stop {
             desc: "".to_string(),
             wheelchair_boarding: None,
             url: None,
-            timezone: None,
+            timezone: obj.timezone.clone(),
         }
     }
 }
--- a/src/gtfs/mod.rs
+++ b/src/gtfs/mod.rs
@@ -26,6 +26,7 @@ use model::{Collections, Model};
 use objects;
 use read_utils::add_prefix;
 use std::path::Path;
+use utils::*;
 use Result;
 
 #[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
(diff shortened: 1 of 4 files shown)


src/gtfs/mod.rs
```diff
@@ -47,25 +48,102 @@ struct Agency {
...
+            lat: obj.coord.lat,
+            lon: obj.coord.lon,
+            fare_zone_id: None,
+            location_type: StopLocationType::StopArea,
+            parent_station: None,
+            code: None,
+            desc: "".to_string(),
+            wheelchair_boarding: None,
+            url: None,
+            timezone: None,
```
TeXitoi: You forgot timezone

src/gtfs/mod.rs
```diff
@@ -47,25 +48,102 @@ struct Agency {
...
+    fn from(obj: &objects::StopArea) -> Stop {
+        Stop {
+            id: obj.id.clone(),
+            name: obj.name.clone(),
+            lat: obj.coord.lat,
+            lon: obj.coord.lon,
+            fare_zone_id: None,
+            location_type: StopLocationType::StopArea,
+            parent_station: None,
+            code: None,
```
TeXitoi: you should take it from codes, no?

src/gtfs/mod.rs
```diff
@@ -47,25 +48,102 @@ struct Agency {
...
+            lat: obj.coord.lat,
+            lon: obj.coord.lon,
+            fare_zone_id: obj.fare_zone_id.clone(),
+            location_type: StopLocationType::StopPoint,
+            parent_station: Some(obj.stop_area_id.clone()),
+            code: None,
+            desc: "".to_string(),
+            wheelchair_boarding: None,
+            url: None,
+            timezone: None,
```
TeXitoi: You forgot timezone

src/gtfs/mod.rs
```diff
@@ -47,25 +48,102 @@ struct Agency {
...
+    fn from(obj: &objects::StopPoint) -> Stop {
+        Stop {
+            id: obj.id.clone(),
+            name: obj.name.clone(),
+            lat: obj.coord.lat,
+            lon: obj.coord.lon,
+            fare_zone_id: obj.fare_zone_id.clone(),
+            location_type: StopLocationType::StopPoint,
+            parent_station: Some(obj.stop_area_id.clone()),
+            code: None,
```
TeXitoi: code from codes?

src/gtfs/mod.rs
prhod: We do not read it for the moment. I've added this issue for this : https://github.com/CanalTP/navitia_model/issues/130

src/gtfs/mod.rs
prhod: same as above :p

PR #129, opened 2018-10-12
[ntfs2gtfs] Write routes.txt

datanel:
see https://github.com/CanalTP/navitia_model/blob/master/src/documentation/ntfs2gtfs.md#routestxt

### DIFF
--- a/src/gtfs/write.rs
+++ b/src/gtfs/write.rs
@@ -14,8 +14,11 @@
 // along with this program.  If not, see
 // <http://www.gnu.org/licenses/>.
 
-use super::{Agency, DirectionType, Route, RouteType, Shape, Stop, StopLocationType, StopTime, Transfer, Trip};
-use collection::{Collection, CollectionWithId, Id};
+use super::{
+    Agency, DirectionType, Route, RouteType, Shape, Stop, StopLocationType, StopTime, Transfer,
+    Trip,
+};
+use collection::{Collection, CollectionWithId, Id, Idx};
 use common_format::Availability;
 use csv;
 use failure::ResultExt;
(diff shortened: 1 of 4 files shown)


src/gtfs/write.rs
```diff
@@ -200,10 +202,78 @@ pub fn write_trips(
...
+            _ => RouteType::Other(3),
+        }
+    }
+}
+
+fn get_gtfs_route_id_from_ntfs_line_id(line_id: &str, pm: &objects::PhysicalMode) -> String {
+    match pm.id.as_str() {
+        "RailShuttle" | "Tramway" | "Metro" | "LocalTrain" | "LongDistanceTrain"
+        | "RapidTransit" | "Train" | "Other (or unknown)" | "Bus" | "BusRapidTransit" | "Coach"
+        | "Boat" | "Ferry" | "Funicular" | "Shuttle" => line_id.to_string(),
```
TeXitoi: If you have a line with, as physical_modes, Bus and RapidTransit, you will generate 2 routes with the same id.

TeXitoi: looks like a spec bug

TeXitoi: @prhod

src/gtfs/write.rs
```diff
@@ -282,6 +287,108 @@ pub fn write_stop_extensions(
...
+
+impl<'a> From<&'a objects::PhysicalMode> for RouteType {
+    fn from(obj: &objects::PhysicalMode) -> RouteType {
+        match obj.id.as_str() {
+            "RailShuttle" | "Tramway" => RouteType::Tramway_LightRail,
+            "Metro" => RouteType::Metro,
+            "LocalTrain" | "LongDistanceTrain" | "RapidTransit" | "Train" => RouteType::Rail,
+            "Bus" | "BusRapidTransit" | "Coach" => RouteType::Bus,
+            "Boat" | "Ferry" => RouteType::Ferry,
+            "Funicular" | "Shuttle" => RouteType::Funicular,
```
kinnou02: minor: `shuttle` is a Funicular? seems strange to me

src/gtfs/write.rs
```diff
@@ -278,6 +283,108 @@ pub fn write_stop_extensions(
...
+        .enumerate()
+        .map(|(i, pm)| PhysicalModeWithOrder {
+            inner: pm,
+            is_lowest: i == 0,
+        }).collect()
+}
+
+impl<'a> From<&'a objects::PhysicalMode> for RouteType {
+    fn from(obj: &objects::PhysicalMode) -> RouteType {
+        match obj.id.as_str() {
```
pbougue: @prhod questions are linked to what we decided here : https://jira.kisio.org/browse/NAVP-1007
1. Looks like nothing maps to RouteType::CableCar (and I'm curious of the opposite also : cable car should be mapped to funicular?).
2. Same question for Gondola_SuspendedCableCar (and we mentionned creation of SuspendedCableCar in NTFS for that)

Also, some types are missing (Taxi and Air if I'm correct).

It would be nice to have the reverse mapping done also here (even if not used) to compare.

src/gtfs/write.rs
```diff
@@ -278,6 +283,108 @@ pub fn write_stop_extensions(
...
+    if pm.is_lowest {
+        line_id.to_string()
+    } else {
+        line_id.to_string() + ":" + &pm.inner.id
+    }
+}
+
+fn get_physical_mode_order(pm: &objects::PhysicalMode) -> u8 {
+    match pm.id.as_str() {
+        "Tramway" => 1,
```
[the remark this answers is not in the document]
pbougue: OK fine.
Can be nice to comment or do it explicitly (but don't bother if I'm the only one with that opinion).
(Same below.)

src/gtfs/write.rs
pbougue: After discussion with @prhod on the 2 questions I had:
1. GTFS type 5 (CableCar) is to be mapped to Funicular in NTFS (so nothing to do here, maybe some doc)
2. Gondola_SuspendedCableCar will be mapped to SuspendedCableCar in NTFS, and it will have to be managed here, but not urgent.

Everything is tracked in Nav2's JIRA ticket https://jira.kisio.org/browse/NAVP-1007

PR #131, opened 2018-10-16
[ntfs2gtfs] write transfers.txt

nberard:
see https://github.com/CanalTP/navitia_model/blob/master/src/documentation/ntfs2gtfs.md#transferstxt
fix https://github.com/CanalTP/tartare/issues/746

### DIFF
--- a/src/gtfs/mod.rs
+++ b/src/gtfs/mod.rs
@@ -141,6 +141,40 @@ struct Trip {
     bikes_allowed: Availability,
 }
 
+#[derive(Serialize, Deserialize, Debug, Derivative, PartialEq)]
+#[derivative(Default)]
+enum TransferType {
+    #[derivative(Default)]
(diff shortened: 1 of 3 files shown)


src/gtfs/mod.rs
```diff
@@ -141,6 +141,40 @@ struct Trip {
...
+impl<'a> From<&'a objects::Transfer> for Transfer {
+    fn from(obj: &objects::Transfer) -> Transfer {
+        Transfer {
+            from_stop_id: obj.from_stop_id.clone(),
+            to_stop_id: obj.to_stop_id.clone(),
+            transfer_type: TransferType::WithTransferTime,
+            min_transfer_time: obj.min_transfer_time,
+        }
+    }
+}
```
datanel: this should be in write.rs imho

PR #134, opened 2018-10-17
[ntfs2gtfs] write shapes.txt

datanel:
see https://github.com/CanalTP/navitia_model/blob/master/src/documentation/ntfs2gtfs.md#shapestxt

### DIFF
--- a/src/gtfs/write.rs
+++ b/src/gtfs/write.rs
@@ -14,11 +14,12 @@
 // along with this program.  If not, see
 // <http://www.gnu.org/licenses/>.
 
-use super::{Agency, DirectionType, Stop, StopLocationType, StopTime, Transfer, Trip};
+use super::{Agency, DirectionType, Shape, Stop, StopLocationType, StopTime, Transfer, Trip};
 use collection::{Collection, CollectionWithId, Id};
 use common_format::Availability;
 use csv;
 use failure::ResultExt;
+use geo_types::Geometry as GeoGeometry;
 use objects;
 use objects::Transfer as NtfsTransfer;
 use objects::*;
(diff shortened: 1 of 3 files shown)


src/gtfs/write.rs
```diff
@@ -146,9 +147,55 @@ pub fn write_trips(
     Ok(())
 }
 
+fn ntfs_geometry_to_gtfs_shapes(g: &objects::Geometry) -> Vec<Shape> {
```
TeXitoi: You can try to return `impl Iterator<Item=Shape>`, but it will be a bit tricky to have the same return type in the 2 cases.

A solution would be to extract the `&LineString` with using an empty one in the other case, and then `points().enumerate()`...

src/gtfs/write.rs
```diff
@@ -146,9 +147,55 @@ pub fn write_trips(
...
+
+pub fn write_shapes(
+    path: &path::Path,
+    geometries: &CollectionWithId<objects::Geometry>,
+) -> Result<()> {
+    let mut shapes: Vec<Shape> = vec![];
+    for g in geometries.values() {
+        for s in ntfs_geometry_to_gtfs_shapes(g) {
+            shapes.push(s);
+        }
```
TeXitoi:
```rust
shapes.extend(ntfs_geometry_to_gtfs_shapes(g));
```

src/gtfs/write.rs
```diff
@@ -200,9 +201,52 @@ pub fn write_trips(
...
+}
+
+pub fn write_shapes(
+    path: &path::Path,
+    geometries: &CollectionWithId<objects::Geometry>,
+) -> Result<()> {
+    let mut shapes: Vec<Shape> = vec![];
+    for g in geometries.values() {
+        shapes.extend(ntfs_geometry_to_gtfs_shapes(g));
+    }
```
TeXitoi: let shapes: Vec<_> = geometries.values().flat_map(ntfs_geometry_to_gtfs_shapes).collect();

PR #136, opened 2018-10-17
[ntfs2gtfs] stops: export desc

datanel:


src/gtfs/write.rs
```diff
@@ -36,20 +36,70 @@ pub fn write_agencies(path: &path::Path, networks: &CollectionWithId<Network>) -
...
+fn get_first_comment_name<T: CommentLinks>(
+    obj: &T,
+    comments: &CollectionWithId<Comment>,
+) -> String {
+    let mut desc = "".to_string();
+    if !obj.comment_links().is_empty() {
+        let mut obj_comments: Vec<&Comment> = comments.iter_from(obj.comment_links()).collect();
+        obj_comments.sort_unstable_by_key(|c| &c.name);
+        desc = obj_comments.iter().next().unwrap().name.clone();
+    }
```
datanel: @TeXitoi I'm sure you have a better way to do that

TeXitoi:
```rust
desc = comments
    .iter_from(obj.comment_links())
    .map(|c| &c.name)
    .min()
    .cloned()
    .unwrap_or("".into());
```
Or something like that.

PR #141, opened 2018-10-31
use route.line_id instead trip.route_id

nberard:
fix https://github.com/CanalTP/tartare/issues/782

### DIFF
--- a/src/gtfs/write.rs
+++ b/src/gtfs/write.rs
@@ -176,11 +176,7 @@ fn get_gtfs_trip_shortname_and_headsign_from_ntfs_vj(
     }
 }
 
-fn get_gtfs_direction_id_from_ntfs_vj(
-    vj: &objects::VehicleJourney,
-    routes: &CollectionWithId<objects::Route>,
-) -> DirectionType {
-    let route = routes.get(&vj.route_id).unwrap();
+fn get_gtfs_direction_id_from_ntfs_route(route: &objects::Route) -> DirectionType {
     match route.direction_type.as_ref().map(|s| s.as_str()) {
         Some("forward") | Some("clockwise") | Some("inbound") => DirectionType::Forward,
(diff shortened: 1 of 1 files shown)


src/documentation/ntfs2gtfs.md
```diff
@@ -90,7 +90,7 @@ stop_desc | no | comments.txt, comment_links.txt | comment_name | The value of `
 
 GTFS field | Required | NTFS file | NTFS field | Note
 --- | --- | --- | --- | ---
-route_id | yes | trips.txt | route_id |
+route_id | yes | trips.txt and routes.txt | line_id corresponding to the route_id |
```
prhod: The specs have been updated : https://github.com/CanalTP/navitia_model/blob/master/src/documentation/ntfs2gtfs.md#routestxt
I think this modification is not necessary

src/gtfs/read.rs
```diff
@@ -751,8 +751,12 @@ fn make_routes(gtfs_trips: &[Trip], map_line_routes: &MapLineRoutes) -> Vec<obje
         let sr = get_route_with_smallest_name(rs);
         for r in rs {
             let mut route_directions: HashSet<&DirectionType> = HashSet::new();
+            let mut geometry_id = None;
             for t in gtfs_trips.iter().filter(|t| t.route_id == r.id) {
                 route_directions.insert(&t.direction);
+                if t.shape_id.is_some() && geometry_id.is_none() {
+                    geometry_id = t.shape_id.clone();
```
kinnou02: This change doesn't seems to be related to this PR.
If I understand correctly the geometry of the route will be the geometry of it's first vj that have a geom, thats right? 
if so, it's a approximation that will be false sometime, but it's a good first step.
Also I find it a bit strange to do this in the gtfs reader, I would have made that after the model to be independant of the format.

PR #143, opened 2018-11-02
trim location coordinates from gtfs stop points

nberard:
pl_warsaw data set contains spaces before certain longitudes

### DIFF
--- a/src/gtfs/mod.rs
+++ b/src/gtfs/mod.rs
@@ -91,9 +91,15 @@ struct Stop {
     name: String,
     #[serde(default, rename = "stop_desc")]
     desc: String,
-    #[serde(rename = "stop_lon", deserialize_with = "de_location_trim_with_default")]
+    #[serde(
+        rename = "stop_lon",
+        deserialize_with = "de_location_trim_with_default"
+    )]
     lon: f64,
-    #[serde(rename = "stop_lat", deserialize_with = "de_location_trim_with_default")]
+    #[serde(
+        rename = "stop_lat",
+        deserialize_with = "de_location_trim_with_default"
+    )]
     lat: f64,
     #[serde(rename = "zone_id")]
     fare_zone_id: Option<String>,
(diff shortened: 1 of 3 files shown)


src/utils.rs
```diff
@@ -113,6 +113,18 @@ where
...
 
+pub fn de_location_trim_with_default<'de, D>(deserializer: D) -> Result<f64, D::Error>
+    where
+        D: ::serde::Deserializer<'de>,
+{
+    use serde::Deserialize;
+    let s = String::deserialize(deserializer)?;
+    Ok(s.trim().parse::<f64>().unwrap_or_else(|e| {
+        error!("{}", e);
+        0.00
```
kinnou02: Not a fan of putting stop_point on the null island, but if it's ok for @prhod I'am ok too

src/gtfs/mod.rs
```diff
@@ -91,9 +91,9 @@ struct Stop {
     name: String,
     #[serde(default, rename = "stop_desc")]
     desc: String,
-    #[serde(rename = "stop_lon")]
+    #[serde(rename = "stop_lon", deserialize_with = "de_location_trim_with_default")]
```
datanel: you could use https://docs.rs/csv/1.0.2/csv/struct.ReaderBuilder.html#method.trim

nberard: agreed with u if we can trim all fields, what do u think @kinnou02?

src/utils.rs
nberard: not a fan too but he made me do it :p

prhod: @kinnou02  I'm open to any other solution that enable conversion (without panick)  and if possible not much refacto ^^ 
Using a blank value doesn't seem to be a good solution :/

src/gtfs/mod.rs
kinnou02: good for me

prhod: trimming both content and header seems to be a good idea, we know some GTFS have spaces in col names

PR #147, opened 2018-11-08
Write ntfs stop code

datanel:
fix https://github.com/CanalTP/navitia_model/issues/130

### DIFF
--- a/src/gtfs/write.rs
+++ b/src/gtfs/write.rs
@@ -83,6 +83,7 @@ fn ntfs_stop_point_to_gtfs_stop(
     comments: &CollectionWithId<objects::Comment>,
     equipments: &CollectionWithId<objects::Equipment>,
 ) -> Stop {
+    println!("{:#?}", sp);
     let wheelchair = sp
         .equipment_id
         .clone()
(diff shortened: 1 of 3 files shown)


src/gtfs/write.rs
```diff
@@ -97,7 +98,7 @@ fn ntfs_stop_point_to_gtfs_stop(
         fare_zone_id: sp.fare_zone_id.clone(),
         location_type: StopLocationType::StopPoint,
         parent_station: Some(sp.stop_area_id.clone()),
-        code: None,
+        code: sp.codes.iter().next().map(|c| c.1.clone()),
```
datanel: @prhod is it correct or I need to get the first `gtfs_stop_code` code?

src/gtfs/write.rs
```diff
@@ -83,6 +83,7 @@ fn ntfs_stop_point_to_gtfs_stop(
     comments: &CollectionWithId<objects::Comment>,
     equipments: &CollectionWithId<objects::Equipment>,
 ) -> Stop {
+    println!("{:#?}", sp);
```
nberard: to be removed

src/gtfs/write.rs
prhod: If I read this correctly, you write the first code of the stop, even if it's not a `gtfs_stop_code`. 
This can be a problem for later. 
Can you check the `object_system` value ?
Thanks !

PR #148, opened 2018-11-09
remove slashes from gtfs import on ids

nberard:
fix https://github.com/CanalTP/tartare/issues/799


src/utils.rs
```diff
@@ -125,6 +125,24 @@ where
...
+    Ok(s.replace("/", ""))
+}
+
+pub fn de_option_without_slashes<'de, D>(de: D) -> Result<Option<String>, D::Error>
+where
+    D: ::serde::Deserializer<'de>,
+{
+    use serde::Deserialize;
+    let option = Option::<String>::deserialize(de)?;
+    Ok(option.and_then(|s| Some(s.replace("/", ""))))
```
datanel:
```rust
Ok(option.map(|s| s.replace("/", "")))
```
I think you can use `de_without_slashes` in `de_option_without_slashes` or the other way round

PR #149, opened 2018-11-12
handle stop.location_type = 2 for ntfs

### LINKED ISSUE #125: [NTFS] Handling of stop.location_type > 1
currently stops with a location_type equals to 2 or 3 will be ignored.
Reading of the whole ntfs will be aborted if there is a stop_time that use this stop.

This also open the door to the whole ODT thing...

nberard:
fix https://github.com/CanalTP/tartare/issues/760 
fix #125


datanel:
@nberard you have to rebase

datanel:
this will not work if a stop_time references a stop_zone as we don't get the stop zone when reading a stop time

nberard:
up *_*

PR #153, opened 2018-11-16
Issue 750 coverage merge with transfers

nberard:
https://jira.kisio.org/browse/ND-63


src/transfers.rs
```diff
@@ -30,45 +36,132 @@ struct Rule {
...
+    transfers_mode: &TransfersMode,
+    report_opt: Option<&mut Report>,
+) -> bool {
+    let from_contributor: BTreeSet<Idx<Contributor>> = model.get_corresponding_from_idx(from_idx);
+    let to_contributor: BTreeSet<Idx<Contributor>> = model.get_corresponding_from_idx(to_idx);
+    if from_contributor.is_empty() {
+        if let Some(report) = report_opt {
+            report.add_warning(
+                format!(
+                "stop point {} belongs to none of the trips and should not generate any transfer",
```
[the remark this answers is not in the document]
nberard: will is ok?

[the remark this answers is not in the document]
datanel: yes

src/bin/merge-ntfs.rs
```diff
@@ -38,6 +40,40 @@ struct Opt {
...
+
+    #[structopt(
+        long = "walking-speed",
+        short = "s",
+        default_value = "0.785",
+        help = "The walking speed in meters per second. \
+                You may want to divide your initial speed by \
+                sqrt(2) to simulate Manhattan distances"
+    )]
+    walking_speed: f64,
```
TeXitoi: with the latest structopt, you can

```rust
#[derive(StructOpt, Debug)]
#[rename_all = "kebab-case"]
struct Opt {
    // ...

    /// The walking speed in meters per second. You may want to divide
    /// your initial speed by sqrt(2) to simulate Manhattan distances
    #[structopt(long, short = "s", default_value = "0.785")]
    walking_speed: f64,

    // ...
}
```

src/model.rs
```diff
@@ -348,3 +348,8 @@ impl ops::Deref for Model {
         &self.collections
     }
 }
+impl ops::DerefMut for Model {
+    fn deref_mut(&mut self) -> &mut Collections {
+        &mut self.collections
+    }
+}
```
TeXitoi: No, you can't do that. The point of Model only implementing Deref is because modifying the Collections will invalidate the model.

src/transfers.rs
```diff
@@ -30,45 +36,134 @@ struct Rule {
     to_stop_id: String,
     transfer_time: Option<u32>,
 }
+/// Represents the type of transfers to generate
+#[derive(PartialEq, Debug)]
+pub enum TransfersMode {
+    /// `IntraContributor` will generate transfers between stop points belonging to the
+    /// same contributor
+    IntraContributor,
```
TeXitoi: shouldn't it be `All` for normal transferr generation?

src/transfers.rs
```diff
@@ -87,34 +182,37 @@ fn make_transfers_map(transfers: Vec<Transfer>, sp: &CollectionWithId<StopPoint>
...
         let approx = sp1.coord.approx();
-        for (idx2, sp2) in stop_points.iter() {
+        for (idx2, sp2) in model.stop_points.iter() {
             if transfers_map.contains_key(&(idx1, idx2)) {
                 continue;
             }
-            let sq_distance = approx.sq_distance_to(&sp2.coord);
-            if sq_distance > sq_max_distance {
-                continue;
+            if stop_points_need_transfer(model, idx1, idx2, transfers_mode, None) {
```
TeXitoi: what's the performance impact of this?

A continue would be more readable:

```rust
if stop_points_need_transfer(model, idx1, idx2, transfers_mode, None) {
    continue;
}
```

src/transfers.rs
```diff
@@ -216,313 +317,30 @@ fn do_generates_transfers(
...
-                Transfer {
-                    from_stop_id: "sp_2".to_string(),
-                    to_stop_id: "sp_3".to_string(),
-                    min_transfer_time: Some(100),
-                    real_min_transfer_time: Some(220),
-                    equipment_id: None
-                }
-            ]
-        );
+    model.transfers = Collection::new(new_transfers);
```
TeXitoi: This can't be done without `<Model as DerefMut>`

src/transfers.rs
nberard: for now there is no case where we want both generations, do you confirm @prhod

src/transfers.rs
nberard: the condition is stop_points_need_transfer(model, idx1, idx2, transfers_mode, None) AND sq_distance > sq_max_distance to continue

nberard: regarding performances, we have for same data set transfers generation:
- on master 4.2s with 120M ram
- on this PR 6.9s with 124M ram

src/transfers.rs
nberard: yes that's the problem, do you have an idea to do the trick?

TeXitoi: implement an `into_collections(self) -> Collections` on `Model`.

src/transfers.rs
TeXitoi: On ffr-idf? Great!

TeXitoi:
```rust
if !stop_points_need_transfer(model, idx1, idx2, transfers_mode, None) {
    continue;
}
if sq_distance > sq_max_distance {
    continue;
}
```

TeXitoi:
You need to fmt

src/transfers.rs
TeXitoi: yep, and I can't imagine it useful to only add inter and not intra

src/transfers.rs
nberard: ah ok !

src/transfers.rs
nberard: we might want to change the max distance between the two generations

src/model.rs
TeXitoi: You still need to remove this impl.

src/bin/merge-ntfs.rs
TeXitoi: Minor: as shown in the comment, you can use doc comment for the help message.

src/transfers.rs
```diff
@@ -30,48 +36,146 @@ struct Rule {
...
+                    model.stop_points[to_idx].id
+                ),
+                ReportType::TransferOnUnreferencedStop,
+            );
+        }
+        return false;
+    }
+    let same_contributor = from_contributor == to_contributor;
+    same_contributor && *transfers_mode == TransfersMode::IntraContributor
+        || !same_contributor && *transfers_mode == TransfersMode::InterContributor
```
TeXitoi:
```rust
match transfers_mode {
    TransfersMode::All => true,
    TransfersMode::IntraContributor => from_contributor == to_contributor,
    TransfersMode::InterContributor => from_contributor != to_contributor,
}
```

As all values are tested, if you modify TransfersMode, you will not have nasty bugs, you'll have a compile error.

src/transfers.rs
TeXitoi: So now, this value is never used?

src/model.rs
nberard: oups sorry :D

src/transfers.rs
nberard: :+1:

PR #163, opened 2018-11-30
manage GTFS frequencies

### LINKED ISSUE #150: Handle frequencies from NTFS or GTFS
File frequencies from NTFS or GTFS are not handled currently, neither "copied" nor exploded.

(just tracking the issue for those who might miss it)

nberard:
see https://jira.kisio.org/browse/ND-72
fix #150 

included:
2nd commit contains a fix on default value for datetime_estimated when reading a ntfs


src/gtfs/write.rs
```diff
@@ -420,6 +420,7 @@ pub fn write_stop_times(
                     drop_off_type: st.drop_off_type,
                     local_zone_id: st.local_zone_id,
                     stop_headsign: stop_times_headsigns.get(&(vj_idx, st.sequence)).cloned(),
+                    timepoint: st.datetime_estimated,
```
pbougue: From http://gtfs.org/reference#stop_timestxt :
* ` ` (empty) = `1` = exact
* `0` = estimated

So here it would do the opposite for 0 and 1, no?

nberard: nice catch mb

nberard:
@pbougue i think the test read.rs::gtfs_stop_times covers the case without timepoint, i will add another one with the column and specific values

src/gtfs/read.rs
```diff
@@ -879,6 +881,155 @@ pub fn set_dataset_validity_period(
...
+                frequency.trip_id
+            );
+            continue;
+        }
+        let mut arrival_time_delta = corresponding_vj
+            .stop_times
+            .iter()
+            .min()
+            .unwrap()
+            .arrival_time;
```
TeXitoi: You can mutualize these 2 lines to avoid unwrap:
```rust
let mut arrival_time_delta = match corresponding_vj.stop_times.iter().min() {
    None => {
      warn!(
                "frequency mapped to trip {:?} with no stop_times",
                frequency.trip_id
            );
            continue;
    }
    Some(st) => st.arrival_time,
};
```

src/gtfs/read.rs
```diff
@@ -879,6 +881,155 @@ pub fn set_dataset_validity_period(
...
+                    .dates
+                    .iter()
+                    .map(|d| *d + chrono::Duration::days(1))
+                    .collect();
+                let new_service_id = format!("{}:{}", service.id, generated_trip_id);
+                let new_service = objects::Calendar {
+                    id: new_service_id.clone(),
+                    dates: new_dates,
+                };
+                collections.calendars.push(new_service)?;
```
TeXitoi: It will generate a lot of useless services: one for each instance another day that the original day.

src/gtfs/read.rs
```diff
@@ -879,6 +881,155 @@ pub fn set_dataset_validity_period(
...
+                    boarding_duration: stop_time.boarding_duration,
+                    alighting_duration: stop_time.alighting_duration,
+                    pickup_type: stop_time.pickup_type,
+                    drop_off_type: stop_time.drop_off_type,
+                    datetime_estimated,
+                    local_zone_id: stop_time.local_zone_id,
+                })
+            }
+            start_time = start_time + Time::new(0, 0, frequency.headway_secs);
+            if !stop_times.is_empty() {
```
TeXitoi: I think you've already done this check before.

src/gtfs/read.rs
nberard: @prhod i talked about it with @papailio what do you think?

src/gtfs/read.rs
```diff
@@ -879,6 +881,155 @@ pub fn set_dataset_validity_period(
...
+            continue;
+        }
+        let datetime_estimated = match frequency.exact_times {
+            FrequencyExactTime::Exact => false,
+            FrequencyExactTime::Inexact => true,
+        };
+        let corresponding_vj = skip_fail!(
+            vehicle_journeys
+                .iter()
+                .find(|&vj| vj.id == frequency.trip_id)
```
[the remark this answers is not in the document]
nberard: u mean collect all vj that will be used to generate trips? (because the take is outside the loop...)

src/gtfs/read.rs
```diff
@@ -879,6 +881,155 @@ pub fn set_dataset_validity_period(
...
+                .and_modify(|counter| *counter += 1)
+                .or_insert(0);
+            let generated_trip_id = format!(
+                "{}-{}",
+                frequency.trip_id, trip_id_sequence[&frequency.trip_id]
+            );
+            // the following handles generated trip starting after midnight, we need to generate a
+            // new service in case the next day is not covered
+            let service_id = if start_time.hours() >= 24 {
+                arrival_time_delta = arrival_time_delta + Time::new(24, 0, 0);
```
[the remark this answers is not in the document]
nberard: should i create n services if start_time is "n*24::00"? ping @prhod

src/gtfs/read.rs
nberard: :+1:

src/gtfs/read.rs
nberard: updated i generate less services, and used format "service:3:+2days" are u all ok?

src/gtfs/read.rs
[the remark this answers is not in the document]
TeXitoi: I mean take after the loop, and use `CollectionWithId::get` in this loop instead of doing a linear search.

src/gtfs/read.rs
TeXitoi: I don't have opinion on the id generation, but having one service by shift resolve my comment.

src/gtfs/read.rs
```diff
@@ -1003,32 +1011,31 @@ pub fn manage_frequencies<P: AsRef<path::Path>>(
...
+                physical_mode_id: corresponding_vj.physical_mode_id.clone(),
+                dataset_id: corresponding_vj.dataset_id.clone(),
+                service_id,
+                headsign: corresponding_vj.headsign.clone(),
+                block_id: corresponding_vj.block_id.clone(),
+                company_id: corresponding_vj.company_id.clone(),
+                trip_property_id: corresponding_vj.trip_property_id.clone(),
+                geometry_id: corresponding_vj.geometry_id.clone(),
+                stop_times,
+            };
```
TeXitoi:
```rust
          let generated_vj = VehicleJourney {
                id: generated_trip_id,
                service_id,
                stop_times,
                .. corresponding_vj.clone()
            };
```

datanel: bogoss

nberard: :+1:

PR #168, opened 2018-12-07
Improve performance when handling comments on stop_times

kinnou02:
Previoulsy the copy to invert the stop_time_ids was done for every
comments, so it's a bit slow. This do it only one time to speed it.

This can be improved by doing the copy lazily to not pay the cost if
there isn't any comments on stop_times

### DIFF
--- a/src/ntfs/read.rs
+++ b/src/ntfs/read.rs
@@ -325,7 +325,7 @@ pub fn manage_comments(collections: &mut Collections, path: &path::Path) -> Resu
         let path = path.join("comment_links.txt");
         if let Ok(mut rdr) = csv::Reader::from_path(&path) {
             // invert the stop_time_ids map to search a stop_time by it's id
-            let stop_time_ids: HashMap<&String, (Idx<VehicleJourney>, u32)> = collections
+            let stop_time_ids = collections
                 .stop_time_ids
                 .iter()
                 .map(|(k, v)| (v, *k))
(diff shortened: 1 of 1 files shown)


src/ntfs/read.rs
```diff
@@ -327,6 +324,12 @@ pub fn manage_comments(collections: &mut Collections, path: &path::Path) -> Resu
 
         let path = path.join("comment_links.txt");
         if let Ok(mut rdr) = csv::Reader::from_path(&path) {
+            // invert the stop_time_ids map to search a stop_time by it's id
+            let stop_time_ids: HashMap<String, (Idx<VehicleJourney>, u32)> = collections
+                .stop_time_ids
+                .iter()
+                .map(|(k, v)| (v.clone(), *k))
+                .collect();
```
datanel: with 
```rust
let stop_time_ids: HashMap<&String, (Idx<VehicleJourney>, u32)> = collections
                .stop_time_ids
                .iter()
                .map(|(k, v)| (v, *k))
                .collect();
```
you can improve memory
on fr-idf, before 4,42GB, after: 3,74GB

TeXitoi:
```rust
HashMap<&str, (Idx<VehicleJourney>, u32)>
```

Using `BTreeMap` or a `flat_map::FlatMap` can also improve memory.

TeXitoi: `&str` add useless complexity, so my comment is fixed.

PR #171, opened 2018-12-11
Read the GTFS from a zip or an url

### LINKED ISSUE #170: Add the ability to read a zip
Hi here!

It would be quite nice to be able to directly read a `.zip` since it's usually the packaging of a GTFS.
In the same spirit, it might be helpful to read directly an url.

Would you be interested in such capabilities ?

It might need a bit of a refactor in the reader, I think we'll need to change all the utilities that takes a `AsRef<path::Path>` by  feeding them a `std::io::Read` (a bit like it's done in [gtfs-structure](https://github.com/rust-transit/gtfs-structure/blob/master/src/lib.rs#L308-L376).

As a side effect, would you be interested by a mean to have real time discussion ? It might be easier to exchange. We could set up a gitter channel or use IRC.

See you,
Antoine

antoine-de:
fix #170 

add 2 capabilities:

read a zip:
`cargo run --bin gtfs2ntfs -- --input=<my_zip> -o /tmp`

read directly a zip from an url:
`cargo run --bin gtfs2ntfs -- --url=https://static.data.gouv.fr/resources/reseau-de-transports-en-commun-de-la-communaute-dagglomeration-de-lauxerrois/20181030-150143/gtfs-20181024-04.zip -o /tmp`

:tada: 

:sparkling_heart:  A huge thanks to @TeXitoi for the help with the `FileHandler` trait :sparkling_heart: 

This could be adapted to have the same capabilities for NTFS, but that is left as an exercice to the reader :stuck_out_tongue_winking_eye: 

Note: I did not add a test on the url reading because I don't think it's a good thing to have an external dependency for the test, but if you want I can add one.

Note2: I changed the crate's version number to launch the discussion about releasing this crate on crates.io :stuck_out_tongue_winking_eye:


src/common_format.rs
```diff
@@ -117,72 +118,83 @@ impl Calendar {
...
-            let calendar: Calendar = calendar.with_context(ctx_from_path!(calendar_path))?;
-            calendars.push(objects::Calendar {
-                id: calendar.id.clone(),
-                dates: calendar.get_valid_dates(),
-            });
+    let mut calendars: Vec<objects::Calendar> = vec![];
+    {
+        let calendar_reader = file_handler.get_file(file);
+        match calendar_reader {
+            Err(_) => {
```
TeXitoi: Same here.

src/gtfs/read.rs
```diff
@@ -235,63 +236,73 @@ impl Trip {
...
     let file = "shapes.txt";
-    let path = path.as_ref().join(file);
-    if !path.exists() {
-        info!("Skipping {}", file);
-        return Ok(());
-    }
+    let reader = file_handler.get_file(file).ok();
+    match reader {
+        None => {
+            info!("Skipping {}", file);
```
TeXitoi: Same here

src/gtfs/read.rs
```diff
@@ -235,63 +236,73 @@ impl Trip {
...
 
-pub fn manage_stop_times<P: AsRef<path::Path>>(
+pub fn manage_stop_times<'a, H>(
     collections: &mut Collections,
-    path: P,
-) -> Result<()> {
+    file_handler: &'a mut H,
+) -> Result<()>
+where
+    &'a mut H: FileHandler,
```
TeXitoi: same here

src/gtfs/read.rs
```diff
@@ -235,63 +236,73 @@ impl Trip {
     }
 }
 
-pub fn manage_shapes<P: AsRef<path::Path>>(collections: &mut Collections, path: P) -> Result<()> {
+pub fn manage_shapes<'a, H>(collections: &mut Collections, file_handler: &'a mut H) -> Result<()>
+where
+    &'a mut H: FileHandler,
```
TeXitoi: same here

src/gtfs/read.rs
```diff
@@ -433,24 +443,28 @@ fn get_equipment_id_and_populate_equipments(
...
+    file_handler: &'a mut H,
     comments: &mut CollectionWithId<objects::Comment>,
     equipments: &mut EquipmentList,
 ) -> Result<(
     CollectionWithId<objects::StopArea>,
     CollectionWithId<objects::StopPoint>,
-)> {
+)>
+where
+    &'a mut H: FileHandler,
```
TeXitoi: same here

src/read_utils.rs
```diff
@@ -93,3 +97,112 @@ pub fn get_validity_period(
...
+    fn files_by_name(archive: &mut zip::ZipArchive<R>) -> BTreeMap<String, usize> {
+        let mut res = BTreeMap::new();
+        for i in 0..archive.len() {
+            let file = archive.by_index(i).unwrap();
+            // we get the name of the file, not regarding it's patch in the ZipArchive
+            let real_name = Path::new(file.name()).file_name().unwrap();
+            let real_name: String = real_name.to_str().unwrap().into();
+            res.insert(real_name, i);
+        }
+        res
```
TeXitoi: With a collect and no unwrap:

```rust
fn files_by_name(archive: &mut zip::ZipArchive<R>) -> BTreeMap<String, usize> {
    (0..archive.len())
        .filter_map(|i| {
            let file = archive.by_index(i)?;
            // we get the name of the file, not regarding it's patch in the ZipArchive
            let real_name = Path::new(file.name()).file_name()?;
            let real_name: String = real_name.to_str()?.into();
            (real_name, i)
        })
        .collect()
}

File: src/read_utils.rs [hunk 7]
```diff
@@ -93,3 +97,112 @@ pub fn get_validity_period(
...
+impl<'a, R> FileHandler for &'a mut ZipHandler<R>
+where
+    R: std::io::Seek + std::io::Read,
+{
+    type Reader = zip::read::ZipFile<'a>;
+    fn get_file(self, name: &str) -> Result<Self::Reader> {
+        // self.index_by_name
+        //     .get(name)
+        //     .map(|i| self.archive.by_index(i.clone()).unwrap())
+        //     .ok_or(format_err!("impossible to find file {}", name))
```
TeXitoi: This comment can be removed.

src/common_format.rs
```diff
@@ -117,72 +118,83 @@ impl Calendar {
...
-) -> Result<()> {
+    file_handler: &'a mut H,
+) -> Result<()>
+where
+    &'a mut H: FileHandler,
+{
     let file = "calendar_dates.txt";
-    let path = path.join(file);
-    if !path.exists() {
-        info!("Skipping {}", file);
```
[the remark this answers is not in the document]
prhod: for me, it's a technical issue, I'm not sure to be relevant here ^^

src/common_format.rs
```diff
@@ -117,72 +118,83 @@ impl Calendar {
...
+    let reader = file_handler.get_file(file);
+    match reader {
+        Err(_) => info!("Skipping {}", file),
+        Ok(reader) => {
+            info!("Reading {}", file);
+
+            let mut rdr = csv::Reader::from_reader(reader);
+            for calendar_date in rdr.deserialize() {
+                let calendar_date: CalendarDate =
+                    calendar_date.with_context(ctx_from_filename!(file))?;
```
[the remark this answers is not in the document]
prhod: indeed, it would be a better to have a full path with the name of the content file causing problem. Deal ?

src/common_format.rs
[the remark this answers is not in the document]
antoine-de: is it ok for you if I change the `get_file` to return `Result<Option<Reader>>` with Ok(None) if the file does not exists ?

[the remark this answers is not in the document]
TeXitoi: In this case you want `get_file() -> Result<File>` and `get_file_if_exists() -> Result<Option<File>>`, else managing the case of I want this file will be dirty.

src/common_format.rs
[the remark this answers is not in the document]
antoine-de: gnagnagna mutlitple mutable borrow, gnagnagna :gun:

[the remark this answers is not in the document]
antoine-de: yes that's what I did finally, because spliting the method is not possbile

I thus Returns `Result<(Option<Reader>, PathBuf)>` is that ok for you ? (did not see the comment on `get_file_exists` before :roll_eyes: , but the use is not that bad)

src/common_format.rs
```diff
@@ -174,16 +173,16 @@ where
     let file = "calendar.txt";
```
TeXitoi: Minor: The only reason of creating this variable was to log. You can directly put the string literal in the get_file call if you want to win a line.

src/read_utils.rs
```diff
@@ -100,7 +100,8 @@ pub fn get_validity_period(
 
 pub trait FileHandler {
     type Reader: std::io::Read;
-    fn get_file(self, name: &str) -> Result<Self::Reader>;
+
+    fn get_file(self, name: &str) -> Result<(Option<Self::Reader>, PathBuf)>;
```
TeXitoi: I know that's only 3 copy/paste for the moment, but it might be more later

```rust
fn get_file_if_exists(self, name: &str) -> Result<(Option<Self::Reader>, PathBuf)>;
fn get_file(self, name: &str) -> Result<(Self::Reader, PathBuf)> {
    let (reader, path) = self.get_file_if_exists(name)?;
    Ok((reader.ok_or_else(|| format_err!("file {:?} not found", path))?, path)
}
```

src/common_format.rs
antoine-de: but won't the log line be clustered a bit if I log the path?
`"skipping /srv/data/gtfs/calendar.txt` instead of just 
`"skipping calendar.txt` 

it's up to you, I don't really care

PR #177, opened 2018-12-26
[gtfs2ntfs] remove commercial mode prefix

datanel:
Problem: when we convert 2 GTFS to NTFS and merge them,  we can have 2 same commercial mode, for example "Bus" with 2 different ids.

We decided with @prhod and @papailio to normalize commercial modes and not add prefix on it

### DIFF
--- a/src/objects.rs
+++ b/src/objects.rs
@@ -311,11 +311,6 @@ impl Id<CommercialMode> for CommercialMode {
         &self.id
     }
 }
-impl AddPrefix for CommercialMode {
-    fn add_prefix(&mut self, prefix: &str) {
-        self.id = prefix.to_string() + &self.id;
-    }
-}
 
 impl_with_id!(CommercialMode);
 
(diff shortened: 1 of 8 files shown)


src/objects.rs
```diff
@@ -516,6 +516,11 @@ impl AddPrefix for Line {
     fn add_prefix(&mut self, prefix: &str) {
         self.id = prefix.to_string() + &self.id;
         self.network_id = prefix.to_string() + &self.network_id;
+    }
+}
+
+impl Line {
+    pub fn add_cm_prefix(&mut self, prefix: &str) {
         self.commercial_mode_id = prefix.to_string() + &self.commercial_mode_id;
```
kinnou02: It's a bit hacky, if we had some commercial_mode to something else than Line I'm pretty sure that we will forgot to update this `cm_prefix ` thing.
I suggest to update `add_prefix` with the `cm_prefix`.
It's probably possible to do something way better with some Trait that every object that have a commercial_mode implements.

datanel: cm prefix is now removed

Issue #176, opened 2018-12-18
Create helpers to ease unit testing


antoine-de:
This crate is well tested (good job :1st_place_medal: :sparkling_heart:  !) but mainly with integration tests, not unit tests.

For our crate https://github.com/etalab/transpo-rt we also started with integration tests, but I plan to add unit tests too.

To ease unit tests creation it would be nice to have a helper to create datasets (a bit like https://github.com/CanalTP/navitia/blob/dev/source/ed/build_helper.h#L216)

@Tristramg implemented [one in rust](https://github.com/Tristramg/csa-rust/blob/master/src/structures.rs#L52) for his CSA crate.

Would you be interested by such a feature ?

If so, since it's not a trivial problem, I think it would be nice to design at first the end use of such a builder in this issue.
Do you have an idea on how you want to write the tests ?

Note: I can't spend too much time on this, we should try to have a minimal design if you can't spend time on implementing it too.

datanel:
What we note is that with unit tests we cannot cover every cases.
And it's easier to create fixtures than create model. 
But if a piece of code need to be unit tested and has a value, we "unit test" it.

antoine-de:
Even if the subject does not seems to raise much intersset, I'd make a proposal.

What do you think ok:

```rust
 let model = ModelBuilder::default()
    .vj("toto", |vj_builder| {
        vj_builder
            .sa("A", "10:00:00".into(), "10:01:00".into())
            .make_sa("B", "11:00:00".into(), "11:01:00".into(), |mut sa| { // <-- a stop creation with total access to the object could be usefull for complex initialization
                sa.visible = false;
            })
    })
    // .vj(...)
    .build();
```

We could also go for something with less lambda (but then it cannot be chained):
```rust
let mut builder = ModelBuilder::default();

builder
    .vj("toto")
    .sa("A", "10:00:00".into(), "10:01:00".into())
    .make_sa("B", "11:00:00".into(), "11:01:00".into(), |mut sa| {
        sa.visible = false;
    });
    // builder.vj(...);
let model = builder.build();
```

what do you think ? do you see better/simplier api ?

datanel:
why you cannot chain in the 2nd proposition if you always return a builder?

antoine-de:
because `.vj(..)` will return a `VJBuilder` and not a builder (to mutate the current `vj`)

we could return a `Builder` and make `.sa(...)` work on the last `vj` but I'm not a fan of such implicit stuff

[...]

PR #179, opened 2018-12-26
Add a model builder to ease test creation

antoine-de:
linked to #176 

first working draft to a `Model` builder to ease unit test creation.

It is an early draft, the implementation is lacking in several ways for the moment, I suggest a first review on the final API before reviewing the code (there are lot of stuff to discuss on it :roll_eyes: : ).

To see the api, the unittest is a good starting point.
it looks like:

```rust
let model = ModelBuilder::default()
    .vj("toto", |vj| {
        vj.st("A", "10:00:00", "10:01:00")
          .st("B", "11:00:00", "11:01:00");
    })
    .vj("tata", |vj| {
        vj.st("A", "10:00:00", "10:01:00")
          .st("D", "11:00:00", "11:01:00");
    })
    .build();
```

I'm not a big lambda fan, but it seems the only way to "nest" a builder pattern.

Note: the builder return itself by value which can lead to  some twerk if not chaining the method, but this way the minimal example is quite simple.

Note: to control the stop_point/stop_area creation we could also add a `.sa(...)`method to the `ModelBuilder`

What do you think about this ?

late Node: I'm in holidays tonight, you have some time to think about this :stuck_out_tongue_winking_eye:


antoine-de:
pouet, any thought on this ?

I'm not convinced at all with the way to initialize the route (and even more for the line, network, ....) of the VJ, but honestly since I don't need it for my use case, I think I'll keep it simple for the moment

TeXitoi:
My point would be "go for it, we can refactor later if needed."

@datanel @nberard what do you think?

datanel:
yes, if it is just for test (currently) it is not so critical so have fun and we refactor later.

antoine-de:
ok, nice, I'll remove the ` #![allow(missing_docs)]` and some minor cleanup (I think I'll put the `WithId` behind a macro), but I think you can start reviewing the code ?

TeXitoi:
If we want to publish on crates.io, maybe better to do a dedicated sub crate to have a different semver on that.

antoine-de:
since this subcrate would depend on navitia_model how do you propose to make this link ?

antoine-de:
I made all the modification I planned to do, so I think you can review it

src/objects.rs
```diff
@@ -740,6 +784,16 @@ impl FromStr for Time {
...
 }
+
+// non failing conversion from String to Time
+// if the result has to be checked, it's better to use `parse()`
+// mainly used for the ModelBuilder
+impl From<&str> for Time {
+    fn from(s: &str) -> Time {
+        s.parse().unwrap()
+    }
+}
```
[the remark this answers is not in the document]
TeXitoi: What I don't like is that this impl is public for any duration, not only the builder need.

You can define a IntoDuration trait and implement it for &str, &Duration and Duration as a workaround, specifying that IntoDuration might fail.

src/objects.rs
```diff
@@ -122,6 +122,22 @@ macro_rules! impl_comment_links {
...
+pub trait WithId {
+    fn with_id(id: &str) -> Self;
+}
+
+macro_rules! impl_with_id {
+    ($ty:ty) => {
+        impl WithId for $ty {
+            fn with_id(id: &str) -> Self {
+                let mut r = Self::default();
+                r.id = id.to_owned();
```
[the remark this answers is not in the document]
antoine-de: hum for this, how do you think it should change the builder ?

should the builder always ask for a name before creating an object ? or take an Option ?

`builder.route("1", "route 1")`  (I'm not a fan, it makes the simple example less simple)
or `builder.route("1", Some("route 1"))` (not a fan too, too complex)

or just set the name as the id by default (but then I don't really see the point of changing this Trait.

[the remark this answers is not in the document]
TeXitoi: The point was to do the macro like that:
```rust
        impl WithId for $ty {
            fn with_id(id: &str) -> Self {
                let mut r = Self::default();
                r.id = id.to_owned();
                r.name = id.to_owned();
                r
            }
        }
```

like that, if you ask for
```rust
model = builder.vj("vj1", |vj| vj.route("r1")).build();
```
the name of the created route will be useful.

antoine-de:
thanks for the review (and for the help!), I think I changed everything

model-builder/src/builder.rs
```diff
@@ -115,13 +108,36 @@ impl<'a> ModelBuilder {
...
             model: &mut self,
             vj_idx,
-            route_modifier: None,
         };
 
         vj_initer(vj_builder);
         self
     }
 
+    /// Add a new VehicleJourney to the model
```
TeXitoi: bad comment

model-builder/src/builder.rs
```diff
@@ -115,13 +108,36 @@ impl<'a> ModelBuilder {
...
+    /// let model = model_builder::ModelBuilder::default()
+    ///        .route("l1", |r| {
+    ///             r.name = "ligne 1".to_owned();
+    ///         })
+    ///      .vj("toto", |vj| {
+    ///          vj.route("l1")
+    ///            .st("A", "10:00:00", "10:01:00")
+    ///            .st("B", "11:00:00", "11:01:00");
+    ///      })
+    ///        .build();
```
TeXitoi: bad indentation

model-builder/src/builder.rs
```diff
@@ -169,7 +185,7 @@ impl<'a> VehicleJourneyBuilder<'a> {
                 get_or_create_with(
                     &mut self.model.collections.stop_areas,
                     &sa_id,
-                    Some(|mut sa: &mut crate::objects::StopArea| sa.name = format!("sa {}", sp)),
+                    |mut sa: &mut crate::objects::StopArea| sa.name = format!("sa {}", sp),
```
TeXitoi: That's now useless, no?

model-builder/src/builder.rs
```diff
@@ -226,23 +242,24 @@ impl<'a> VehicleJourneyBuilder<'a> {
...
-            id: name.to_owned(),
-            modifier: Box::new(f),
-        });
+    pub fn route(self, id: &str) -> Self {
+        {
+            let vj = &mut self
+                .model
+                .collections
+                .vehicle_journeys
+                .index_mut(self.vj_idx);
```
TeXitoi: you can use `.vehicle_journeys[self.vj_idx];`

model-builder/src/builder.rs
```diff
@@ -252,21 +269,14 @@ impl<'a> Drop for VehicleJourneyBuilder<'a> {
     fn drop(&mut self) {
         let collections = &mut self.model.collections;
         // add the missing objects to the model (routes, lines, ...)
-        let mut new_vj = collections.vehicle_journeys.index_mut(self.vj_idx);
+        let new_vj = collections.vehicle_journeys.index_mut(self.vj_idx);
```
TeXitoi: you can use .vehicle_journeys[self.vj_idx];

model-builder/src/builder.rs
antoine-de: I don't think so, I need a mutable vj here.

TeXitoi: Yeah, my bad, `index_mut` on `CollectionWithId` is custom.

PR #188, opened 2019-02-04
Nd 71 restrict validity period

nberard:


src/collection.rs
```diff
@@ -488,7 +488,7 @@ impl<T: Id<T>> CollectionWithId<T> {
         }
     }
 
-    /// Returns a CollectionWithId with elements having ids included in the `ids_to_keep` parameter
+    /// Returns a CollectionWithId with elements matching predicate parameter
```
TeXitoi: It doesn't return, it modify inplace.

PR #189, opened 2019-02-04
#ND-82 - Add failure on gtfs/ntfs reading if no calendar.txt and calendar_dates.txt

ArnaudOggy:


src/common_format.rs
```diff
@@ -124,14 +124,20 @@ impl Calendar {
...
 ) -> Result<()>
 where
     for<'a> &'a mut H: FileHandler,
 {
     let file = "calendar_dates.txt";
     let (reader, path) = file_handler.get_file_if_exists(file)?;
     match reader {
-        None => info!("Skipping {}", file),
+        None => {
+            info!("Skipping {}", file);
```
datanel: I think this should be after the condition `if !calendar_exists {`
if none of the calendars exist we don't want to see that message

ArnaudOggy: Yep. Well seen.

PR #190, opened 2019-02-06
add a way to specify the calendar of the trip

antoine-de:
add a way to add some calendars to the builder.

btw there is a confusion between `calendar` and `service`, I know this confusion comes from the GTFS, but shouldn't we always use the same term ?

### DIFF
--- a/model-builder/src/builder.rs
+++ b/model-builder/src/builder.rs
@@ -34,7 +34,7 @@
 
 use crate::collection::{CollectionWithId, Id, Idx, RefMut};
 use crate::model::{Collections, Model};
-use crate::objects::{Route, StopPoint, StopTime, Time, VehicleJourney, WithId};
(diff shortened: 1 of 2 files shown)


antoine-de:
ping

model-builder/src/builder.rs
```diff
@@ -138,6 +138,30 @@ impl<'a> ModelBuilder {
...
 
+    /// Add a new Calendar to the model
+    ///
+    /// ```
+    /// # fn main() {
+    /// let model = model_builder::ModelBuilder::default()
+    ///      .calendar("c1", |c| {
+    ///             c.dates.insert(Date::from_ymd(2019, 2, 6));
+    ///         })
+    ///      .vj("toto", |vj| {
```
datanel: where is Bob?

PR #191, opened 2019-02-06
Manage property rules for routes

datanel:
## Need
Ad the possibility to add or modify properties for routes.

| Column | Description |
| --- | --- |
| object | route |
| object_id | The id of the object in the NTFS |
| property_name | route_name, direction_type, route_geometry, destination_id |
| property_old_value | Value before modification of the property. If the field contains *, the rule is systematically applied. Otherwise, the change is only made if the old value matches. |
| property_value | The value to be applied by the process (if the verification is ok) |

see the fixtures if you want to know this should do
### inputs
- fixtures/apply_rules/input/routes.txt
- fixtures/apply_rules/input/geometries.txt
- fixtures/apply_rules/property_rules.txt
### ouputs
- fixtures/apply_rules/output/routes.txt
- fixtures/apply_rules/output/geometries.txt


src/apply_rules.rs
```diff
@@ -82,6 +96,91 @@ fn read_complementary_code_rules_files<P: AsRef<Path>>(
...
+        }
+    }
+
+    let properties = properties
+        .iter()
+        .filter(|((object_type, object_id, property_name), property)| {
+            match object_type {
+                ObjectType::Route
+                    if ![
+                        "route_name",
```
antoine-de: not really sure, (I don't know exactly the features you want), but shouldn't stuff like this be handled at serialization (maybe en enum with different kind of rules ?) 

if not handled by the type system, it might be worth it to put this inside a `is_valid` method

@TeXitoi what do you think ?

src/apply_rules.rs
```diff
@@ -110,18 +209,137 @@ fn insert_code<T>(
...
+        );
+    }
+}
+
+fn wkt_to_geo(wkt: &str, report: &mut Report, p: &PropertyRule) -> Option<GeoGeometry<f64>> {
+    if let Ok(wkt) = wkt::Wkt::from_str(wkt) {
+        if let Ok(geo) = try_into_geometry(&wkt.items[0]) {
+            Some(geo)
+        } else {
+            warn!("impossible to convert empty point");
```
antoine-de: why would this be a point ?

src/apply_rules.rs
```diff
@@ -110,18 +209,137 @@ fn insert_code<T>(
...
+    report: &mut Report,
+) -> Option<String> {
+    if let Some(geo) = wkt_to_geo(wkt, report, p) {
+        let id = p.object_type.as_str().to_owned() + ":" + &p.object_id;
+        collection
+            .get_mut(&id)
+            .map(|mut g| {
+                g.geometry = geo.clone();
+            })
+            .unwrap_or_else(|| {
```
antoine-de: wouldn't it be better to add something like https://doc.rust-lang.org/std/collections/hash_map/enum.Entry.html#method.or_insert_with for `CollectionWithId` ?
(but I don't know the best api to do this. @TeXitoi any idea?)
or maybe reuse https://github.com/CanalTP/navitia_model/blob/master/model-builder/src/builder.rs#L61 ?

src/apply_rules.rs
datanel: the only time the conversion to wkt -> geo fails is when wkt = "EMPTY POINT" but the library doesn't know how to handle this case because it uses rust-geo which doesn't know how to handle empty points either

src/apply_rules.rs
```diff
@@ -82,6 +96,91 @@ fn read_complementary_code_rules_files<P: AsRef<Path>>(
...
+                        ),
+                        ReportType::MultipleValue,
+                    );
+                }
+                return false;
+            }
+            true
+        })
+        .flat_map(|(_, p)| p)
+        .cloned()
```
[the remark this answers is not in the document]
datanel: it's needed because
```
a collection of type `std::vec::Vec<apply_rules::PropertyRule>` cannot be built from `std::iter::Iterator<Item=&apply_rules::PropertyRule>
```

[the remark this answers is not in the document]
antoine-de: I think this can be resolved by using `into_iter()` instead of `iter()`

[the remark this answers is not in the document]
datanel: :+1:

src/apply_rules.rs
```diff
@@ -82,6 +96,90 @@ fn read_complementary_code_rules_files<P: AsRef<Path>>(
...
+            .trim(csv::Trim::All)
+            .from_path(&path)
+            .with_context(ctx_from_path!(path))?;
+        for p in rdr.deserialize() {
+            let p: PropertyRule = match p {
+                Ok(val) => val,
+                Err(e) => {
+                    report.add_warning(
+                        format!("Error reading {:?}: {}", path.file_name().unwrap(), e),
+                        ReportType::PropertyRulesRead,
```
nberard: Invalid file more appropriate no ?

src/apply_rules.rs
TeXitoi: Using an enum would be better.

TeXitoi: Mmm, enum is complicated as it depends on the object_type. But better to abstract that somewhere.

src/apply_rules.rs
TeXitoi: yes, get_or_create_with can be imported in CollectionWithId and use here and in the builder. An entry API is quite a lot of work.

src/collection.rs
```diff
@@ -538,6 +538,78 @@ impl<T: Id<T>> CollectionWithId<T> {
...
+    /// # Ok(())
+    /// # }
+    /// # fn main() { run().unwrap() }
+    /// ```
+    pub fn get_or_create<'a>(&'a mut self, id: &str) -> RefMut<'a, T> {
+        let elt = self
+            .get_idx(id)
+            .unwrap_or_else(|| self.push(T::with_id(id)).unwrap());
+        self.index_mut(elt)
+    }
```
TeXitoi: Can't you impl this with `get_or_create_with`?

src/objects.rs
```diff
@@ -1291,12 +1291,31 @@ pub struct Geometry {
...
+impl Default for Geometry {
+    fn default() -> Geometry {
+        use geo_types::Point;
+        let point: Point<f64> = (0.0, 0.0).into();
+        Geometry {
+            id: "default_geometry".to_string(),
+            geometry: point.into(),
+        }
+    }
+}
```
TeXitoi: I'm not found of that, is it really needed?

antoine-de: fond :stuck_out_tongue_winking_eye:

datanel: not fan too
it is used by WithId trait for Geometry

```rust
impl WithId for Geometry {
    fn with_id(id: &str) -> Self {
        let mut r = Self::default();
        r.id = id.to_owned();
        r
    }
}
```

PR #207, opened 2019-03-20
interpolated undefined gtfs's stop_times

antoine-de:
related to https://github.com/CanalTP/navitia_model/issues/204

implement the stoptime's interpolation defined in the [documentation](https://github.com/CanalTP/navitia_model/blob/master/src/documentation/gtfs_read.md#reading-stop_timestxt) (a simple ventilation of the unknown stops).

Note: I'm not proud of the `Time::undefined`, but I don't see an efficient way to wait for the whole file to be read. (Since the order of the stop_times is not defined, we need to read the whole file, and for
the moment use thus create an `object::StopTime`, and modify it afterward)


src/gtfs/read.rs
```diff
@@ -341,11 +345,84 @@ where
...
+}
+
+impl Undefined for objects::Time {
+    fn undefined() -> Self {
+        Time::new(0, 0, u32::max_value())
+    }
+    fn is_undefined(&self) -> bool {
+        self.total_seconds() == u32::max_value()
+    }
+}
```
TeXitoi: No way, it allows unchecked reads.

antoine-de:
I rebased this PR and made the requested changes (remove the use of `undefined` for the stoptimes), I think it's ready for review

src/gtfs/read.rs
```diff
@@ -320,35 +311,133 @@ where
...
+    }
+    collections.stop_time_headsigns = headsigns;
+
+    for (vj_idx, mut stop_times) in tmp_vjs {
+        stop_times.sort_unstable_by_key(|st| st.stop_sequence);
+
+        let mut vj = collections.vehicle_journeys.index_mut(vj_idx);
+        let st_values = interpolate_undefined_stop_times(&vj.id, &stop_times)?;
+
+        for (stop_time, st_values) in stop_times.iter().zip(st_values.iter()) {
```
datanel: you can remove `iter()` from `st_values.iter()`

src/gtfs/read.rs
```diff
@@ -320,35 +311,133 @@ where
...
     Ok(())
 }
 
+fn ventilate_stop_times(
+    undefined_stop_times: &[&StopTime],
+    before: &StopTimesValues,
+    after: &StopTimesValues,
+) -> Vec<StopTimesValues> {
+    let duration = after.arrival_time - before.departure_time;
+    let distance = duration / (undefined_stop_times.len() + 1) as u32;
```
datanel: I would choose another name than `distance` as it is a time

PR #211, opened 2019-03-29
Changed name of repository and publish to crate.io

datanel:


Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 authors = ["Kisio Digital <guillaume.pinot@kisio.org>"]
```
antoine-de: I think you can add you names as the package owner :stuck_out_tongue_winking_eye:

PR #218, opened 2019-04-10
first implementation of vptranslator

patochectp:


src/vptranslator.rs
```diff
@@ -0,0 +1,192 @@
...
+    exceptions: Vec<ExceptionDate>,
+}
+
+fn get_prev_monday(date: NaiveDate) -> NaiveDate {
+    let res = date + Duration::days(-1 * date.weekday().num_days_from_monday() as i64);
+    res
+}
+
+fn weeks(dates: &BTreeSet<NaiveDate>) -> Vec<u8> {
+    let start_date: NaiveDate = get_prev_monday(*dates.iter().next().unwrap());
```
antoine-de: each `unwrap` is a kitty dying somewhere :laughing: 

maybe you can rewrite it like:

```rust
    let start_date = match dates.iter().next().map(|d|get_prev_monday(*d)) {
        Some(d) => d,
        None => return vec![]
    };
```

src/vptranslator.rs
```diff
@@ -0,0 +1,192 @@
...
+
+    let mut res = vec![0; length];
+    for date in dates {
+        let w = date.signed_duration_since(start_date).num_weeks() as usize;
+        res[w] |= 1 << (7 - date.weekday().number_from_monday());
+    }
+    res
+}
+
+fn dists(w: u8, weeks: &Vec<u8>) -> u32 {
```
antoine-de: it's a bit more rustic to it a slice:
 ` fn dists(w: u8, weeks: &[u8]) -> u32 { `

src/vptranslator.rs
```diff
@@ -0,0 +1,192 @@
...
+    let mut res: u32 = 0;
```
antoine-de: it's completly minor and the way you wrote it is perfect, but just so you know I think you could have written it like

```rust
    weeks.iter()
    .filter(|n| **n != 0u8)
    .map(|n| (*n as u8 ^ w).count_ones())
    .sum()
```

src/vptranslator.rs
```diff
@@ -0,0 +1,192 @@
...
+    for n in weeks.iter() {
+        if *n != 0 {
+            res += (*n as u8 ^ w).count_ones();
+        }
+    }
+    res
+}
+
+fn get_min_week_pattern(weeks: &Vec<u8>) -> u8 {
```
antoine-de: same here for `&[u8]`

src/vptranslator.rs
antoine-de: can you add a better name or a comment ? I'm not sure of what this does

src/vptranslator.rs
```diff
@@ -0,0 +1,192 @@
...
+        }
+    }
+    operating_days.sort_by_key(|w| w.num_days_from_monday());
+}
+
+fn clean_extra_dates(start_date: NaiveDate, end_date: NaiveDate, dates: &mut Vec<ExceptionDate>) {
+    dates.retain(|d| d.date >= start_date && d.date <= end_date);
+}
+
+pub fn translate(dates: &BTreeSet<NaiveDate>) -> BlockPattern {
```
antoine-de: travis will complain, as it is required to have some documentation on every public method/struct, could you add some on this ?

src/vptranslator.rs
```diff
@@ -0,0 +1,192 @@
...
+    Add,
+}
+
+#[derive(Debug)]
+struct ExceptionDate {
+    date: NaiveDate,
+    exception_type: ExceptionType,
+}
+
+impl PartialEq for ExceptionDate {
```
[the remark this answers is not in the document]
antoine-de: isn't this comment still relevant ?

src/vptranslator.rs
```diff
@@ -0,0 +1,253 @@
...
+    ///Indicates operating days of the service
+    pub operating_days: Vec<Weekday>,
+    ///Start and end service day for the service interval
+    pub validity_periods: Vec<ValidityPeriod>,
+    ///List of dates where service is available or not in interval.
+    pub exceptions: Vec<ExceptionDate>,
+}
+
+fn get_prev_monday(date: Date) -> Date {
+    date + Duration::days(-1 * date.weekday().num_days_from_monday() as i64)
```
antoine-de: can't you do `date - ...` ?

src/vptranslator.rs
```diff
@@ -0,0 +1,253 @@
...
+    let mut res = vec![0; length];
+    for date in dates {
+        let w = date.signed_duration_since(start_date).num_weeks() as usize;
+        res[w] |= 1 << (7 - date.weekday().number_from_monday());
+    }
+    res
+}
+
+fn dists(w: u8, weeks: &[u8]) -> u32 {
+    let res = weeks
```
antoine-de: you can directly return res, but removing the assignation

src/vptranslator.rs
```diff
@@ -0,0 +1,253 @@
...
+    }
+    operating_days.sort_by_key(|w| w.num_days_from_monday());
+}
+
+fn clean_extra_dates(start_date: Date, end_date: Date, dates: &mut Vec<ExceptionDate>) {
+    dates.retain(|d| d.date >= start_date && d.date <= end_date);
+}
+///Allows you to present a list of dates in a readable way.
+pub fn translate(dates: &BTreeSet<Date>) -> BlockPattern {
+    let mut res = BlockPattern::default();
```
antoine-de: can't you make this variable not mutable, and create it from all the fields at the end ?

src/vptranslator.rs
```diff
@@ -0,0 +1,253 @@
...
+    let end_date = match dates.iter().next_back() {
+        Some(d) => *d,
+        None => return res,
+    };
+
+    let validity_period = ValidityPeriod {
+        start_date: start_date,
+        end_date: end_date,
+    };
+    res.validity_periods.push(validity_period);
```
antoine-de: to make it more clear that the validity has for the moment only one value, you can do:
```rust
let validity_periods = vec![ValidityPeriod {
        start_date: start_date,
        end_date: end_date,
    }];
```

src/vptranslator.rs
```diff
@@ -0,0 +1,253 @@
...
+    end_date = Date::from_ymd(2012, 7, 8);
+    dates.insert(start_date);
+    dates.insert(end_date);
+    assert_eq!(
+        compute_validity_pattern(get_prev_monday(start_date), end_date, &dates).len(),
+        1
+    );
+
+    // partial + one week
+    dates.clear();
```
antoine-de: It's better to split this in several test, this way you don't have to cleanup, and you can have several fails at once (else it will stop at the first one)

src/vptranslator.rs
```diff
@@ -0,0 +1,253 @@
...
+    assert_eq!(
+        compute_validity_pattern(get_prev_monday(start_date), end_date, &dates).len(),
+        2
+    );
+
+    // one week + partial
+    dates.clear();
+    start_date = Date::from_ymd(2012, 7, 2);
+    end_date = Date::from_ymd(2012, 7, 13);
+    dates.insert(start_date);
```
antoine-de: to ease the writting of the test, you can use https://github.com/bluss/maplit
that makes it possible to write:
`let dates = btreeset!{Date::from_ymd(2012, 7, 2), Date::from_ymd(2012, 7, 13)};`

tests/vptranslator.rs
```diff
@@ -0,0 +1,474 @@
...
+            end_date: Date::from_ymd(2012, 7, 2),
+        }
+    );
+}
+
+#[test]
+fn bound_cut() {
+    let res = translate(&get_dates_from_bitset(
+        Date::from_ymd(2012, 7, 16),
+        &format!("{}{}", "0011101", "000"),
```
antoine-de: why do you need this `format!` since all the values are hardcoded ?

src/vptranslator.rs
patochectp: I had an anomaly and I thought it was coming from there and I forgot to put the code back.
I'm sorry

src/vptranslator.rs
patochectp: but start_date is re-use after

tests/vptranslator.rs
patochectp: Reading the test is more visible because you can see the weeks

tests/vptranslator.rs
```diff
@@ -0,0 +1,474 @@
...
+    assert!(res.exceptions.is_empty());
+    assert_eq!(
+        res.validity_periods.iter().next().unwrap(),
+        &ValidityPeriod {
+            start_date: Date::from_ymd(2012, 7, 18),
+            end_date: Date::from_ymd(2012, 7, 22),
+        }
+    );
+
+    let res = translate(&get_dates_from_bitset(
```
antoine-de: you can also split this one

tests/vptranslator.rs
```diff
@@ -0,0 +1,474 @@
...
+    let ss = "0000011";
+    let pattern = format!(
+        "{}{}{}{}{}{}{}{}",
+        mttf, mttf, mttf, mtwtf, mtwtf, mttf, ss, ss
+    );
+    let res = translate(&get_dates_from_bitset(Date::from_ymd(2012, 7, 2), &pattern));
+
+    assert_eq!(get_week_from_weekday(res.operating_days), 0b1101100);
+    assert_eq!(res.exceptions.len(), 14);
+    //assert!(res.exceptions.is_empty());
```
antoine-de: can you clean up this ?

src/vptranslator.rs
antoine-de: since lots of tests seems to share the same patterns, in addition to split them you can maybe add a test utility function to reduce the noise.

to add a test utility function you can do:

```rust

#[cfg(test)]
mod tests {
   fn the_utility_function() {... }

    #[test]
    fn one_function(){...}
}
```


the utility could be 
`compute_vp_between(date1, date2) -> Vec<u8>`

PR #219, opened 2019-04-17
apply-rules on stops

datanel:
need https://github.com/CanalTP/transit_model/pull/215


fixtures/apply_rules/output/report.json
```diff
@@ -64,6 +64,26 @@
     {
       "category": "ObjectNotFound",
       "message": "object_type=route, object_id=RERAF: geometry unknown not found"
+    },
+    {
+      "category": "ObjectNotFound",
+      "message": "object_type=stop_point, object_id=CDGR, property_name=stop_position, property_old_value=POINT(1 2): WKT should be POINT"
```
nberard: confusing that error is on new value but old values is in the message

fixtures/apply_rules/output/stops.txt
```diff
@@ -0,0 +1,21 @@
+stop_id,stop_name,visible,fare_zone_id,stop_lon,stop_lat,location_type,parent_station,stop_timezone,geometry_id,equipment_id
+GDLR,Another stop point name,1,,2.372987,48.844746,0,GDL,,,
```
nberard: u can filter out those not used

PR #228, opened 2019-05-09
KV1 - Read networks and companies from LINEXXXXXX.TMI file

nberard:


src/kv1/read.rs
```diff
@@ -128,16 +129,18 @@ where
     Ok(())
 }
 
+#[derive(Deserialize, Debug)]
```
woshilapin: I don't know what is the best practice here but having all the `struct` defined at the beginning of the file (instead of in-between functions definitions) wouldn't it be better?

PR #232, opened 2019-05-14
Add kv12ntfs bin in Dockerfile

ArnaudOggy:


ArnaudOggy:
> this will not work, we need to install proj.4
ok

PR #233, opened 2019-05-15
KV1: Trip properties

datanel:
the rule is:

The trip is considered accessible if the value is **ACCESSIBLE** for all stop_times of the trip. The trip is considered not accessible if the value is **NOTACCESSIBLE** for all stop_times of the trip. The information on a trip's accessibility is considered unknown if the value is **UNKNOWN** for at least one stop_time of the trip or in case the value is the same for all stop_times of the trip


woshilapin:
The what if some are `ACCESSIBLE` and some other are `NOTACCESSIBLE`? Didn't you mean "The trip is considered not accessible if the value is NOTACCESSIBLE for **at least** 1 stop_times of the trip"?

src/kv1/read.rs
```diff
@@ -308,6 +308,56 @@ where
...
+                match acc.iter().next() {
+                    Some(acc) if *acc == Accessibility::Accessible => Availability::Available,
+                    Some(acc) if *acc == Accessibility::NotAccessible => Availability::NotAvailable,
+                    _ => Availability::InformationNotAvailable,
+                }
+            } else {
+                Availability::InformationNotAvailable
+            }
+        };
+        map_acc_vjs
```
antoine-de: this map does not seems very useful (and the number of vj can be quite large).

I think you can just `get_or_create` the TripProperty on the fly (maybe by having a local map `Availability` -> `TripPropertyId`)

src/kv1/read.rs
```diff
@@ -308,6 +308,56 @@ where
...
+            .entry(avaibility)
+            .or_insert_with(Vec::new)
+            .push(vj_id.clone());
+    }
+
+    let mut id_incr: u8 = 1;
```
woshilapin: Instead of using an manually incremented integer, why don't we start using real UUID with [`uuid`](https://www.crates.io/crates/uuid) crate? This way you don't have to manually handle the ID uniqueness but rather use a `Uuid::new_v5()` or something similar. Note however that if the final loading in memory is based on a `String` then it will probably increase the needed memory...

antoine-de: don't use use id like `<type>:<id>` in general ? it does not invalidate the `uuid` stuff (I don't have a strong opinion on this) but maybe it's less useful then

src/kv1/read.rs
antoine-de: maybe my comment was not very clear, here are some speudo code (not tested at all):
```rust
let mut trip_propeties: BTreeMap<Availability, TripProperty> = BTreeMap::new();
for (vj_id, acc) in map_vj_accs {
        let avaibility = compute_availability();

        let associated_trip_property = trip_properties.entry(availability).or_insert_with(TripProperty {....});
        vj.trip_property_id = Some(associated_trip_property.id);
}
// + add all tripproperties in collections nif not done at the creation
```

PR #237, opened 2019-05-17
ND-213 - Add kv12ntfs binary to docker's image

woshilapin:
I refactor entirely the `Dockerfile` because of the new presence of Proj.4. Before, the `Dockerfile` was based on a `scratch` version which was very light with embedded binary and only the binaries. But now, we need a dynamic link with Proj.4 llibrary so this strategy is not possible anymore.


woshilapin:
@datanel I rewrite entirely the commit with your proposition. Thank you for the work you made on it.

PR #241, opened 2019-05-21
KV1 - Create routes

woshilapin:
Add routes to the model from KV1. The `line_id` is still fake but only a line away from being complete...


src/kv1/read.rs
```diff
@@ -560,19 +564,73 @@ where
...
+fn make_routes(collections: &mut Collections, jopas: &BTreeMap<String, Jopa>) -> Result<()> {
+    for (id, mut group) in &jopas.values().group_by(|j| j.get_route_id()) {
+        let first = group.next().ok_or_else(|| {
+            format_err!(
+                "The group based on the identifier {} doesn't contain any element. Weird!",
+                id
+            )
+        })?;
+        let (origin, destination) =
+            get_route_origin_destination(collections, &id).ok_or_else(|| {
```
datanel: get_route_origin_destination always returns a `Some` so it never fails

src/kv1/read.rs
```diff
@@ -560,19 +564,73 @@ where
...
+            })?;
+        let destination_id = Some(destination_stop_area.id.clone());
+        let direction_type = if first.direction == "1" || first.direction == "A" {
+            "forward"
+        } else {
+            "backward"
+        };
+        let direction_type = Some(direction_type.to_string());
+        let line_id = first.line_planning_number.clone();
+        // TODO: Remove line below once line_id have been completed (ND-215)
```
datanel: yes but lines need routes :)

src/kv1/read.rs
```diff
@@ -560,19 +564,73 @@ where
...
-    }])?;
+fn get_route_origin_destination<'a>(
+    collections: &'a Collections,
+    route_id: &str,
+) -> Option<(&'a StopPoint, &'a StopPoint)> {
+    let stop_times = &collections
+        .vehicle_journeys
+        .values()
+        .filter(|vj| vj.route_id == route_id)
+        .min_by_key(|vj| &vj.id)? // TODO: instead of picking the first trip, find the most frequence origin (or destination) from all the trips
```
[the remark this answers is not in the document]
woshilapin: If it's `None`, then the whole function returns `None`.

src/kv1/read.rs
woshilapin: Not really, see this [coment](https://github.com/CanalTP/transit_model/pull/241/files#r285974923).

src/kv1/read.rs
woshilapin: I'll let you figure this one out :stuck_out_tongue:

src/kv1/read.rs
```diff
@@ -116,6 +117,12 @@ struct Jopa {
     journey_pattern_code: String,
 }
 
+impl Jopa {
+    fn get_route_id(&self) -> String {
```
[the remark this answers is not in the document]
woshilapin: I'll change that. I didn't see it that way since `route_id` was not a field of the `Jopa` object. But I do agree...

src/kv1/read.rs
```diff
@@ -560,19 +564,73 @@ where
...
+        .min_by_key(|vj| &vj.id)? // TODO: instead of picking the first trip, find the most frequence origin (or destination) from all the trips
+        .stop_times;
+    let origin = &collections.stop_points[stop_times[0].stop_point_idx];
+    let destination = &collections.stop_points[stop_times[stop_times.len() - 1].stop_point_idx];
+    Some((origin, destination))
+}
 
+fn make_routes(collections: &mut Collections, jopas: &BTreeMap<String, Jopa>) -> Result<()> {
+    for (id, mut group) in &jopas.values().group_by(|j| j.get_route_id()) {
+        let first = group.next().ok_or_else(|| {
```
datanel: this is false. you need to create routes by direction

src/kv1/read.rs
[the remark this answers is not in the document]
datanel: I'd rather return an Err here

src/kv1/read.rs
woshilapin: `get_route_id` returns a `String` based on `LinePlanningNumber` and `Direction`. So each `group` should be unique by direction...

PR #244, opened 2019-05-23
KV1 - Add Notices/Comments files

ArnaudOggy:
Convert KV1 comments files (NOTICEXXXX.TMI/NTCASSGNMX.TMI) to NTFS (comments.txt/comment_links.txt)

### DIFF
--- a/src/kv1/read.rs
+++ b/src/kv1/read.rs
@@ -104,6 +104,18 @@ struct PujoPass {
     wheelchair_accessible: Accessibility,
 }
 
+impl PujoPass {
+    fn vehiclejourney_id(&self) -> String {
+        format!(
+            "{}:{}:{}:{}",
+            self.line_planning_number,
+            self.journey_pattern_code,
+            self.journey_number,
+            self.schedule_code
+        )
+    }
+}
+
 #[derive(Deserialize, Debug)]
 struct Jopa {
     #[serde(rename = "[LinePlanningNumber]")]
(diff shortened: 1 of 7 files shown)


src/kv1/read.rs
```diff
@@ -767,21 +791,89 @@ where
...
+    let mut rdr = csv::ReaderBuilder::new()
+        .delimiter(b'|')
+        .trim(csv::Trim::All)
+        .from_reader(reader);
+
+    for notice_assignment in rdr.deserialize() {
+        let notice_assignment: NoticeAssignment =
+            notice_assignment.with_context(ctx_from_path!(path))?;
+
+        if let Some(notice_idx) = collections.comments.get_idx(&notice_assignment.notice_code) {
```
datanel:
```rust
for notice_assignment in rdr.deserialize() {
        let notice_assignment: NoticeAssignment =
            notice_assignment.with_context(ctx_from_path!(path))?;

        if let Some(comment_idx) = collections.comments.get_idx(&notice_assignment.notice_code) {
            if let Some(pujopasses) = map_pujopass.get(&(
                notice_assignment.line_planning_number,
                notice_assignment.journey_number,
            )) {
                for pujopass in pujopasses.iter().filter(|p| p.stop_order == 1) {
                    collections
                        .vehicle_journeys
                        .get_mut(&pujopass.vehiclejourney_id())
                        .unwrap()
                        .comment_links
                        .insert(comment_idx);
                }
            }
        }
    }
```

woshilapin: You should be able to replace the `.unwrap()` with something like
```
.ok_or_else(
    format_err!(
        "Failed to find a vehicle journey with ID {}",
        pujopass.vehiclejourney_id()))?
```

ArnaudOggy: I wonder if it's really relevant (in logs) because it's not critical. We'll see that tomorrow.

datanel: I'm pretty sure that this unwrap never fails
and I think you can remove the filter comment_links is a BtreeSet (unique values)

PR #248, opened 2019-05-24
Networks consolidation

patochectp:


src/apply_rules.rs
```diff
@@ -730,10 +830,35 @@ pub fn apply_rules(
         })
         .collect();
 
+    let line_by_network: HashMap<String, IdxSet<Line>> = model
+        .networks
+        .iter()
+        .filter_map(|(idx, obj)| {
+            let line = model.get_corresponding_from_idx(idx);
```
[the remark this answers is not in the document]
patochectp: an index corresponds to a line only

src/apply_rules.rs
```diff
@@ -73,6 +75,103 @@ struct PropertyRule {
...
+    Ok(consolidation.networks_consolidation)
+}
+
+fn check_networks_consolidation(
+    report: &mut Report,
+    networks: &CollectionWithId<Network>,
+    networks_consolidation: Vec<NetworkConsolidation>,
+) -> Result<Vec<NetworkConsolidation>> {
+    #[derive(Debug, Deserialize)]
+    struct Consolidation {
```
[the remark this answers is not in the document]
patochectp: Above all, I can delete it because it is useless

src/apply_rules.rs
```diff
@@ -73,6 +76,100 @@ struct PropertyRule {
...
+) -> Result<Vec<NetworkConsolidation>> {
+    info!("Checking networks consolidation.");
+    let mut res: Vec<NetworkConsolidation> = vec![];
+
+    for ntw in networks_consolidation.into_iter() {
+        let mut network_consolidation = false;
+        if networks.get(&ntw.network.id).is_none() {
+            if ntw.grouped_from.is_empty() {
+                report.add_error(
+                    "The grouped network is empty".to_string(),
```
datanel: which group network is empty?

src/apply_rules.rs
```diff
@@ -73,6 +76,100 @@ struct PropertyRule {
...
+                if networks.get(&ntw_grouped).is_none() {
+                    report.add_error(
+                        format!("The grouped network {} don't exists", ntw_grouped),
+                        ReportType::ObjectNotFound,
+                    );
+                } else {
+                    network_consolidation = true;
+                }
+            }
+        } else {
```
datanel: minor: I would return error earlier to have less nested `if`

```rust
if networks.get(&ntw.network.id).is_some() {
    bail!(format!("The network {} already exists", ntw.network.id));
}
if ntw.grouped_from.is_empty() {
    report.add_error(
        "The grouped network is empty".to_string(),
        ReportType::ObjectNotFound,
    );
    continue;
}
```

src/apply_rules.rs
```diff
@@ -73,6 +76,100 @@ struct PropertyRule {
...
+            bail!(format!("The network {} already exists", ntw.network.id));
+        };
+        if network_consolidation {
+            res.push(ntw);
+        } else {
+            report.add_error(
+                format!("No grouped network exists {}", ntw.network.id),
```
datanel: `No grouped networks exist` ? not sure

src/apply_rules.rs
```diff
@@ -73,6 +76,100 @@ struct PropertyRule {
...
+                report.add_error(
+                    "The grouped network is empty".to_string(),
+                    ReportType::ObjectNotFound,
+                );
+                continue;
+            }
+            for ntw_grouped in &ntw.grouped_from {
+                if networks.get(&ntw_grouped).is_none() {
+                    report.add_error(
+                        format!("The grouped network {} don't exists", ntw_grouped),
```
datanel: if it's `grouped network`, it should be ` does not exist
otherwise
`The grouped networks {} don't exist`

I'm not sure if it's singular or plural

src/apply_rules.rs
```diff
@@ -73,6 +75,102 @@ struct PropertyRule {
...
+    info!("Reading networks consolidation rules.");
+
+    #[derive(Debug, Deserialize)]
+    struct Consolidation {
+        #[serde(rename = "networks")]
+        networks_consolidation: Vec<NetworkConsolidation>,
+    }
+
+    let file = File::open(networks_consolidation_file)?;
+    let reader = BufReader::new(file);
```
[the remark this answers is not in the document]
woshilapin: Won't we affect the performance by doing this? I can see something in the documentation about using `BufReader` might help for short reads from a file
https://docs.serde.rs/serde_json/fn.from_reader.html

> When reading from a source against which short reads are not efficient, such as a `File`, you will want to apply your own buffering because `serde_json` will not buffer the input. See `std::io::BufReader`.

PR #250, opened 2019-05-27
Doc: kv1 read - adding timezone

prhod:


src/documentation/kv1_read.md
```diff
@@ -12,6 +12,7 @@ NTFS field | KV1 file | KV1 field | Mapping rule/Comment
 --- | --- | --- | ---
 network_id | LINEXXXXXX.TMI | *DataOwnerCode* | This field is prefixed.
 network_name | LINEXXXXXX.TMI | *DataOwnerCode* |
+network_timezone | | | Fixed value `Europe/Amsterdam`
```
woshilapin: So we assume a network is present in only one timezone? We don't provide an optional timezone per stop for example?

woshilapin: After discussion with @prhod, KV1 doesn't have the need for a timezone (it's going to be only one timezone). So my comment doesn't really apply here.

PR #254, opened 2019-05-29
Deprecate and rename old Fare system to prepare for implementation of V2

woshilapin:


woshilapin:
I'll provide another PR with some other solutions to avoid breaking backward compatibility. I'm closing this one for now.

PR #255, opened 2019-05-29
ND-255 - Rename old Fare model and introduce new version of the Fare model

woshilapin:
The first commit is only a renaming of `struct` so to simplify the reviewing process, you might want to look at each commit individually.


src/objects.rs
```diff
@@ -1277,9 +1282,81 @@ pub struct ODRule {
...
+    fn get_object_type() -> ObjectType {
+        ObjectType::Ticket
+    }
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone)]
+#[cfg(feature = "fare-v2")]
+pub struct TicketPrice {
+    pub ticket_id: String,
+    pub price: f64, // TODO: how to ensure this value is always positive (u64 on cents? Setter?)
```
woshilapin: Any idea how to deal with this in the model? One solution would be to have an Integer, so we would give the total number of cents. I'm not a fan of this solution.  The other solution would be to have either a setter on this field or a constructor on this `struct`, but that doesn't prevent us from doing
```
TicketPrice {
	ticket_id: "some_id",
	price: -42,
	...
}
```

src/objects.rs
```diff
@@ -1277,9 +1282,81 @@ pub struct ODRule {
...
+    pub currency: String, //TODO: evaluate crates 'iso4217' and 'steel-cent' for encoding currency
```
woshilapin: Are we interested by exploring some crate to handle the ISO-4217 for us. I've found these 2 [`iso4217`](https://crates.io/crates/iso4217) and [`steel-cent`](https://crates.io/crates/steel-cent) but they don't seem to have a lot of traction or support.

src/objects.rs
datanel: you could do a custom deserialization and check in it if the value is negative

src/objects.rs
```diff
@@ -1277,9 +1288,81 @@ pub struct ODRule {
...
+    pub id: String,
+    pub name: String,
+    pub comment: String,
+}
+#[cfg(feature = "fare-v2")]
 impl_id!(Ticket);
-impl_id!(ODRule);
+
+#[cfg(feature = "fare-v2")]
+impl GetObjectType for Ticket {
```
datanel: why do you need that?

src/objects.rs
datanel: it could be good for validation

src/objects.rs
```diff
@@ -1277,9 +1288,81 @@ pub struct ODRule {
...
+    }
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone)]
+#[cfg(feature = "fare-v2")]
+pub struct TicketPrice {
+    pub ticket_id: String,
+    pub price: f64, // TODO: how to ensure this value is always positive (u64 on cents? Setter?)
+    pub currency: String, //TODO: evaluate crates 'iso4217' and 'steel-cent' for encoding currency
+    pub validity: ValidityPeriod,
```
datanel: if you do this we need a Deserializer that transforms a  to `ticket_validity_start` and `ticket_validity_end` fields  to `ValidityPeriod` and vice versa

I suggest to have 2 fields for simplicity

src/objects.rs
```diff
@@ -1277,9 +1288,81 @@ pub struct ODRule {
...
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone)]
+#[cfg(feature = "fare-v2")]
+pub struct TicketUse {
+    pub id: String,
+    pub ticket_id: String,
+    pub max_transfers: Option<u64>,
```
datanel: are you sure the 3 fiels below are optional?

datanel:
tests fail because of transit_model=0.5.0 which is not released yet

src/objects.rs
woshilapin: In the reference, it is mentioned the [following sentence](https://github.com/prhod/navitia/blob/doc_new_fares/documentation/ntfs/ntfs_fare_extension.md#other-modifications)

> The `object_properties.txt` and `object_codes.txt` files are extended to enable the reference to a ticket_id.

The serialization of these two files are using the trait `GetObjectType` if I understand correctly so I thought I would do that right now. But I'm not 100% sure I understand fully what I'm doing :rofl:, can you confirm it does what I think it does or confirm it's useless?

src/objects.rs
```diff
@@ -1277,9 +1282,81 @@ pub struct ODRule {
     pub network_id: Option<String>,
     pub physical_mode_id: Option<String>,
 }
+impl_id!(ODRuleV1);
 
+#[derive(Serialize, Deserialize, Debug, Clone)]
```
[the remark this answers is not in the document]
woshilapin: I mostly did that to be consistent with most of the other structure `Line`, `Network`, `StopTime`, `StopPoint`, etc. I can remove it if you think it harms in any way.

src/objects.rs
woshilapin: So, in the [specification](https://github.com/prhod/navitia/blob/doc_new_fares/documentation/ntfs/ntfs_fare_extension.md#ticket_usestxt-required), they are all marked as **Required**. However, the Description says it can be empty. After some discussion, the specification says **Required** because when it is empty, it still has a meaning (on the contrary of an **Optional** field): for example, `max_transfers` empty means `you can do as many transfers as you like`. I thought `Option<>` was still the good way to model that but we could also do `String` everywhere which can be empty too. What do you think?

src/objects.rs
```diff
@@ -1277,9 +1282,81 @@ pub struct ODRule {
...
+#[cfg(feature = "fare-v2")]
+pub struct Ticket {
+    pub id: String,
+    pub name: String,
+    pub comment: String,
```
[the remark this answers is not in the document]
woshilapin: Fixed in 71eea1ef8fac5ca2b1b92913f992e72568d30a32.

src/objects.rs
woshilapin: I didn't think about that. However, SerDe propose an functionality for that kind of problem [`#[serde(flatten)]`](https://serde.rs/attr-flatten.html). I'm going to try that instead (keep reusing our already defined `struct ValidityPeriod`). Done in 532b5cc84ca4cbadfeb85bdff5a0b68bc7907a65.

src/objects.rs
woshilapin: Thanks for the nice suggestion... and then I did in 9892efcff7f7e2cf9cd80f7660a4357e65b19ff8.

src/objects.rs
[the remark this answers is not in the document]
datanel: I prefer to add it when we use it.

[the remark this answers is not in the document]
woshilapin: @datanel and me discussed it, I'm going to remove them, we'll add them later only as we need them.

src/objects.rs
woshilapin: Done in ae2ac8f803cf9ae94a798d6de59cd0716ea4e19b.

woshilapin: I'll try to add some unit test for the custom serializer and deserializer I added.

woshilapin:
I also added the feature `fare-v2` in the build (change of `.travis.yml`).

src/objects.rs
woshilapin: So after, discussion, `#[serde(flatten)]` would help for flattening but would complexify the renaming. So I removed it in 1b2c1fa462d56b340ae84aa21aeeaa3f373a3f9c and replaced with 2 fields as proposed.

woshilapin:
I removed the `serde` feature from the `chrono` crate in 6d9bc74b536ea00717ef80cabf62978d7dc1be38.

woshilapin:
I change the `currency_code` type from `CurrencyCode` to a `String` to avoid having our public API depending on a crate we're not sure is gonna stay. However, I still use this crate `iso4217` to do the validation during serialization and deserialization (see 341a1fe29705a68fcd5c59dd3c2a9b99459314aa).

PR #256, opened 2019-06-03
Add unit test for currency serialization/deserialization

woshilapin:


src/utils.rs
```diff
@@ -389,3 +389,48 @@ impl Report {
...
+            let wrapper: CurrencyCodeWrapper = serde_json::from_str(&json).unwrap();
+
+            assert_eq!(wrapper.currency_code, "EUR");
+        }
+
+        #[test]
+        fn test_de_invalid_currency_code() {
+            let result: Result<CurrencyCodeWrapper, _> =
+                serde_json::from_str("{\"currency_code\":\"XXX\"}");
+            assert!(result.is_err())
```
datanel: you don't have better the `is_err`? Can you use `[should_panic]` here?

woshilapin: I improved the test to check the error we got is the one we expect (see b13b11d982391e6d78f0d759f52a0e0e9e97f627).

woshilapin: I improved the test to check the error we got is the one we expect (see b13b11d982391e6d78f0d759f52a0e0e9e97f627).

src/utils.rs
```diff
@@ -389,3 +389,55 @@ impl Report {
...
+            let wrapper: CurrencyCodeWrapper = serde_json::from_str(&json).unwrap();
+
+            assert_eq!(wrapper.currency_code, "EUR");
+        }
+
+        #[test]
+        fn test_de_invalid_currency_code() {
+            let result: Result<CurrencyCodeWrapper, _> =
+                serde_json::from_str("{\"currency_code\":\"XXX\"}");
+            let err_msg = result.err().unwrap().to_string();
```
datanel: `let err_msg = result.unwrap_err();`

woshilapin: Indeed, I missed that one and it makes sense. Funny fact though, when I did change it, it complains because `CurrencyWrapper` does not implements `std::fmt::Debug`. The reason is because `.unwrap_err()` will panic with the value if the result is of type `Ok()`. I did the change in 62a3702f3096e9df713dc795cf81c0ddd51e2495.

src/utils.rs
```diff
@@ -389,3 +389,55 @@ impl Report {
         });
     }
 }
+
+#[cfg(test)]
+mod tests {
+
+    #[cfg(feature = "fare-v2")]
```
ArnaudOggy: You can remove this line now ;-)

woshilapin: Done in 487f3d0d57d2b212220f353963cd9795dfff7003.

PR #259, opened 2019-06-04
Add resiliency gtfs read

patochectp:


woshilapin:
Do you think we need tests for this? Note that I'm saying that but I'm not even sure how to test for logging... :wink:

src/gtfs/read.rs
```diff
@@ -607,7 +610,14 @@ where
             let mut rdr = csv::Reader::from_reader(reader);
             let mut transfers = vec![];
             for transfer in rdr.deserialize() {
-                let transfer: Transfer = transfer.with_context(ctx_from_path!(path))?;
+                let transfer: Transfer = match transfer {
```
datanel: you can use skip_fail!
```rust
let transfer: Transfer = skip_fail!(transfer.with_context(ctx_from_path!(path)));
```

patochectp: skip_fail! handle "with_context" error and not deserialize error

datanel: skip_fail does exactly what you do with match + warn!

PR #260, opened 2019-06-04
Ntfs - Read and write new fares files (v2)

ArnaudOggy:


src/objects.rs
```diff
@@ -1322,47 +1322,42 @@ pub struct TicketPrice {
...
 }
 
-#[derive(Serialize, Deserialize, Debug)]
-#[cfg(feature = "fare-v2")]
+#[derive(Serialize, Deserialize, Debug, PartialEq)]
 pub struct TicketUse {
+    #[serde(rename = "ticket_use_id")]
     pub id: String,
     pub ticket_id: String,
     pub max_transfers: Option<u64>,
```
datanel: can you just change all the u64 below bu u32. I don't think u64 is necessary

ArnaudOggy: ok done

PR #261, opened 2019-06-06
Doc: Conversion of V2 fares into V1 model

prhod:
Only the two use cases needed for the moment.


src/documentation/ntfs_fare_conversion_v2_to_V1.md
```diff
@@ -0,0 +1,103 @@
...
+"devise" | ticket_prices.txt | ticket_currency | fixed value `centime`, and only for a `EUR` value of `ticket_currency`
+
+#### fares.csv
+
+First, a transition from anything to board in the network is defined:
+
+prices.csv field | Source file | Source field | Notes/Mapping rule
+--- | --- | --- | ---
+"avant changement" |  |  | fixed value `*`
+"après changement" | ticket_use_perimeter.txt | object_id | the content of the field must follow the pattern : "network=network:[object_id]"
```
woshilapin: I'm not I understand completely what would happen if one `ticket` was tied to multiple `ticket_use_perimeter`: which `object_id` should we use in this case? Is it handled by the section below?

prhod: In this conversion, it's assumed there is only one `ticket_use_perimeter`. Do not be afraid ;)

PR #265, opened 2019-06-13
[Feature] Fares V1 - Read and write

ArnaudOggy:
#ND-260
For information, a test with fixtures (fares files in v1) already existed in
read_ntfs.rs/test_minimal_fares_stay_same
Interesting to test the presence of headers (or not) in csv files


src/ntfs/read.rs
```diff
@@ -106,82 +105,57 @@ pub fn manage_stops(collections: &mut Collections, path: &path::Path) -> Result<
...
 pub fn manage_fares_v1(collections: &mut Collections, base_path: &path::Path) -> Result<()> {
-    let file = "prices.csv";
-    if !base_path.join(file).exists() {
-        info!("Skipping fares");
+    let file_prices = "prices.csv";
+    let file_od_fares = "od_fares.csv";
+    let file_fares = "fares.csv";
+
+    if !base_path.join(file_prices).exists() || !base_path.join(file_od_fares).exists() {
+        info!("Skipping {} and {}", file_prices, file_od_fares);
```
datanel: I'm not sure this is the correct message.
you should skip all the reading of fares

ArnaudOggy: Ok I understand what you mean. All 3 files are skipped here (even if 1 is optional).

PR #267, opened 2019-06-17
ND-163 - Enrich NTFS with HelloGo fares

woshilapin:


src/hellogo_fares/read.rs
```diff
@@ -0,0 +1,988 @@
...
+            .collect::<Result<_>>()?;
+        if values.len() == 1 {
+            Ok(values[0].clone())
+        } else {
+            bail!("Failed to find the Line with identifier '{}'", line_ref)
+        }
+    }
+
+    let line_ref = get_line_ref(fare_frame)?;
+    let line_id = get_line_id_from_line_ref(service_frame, line_ref)?;
```
ArnaudOggy: As a reminder, do not forget to check if the line exists in collections.lines.

woshilapin: Thanks for the reminder. It's done in the new commit that introduce all the modification to handle prefixes (see 09154d71f7436db3c98c0ebe7fdc1c93342b55fd).

src/hellogo_fares/read.rs
```diff
@@ -0,0 +1,1120 @@
...
+                .iter()
+                .map(move |destination| (origin.clone(), destination.clone()))
+        })
+        .collect();
+    Ok(origin_destinations)
+}
+
+fn load_fare_file(collections: &mut Collections, root: &Element) -> Result<()> {
+    let prefix = get_prefix(&collections)?;
+    let prefix_column = prefix.clone() + ":";
```
ArnaudOggy: Only concatenate if a prefix exists, otherwise the result of the variable does not make sense (and objects in the collection will not be found).

woshilapin: Do you mean something like this.
```suggestion
    let prefix_column = if prefix.is_empty() {
    	"".to_string()
    } else {
    	prefix.clone() + ":"
    };
```

woshilapin: Oh, I actually made the `get_prefix()` function used just above return a `Result`, meaning, it would fail if no prefix is found. Maybe I need to convert to an `Option`. Let me try to rework all of that.

woshilapin: So I discussed the edge cases with @prhod and:
- If `contributor_id` is `NTM:` (so without a suffix), just act as normal and extract the prefix
- If `contributor_id` is something like `123` (no prefix), process the NTFS as a non-prefixed NTFS

I also realized that the name `prefix_column` didn't make much sense because... what I meant was `prefix_with_colon`. I also did the change accordingly. Look commit 599ab461779fa74b7d1dde216e1e8621d4427836 for the changes.

src/hellogo_fares/utils.rs
```diff
@@ -0,0 +1,954 @@
...
+        })
+        .map(|key_value| key_value.try_only_child("Value"))
+        .collect::<Result<Vec<_>>>()?;
+    if values.len() != 1 {
+        bail!(
+            "Failed to find a unique key '{}' in '{}'",
+            key,
+            element.name()
+        )
+    } else {
```
datanel: you can remove this line

src/hellogo_fares/utils.rs
```diff
@@ -0,0 +1,954 @@
...
+    let validity_end_date = extract_date(version, "EndDate")?;
+    Ok((validity_start_date, validity_end_date))
+}
+
+pub fn get_currency(fare_frame: &Element) -> Result<String> {
+    let currency = fare_frame
+        .try_only_child("FrameDefaults")?
+        .try_only_child("DefaultCurrency")?
+        .text();
+    if iso4217::alpha3(currency.as_str()).is_some() {
```
datanel: or early return, as you want

```rust
if iso4217::alpha3(currency.as_str()).is_none() {
    bail!("Failed to validate '{}' as a currency", currency);
}

Ok(currency)
```

src/hellogo_fares/read.rs
```diff
@@ -0,0 +1,1123 @@
...
+                        rounding_rule,
+                        rust_decimal::RoundingStrategy::RoundHalfUp,
+                    );
+                    let mut ticket_price = TicketPrice::try_from((
+                        ticket.id.clone(),
+                        price,
+                        currency.clone(),
+                        validity,
+                    ))?;
+                    let mut ticket_use = TicketUse::from(ticket.id.clone());
```
datanel: you can `impl From<&str> for TicketUse` if you don't want to clone so much

src/hellogo_fares/read.rs
```diff
@@ -0,0 +1,1123 @@
...
+            ticket_validity_start: validity.0,
+            ticket_validity_end: validity.1,
+        };
+        Ok(ticket_price)
+    }
+}
+
+impl From<String> for TicketUse {
+    fn from(ticket_id: String) -> Self {
+        let ticket_use_id = format!("TU:{}", ticket_id.clone()).to_string();
```
datanel: `let ticket_use_id = format!("TU:{}", &ticket_id);`

src/hellogo_fares/read.rs
```diff
@@ -0,0 +1,1123 @@
...
+        .flat_map(|origin| {
+            end_stop_area_ids
+                .iter()
+                .map(move |destination| (origin.clone(), destination.clone()))
+        })
+        .collect();
+    Ok(origin_destinations)
+}
+
+fn load_fare_file(collections: &mut Collections, root: &Element) -> Result<()> {
```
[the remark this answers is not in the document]
woshilapin: You are right, it's not a file. I changed it to `load_netex_fares` which might not be perfect either but maybe a little bit better (see ee4c4b80eae398a662f0037175a93a83aba3acf8).

src/hellogo_fares/read.rs
```diff
@@ -0,0 +1,1123 @@
...
+            use_origin,
+            use_destination,
+        }
+    }
+}
+
+/// For HelloGo fares connector, we need the prefix of the input NTFS.
+/// The prefix will be extracted from the 'contributor_id'
+fn get_prefix(collections: &Collections) -> Option<String> {
+    if let Some((_, contributor)) = collections.contributors.iter().next() {
```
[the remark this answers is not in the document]
woshilapin: Nice rustic improvement, done in 0c57c903253f66e108b30ac2b0a7562265971400.

src/hellogo_fares/read.rs
```diff
@@ -0,0 +1,1123 @@
...
+            end_stop_area_ids
+                .iter()
+                .map(move |destination| (origin.clone(), destination.clone()))
+        })
+        .collect();
+    Ok(origin_destinations)
+}
+
+fn load_fare_file(collections: &mut Collections, root: &Element) -> Result<()> {
+    let prefix_with_colon = if let Some(prefix) = get_prefix(&collections) {
```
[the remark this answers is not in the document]
woshilapin: Nice rustic improvement, done in 5a3d82bd345a4a4686b6eb995f30d06c3e66dd71.

src/hellogo_fares/utils.rs
woshilapin: Done in 812eec101cf927aabfedd0b13cd85d935ebe9db2.

src/hellogo_fares/utils.rs
woshilapin: Done in 3ccab5336d49fc0bfabe64813654a58ae206a90e.

src/hellogo_fares/read.rs
woshilapin: I choose to do it without the `format!()` macro. See 0c7b22f774729f47041525bdbeffc3fed02d392f.

woshilapin:
I also added some more integration tests for NTFS without prefixes. See 22d130dd94a463d673ae9bd176e82c7f1a2210bf.

PR #269, opened 2019-06-18
Improve test utils reporting for file to file comparison

woshilapin:
A proposition to improve the reporting of file-to-file comparison in the test utils. Today, even after using `pretty_assertions` for text coloration, a failing test is reported like this.

image

This PR is proposing to do an assertion line by line to show a clearer diff. Something like

image

Note: There is one drawback though and I'm open to discussion and/or solution: it's going to report only the first assertion error, not all of them.


woshilapin:
Not very useful to have only the first mismatch reported.

PR #274, opened 2019-06-26
Write fares v1 from v2

ArnaudOggy:


src/ntfs/write.rs
```diff
@@ -179,6 +175,48 @@ pub fn write_fares_v1(
...
+    ticket_uses: &CollectionWithId<TicketUse>,
+    ticket_prices: &Collection<TicketPrice>,
+    ticket_use_perimeters: &Collection<TicketUsePerimeter>,
+    ticket_use_restrictions: &Collection<TicketUseRestriction>,
+) -> Result<()> {
+    info!("xxxxxxx write_fares_v1_from_v2");
+    Ok(())
+}
+
+pub fn write_fares_v1(
```
woshilapin: I think you can change the signature to something more simple: just send the entire model.
```suggestion
pub fn write_fares_v1(
	base_path: &path::Path,
	collections: &Collections,
) -> Result<()>
```

src/ntfs/read.rs
```diff
@@ -110,7 +110,13 @@ pub fn manage_fares_v1(collections: &mut Collections, base_path: &path::Path) ->
     let file_od_fares = "od_fares.csv";
     let file_fares = "fares.csv";
 
-    if !base_path.join(file_prices).exists() || !base_path.join(file_od_fares).exists() {
+    if !base_path.join(file_prices).exists()
+        || !base_path.join(file_od_fares).exists()
+        || (!collections.tickets.is_empty()
```
woshilapin: Just to make the code more self-documenting about what you're doing, I would extract a function `has_fares_v2()`.
```rust
fn has_fares_v2(collections: &Collections) -> bool {
	!(collections.tickets.is_empty() &&
		collections.ticket_prices.is_empty() &&
		collections.ticket_uses.is_empty() &&
		collections.ticket_use_perimeters.is_empty() &&
		collections.ticket_use_restrictions.is_empty())
}
```

And you can then use it
```rust
if !base_path.join(file_prices).exists()
        || !base_path.join(file_od_fares).exists()
        || !has_fares_v2(collections)
```

src/ntfs/read.rs
```diff
@@ -110,7 +110,13 @@ pub fn manage_fares_v1(collections: &mut Collections, base_path: &path::Path) ->
...
+            && !collections.ticket_uses.is_empty()
+            && !collections.ticket_prices.is_empty()
+            && !collections.ticket_use_perimeters.is_empty())
```
woshilapin: Note that even if `ticket_use_restrictions.txt` is optional, I would still test for its presence.

Or the other option is, as long as `collections.tickets` is empty, we assume there isn't any Fares V2 in the model, no need to do an exhaustive test. This is in theory a correct solution. But I would keep the exhaustive solution like you already did in order to be more defensive in our code.

src/ntfs/write.rs
```diff
@@ -179,6 +175,48 @@ pub fn write_fares_v1(
     Ok(())
 }
 
+fn write_fares_v1_from_v2(
+    base_path: &path::Path,
+    tickets: &CollectionWithId<Ticket>,
```
woshilapin: I wonder if it would be useful to create a struct `Fares` (or `FaresV2`) that would hold the five different collections afferent to fares. Something like
```rust
struct Fares {
	tickets: &CollectionWithId<Ticket>,
   ticket_uses: &CollectionWithId<TicketUse>,
   ticket_prices: &Collection<TicketPrice>,
   ticket_use_perimeters: &Collection<TicketUsePerimeter>,
   ticket_use_restrictions: &Collection<TicketUseRestriction>,
}
```

It would simplify the signature of our functions.
```rust
fn write_fares_v1_from_v2(
    base_path: &path::Path,
    fares: &Fares,
) -> Result<()>
```

src/ntfs/write.rs
```diff
@@ -179,6 +175,48 @@ pub fn write_fares_v1(
...
+    {
+        return write_fares_v1_from_v2(
+            base_path,
+            tickets,
+            ticket_uses,
+            ticket_prices,
+            ticket_use_perimeters,
+            ticket_use_restrictions,
+        );
+    } else if !prices_v1.is_empty() && !od_fares_v1.is_empty() {
```
woshilapin: Similarly, you could write a `has_fares_v1()` function. Even if it's only for factorizing much less lines and it would be used only once, I still find it useful because only the name of the function documents what you're doing.

src/ntfs/write.rs
```diff
@@ -179,6 +175,48 @@ pub fn write_fares_v1(
...
+    prices_v1: &Collection<PriceV1>,
+    od_fares_v1: &Collection<ODFareV1>,
+    fares_v1: &Collection<FareV1>,
+    tickets: &CollectionWithId<Ticket>,
+    ticket_uses: &CollectionWithId<TicketUse>,
+    ticket_prices: &Collection<TicketPrice>,
+    ticket_use_perimeters: &Collection<TicketUsePerimeter>,
+    ticket_use_restrictions: &Collection<TicketUseRestriction>,
+) -> Result<()> {
+    if !tickets.is_empty()
```
woshilapin: You could use the `has_fares_v2()` function here (see [the above comment]()).

src/ntfs/read.rs
```diff
@@ -110,7 +111,10 @@ pub fn manage_fares_v1(collections: &mut Collections, base_path: &path::Path) ->
     let file_od_fares = "od_fares.csv";
     let file_fares = "fares.csv";
 
-    if !base_path.join(file_prices).exists() || !base_path.join(file_od_fares).exists() {
+    if !base_path.join(file_prices).exists()
```
woshilapin: I'm not sure I understand this `if` test. We do not want to write Fare V1 if:
1. We do not have Fare V1 in the input
2. We do have Fare V2 in the model

I thought every time we do have Fare V2, we want to write V1. And if we do not have Fare V2 in the model, we want to write V1 only if it's in the input NTFS... so basically the opposite of this test. Just feels like I'm missing something.

src/ntfs/write.rs
```diff
@@ -179,6 +177,229 @@ pub fn write_fares_v1(
...
+    tickets: &CollectionWithId<Ticket>,
+    ticket_uses: &CollectionWithId<TicketUse>,
+    ticket_prices: &Collection<TicketPrice>,
+    ticket_use_perimeters: &Collection<TicketUsePerimeter>,
+    ticket_use_restrictions: &Collection<TicketUseRestriction>,
+) -> Result<()> {
+    fn has_no_transfer(ticket_uses: &CollectionWithId<TicketUse>, ticket_use_id: &String) -> bool {
+        if let Some(ticket_use) = ticket_uses.get(ticket_use_id) {
+            if (ticket_use.max_transfers.is_none() || ticket_use.max_transfers == Some(0))
+                && ticket_use.boarding_time_limit.is_none()
```
woshilapin: I'm not sure `boarding_time_limit` and `alighting_time_limit` are relative to a transfer. I thought you can have a `boarding_time_limit` on the Ticket Use, even if no transfer is allowed.

src/ntfs/write.rs
```diff
@@ -179,6 +177,229 @@ pub fn write_fares_v1(
...
+        }
+        false
+    }
+
+    let mut prices_v1: Vec<PriceV1> = vec![];
+    let mut fares_v1: Vec<FareV1> = vec![];
+
+    // Conversion of OD fares on specific lines
+
+    for ticket_use_restriction in ticket_use_restrictions
```
woshilapin: I think I would extract this for-loop into a function. So you'd have a `convert_od_on_line()` function, then another function `convert_flat_fare_on_network()` and you could even put the section of the specification for each in a RustDoc comment.

src/ntfs/write.rs
```diff
@@ -179,6 +177,229 @@ pub fn write_fares_v1(
...
+            .into_iter()
+            .filter(|(_, ticket_use_perimeter)| {
+                &ticket_use_perimeter.ticket_use_id == ticket_use_id
+                    && ticket_use_perimeter.object_type == ObjectType::Line
+                    && ticket_use_perimeter.perimeter_action == PerimeterAction::Included
+            })
+            .map(|(_, p)| p)
+            .collect();
+
+        if perimeters.is_empty() {
```
woshilapin: I think we should fail if 'parameters.len() != 1` because if there is 2 perimeters, we also are not in the condition of HelloGo Fares.

src/ntfs/write.rs
```diff
@@ -179,6 +177,229 @@ pub fn write_fares_v1(
...
+
+        let prices: Vec<_> = ticket_prices
+            .into_iter()
+            .filter(|(_, ticket_price)| {
+                &ticket_price.ticket_id == &ticket.id && ticket_price.currency == "EUR".to_string()
+            })
+            .map(|(_, p)| p)
+            .collect();
+
+        if prices.is_empty() {
```
woshilapin: I think we should fail if 'prices.len() != 1` because if there is 2 prices, we also are not in the condition of HelloGo Fares.

src/ntfs/write.rs
```diff
@@ -179,6 +177,229 @@ pub fn write_fares_v1(
...
+            continue;
+        }
+
+        for price in prices {
+            let cents_price = if let Some(cents_price) =
+                (price.price * Decimal::from(100)).round_dp(0).to_u32()
+            {
+                cents_price
+            } else {
+                0
```
woshilapin: Do we really want to put a `0` in case we're not able to convert whatever comes in into a number? Let's discuss it.

src/ntfs/write.rs
ArnaudOggy: As seen together let us see further than HelloGo with n perimeters

src/ntfs/write.rs
ArnaudOggy: There can be 2 prices (or more). For Y, Y+1, ...

ArnaudOggy:
Jenkins, test this please

src/ntfs/write.rs
```diff
@@ -179,6 +212,205 @@ pub fn write_fares_v1(
...
+    for ticket_use_perimeter in ticket_use_perimeters {
+        let ticket_use_id = &ticket_use_perimeter.ticket_use_id;
+        let ticket = match fares
+            .ticket_uses
+            .get(ticket_use_id)
+            .and_then(|ticket_use| fares.tickets.get(&ticket_use.ticket_id))
+        {
+            Some(t) => t,
+            _ => {
+                info!("Failed to find Ticket for TicketUseId {:?}", ticket_use_id);
```
datanel: info or warn ?

src/ntfs/write.rs
```diff
@@ -179,6 +212,209 @@ pub fn write_fares_v1(
...
+    ticket_prices: &'a Collection<TicketPrice>,
+    ticket_uses: &'a CollectionWithId<TicketUse>,
+    ticket_use_perimeters: &'a Collection<TicketUsePerimeter>,
+    ticket_use_restrictions: &'a Collection<TicketUseRestriction>,
+}
+
+fn has_constraints(ticket_use: &TicketUse) -> bool {
+    ticket_use.max_transfers.map(|mt| mt != 0).unwrap_or(false)
+        || ticket_use.boarding_time_limit.is_some()
+        || ticket_use.alighting_time_limit.is_some()
```
datanel:
```rust
ticket_use
        .max_transfers
        .filter(|&mt| mt != 0)
        .or(ticket_use.boarding_time_limit)
        .or(ticket_use.alighting_time_limit)
        .is_some()
```

src/ntfs/write.rs
ArnaudOggy: warn

PR #280, opened 2019-07-02
Ntfs version 0.9

datanel:
and removed CHANGELOG.md


src/ntfs/mod.rs
```diff
@@ -306,7 +306,7 @@ mod tests {
                     ("feed_end_date".to_string(), "20180131".to_string()),
                     ("feed_publisher_name".to_string(), "Nicaragua".to_string()),
                     ("feed_start_date".to_string(), "20180130".to_string()),
-                    ("ntfs_version".to_string(), "0.7".to_string()),
+                    ("ntfs_version".to_string(), "0.9".to_string()),
```
woshilapin: Why not using `crate::NTFS_VERSION` here?

woshilapin: I didn't see but it's in a test so... to be discussed another day... maybe.

PR #281, opened 2019-07-02
Add utils to build and use a .deb for PROJ

woshilapin:


Dockerfile
```diff
@@ -1,30 +1,36 @@
...
-	&& make install \
-	&& cd .. && rm -rf proj-$PROJ_VERSION proj-$PROJ_VERSION.tar.gz \
-	&& apt-get purge -y wget build-essential sqlite3 libsqlite3-dev \
-	&& apt-get autoremove -y \
-	&& rm -rf /var/lib/apt/lists/* \
-	&& cp -r /build_proj/usr/include/* /usr/include \
-	&& cp -r /build_proj/usr/lib/* /usr/lib
+RUN apt update && apt install --yes apt-transport-https gnupg2
+RUN echo "deb [arch=amd64] https://kisiodigital.jfrog.io/kisiodigital/debian-local stretch main" > /etc/apt/sources.list.d/kisio-digital.list
+RUN gpg2 --receive-keys ${GPG_KEY}
```
woshilapin: It's possible to get the key in a different way, like TravisCI is doing. Something like
```
curl -sSL "https://kisiodigital.jfrog.io/kisiodigital/api/gpg/key/public" | sudo -E apt-key add -
```
We might need to use `wget` instead (because `curl` is probably not present on a stretch distribution) but you get the idea.  I'll try to work on that.

woshilapin: I just made the update so now, it's using `apt-key` to import the key directly from the Artifactory instance (and not from a GPG key server anymore).

Dockerfile
```diff
@@ -1,30 +1,35 @@
...
-	&& cd .. && rm -rf proj-$PROJ_VERSION proj-$PROJ_VERSION.tar.gz \
-	&& apt-get purge -y wget build-essential sqlite3 libsqlite3-dev \
-	&& apt-get autoremove -y \
-	&& rm -rf /var/lib/apt/lists/* \
-	&& cp -r /build_proj/usr/include/* /usr/include \
-	&& cp -r /build_proj/usr/lib/* /usr/lib
+RUN apt update && apt install --yes apt-transport-https gnupg2 wget
+RUN wget --quiet --output-document - "https://kisiodigital.jfrog.io/kisiodigital/api/gpg/key/public" | apt-key add -
+RUN echo "deb [arch=amd64] https://kisiodigital.jfrog.io/kisiodigital/debian-local stretch main" > /etc/apt/sources.list.d/kisio-digital.list
+RUN apt update && apt install --yes pkg-config libsqlite3-dev libssl-dev clang proj=${PROJ_VERSION}
```
datanel:
```suggestion
RUN apt update && apt install --yes pkg-config libssl-dev clang proj=${PROJ_VERSION}
```

Dockerfile
```diff
@@ -1,30 +1,35 @@
...
-RUN apt update && apt install -y  libssl-dev libsqlite3-dev && rm -rf /var/lib/apt/lists/*
-COPY --from=builder  /build_proj/usr/share/proj/ /usr/share/proj/
-COPY --from=builder  /build_proj/usr/include/ /usr/include/
-COPY --from=builder  /build_proj/usr/lib/ /usr/lib/
+
+RUN apt update && apt install --yes apt-transport-https gnupg2 wget \
+	&& wget --quiet --output-document - "https://kisiodigital.jfrog.io/kisiodigital/api/gpg/key/public" | apt-key add - \
+	&& echo "deb [arch=amd64] https://kisiodigital.jfrog.io/kisiodigital/debian-local stretch main" > /etc/apt/sources.list.d/kisio-digital.list \
+	&& apt update \
+	&& apt install --yes libssl-dev libsqlite3-dev proj=${PROJ_VERSION} \
```
datanel:
```suggestion
	&& apt install --yes proj=${PROJ_VERSION} \
```

.dockerignore
```diff
@@ -1,2 +1,5 @@
-./target
+examples/
+target/
```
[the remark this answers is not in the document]
woshilapin: This is not needed. `./target/` would fail but `target/` will work in any nested folder.

PR #284, opened 2019-07-08
Write Fare V1 if od_fares empty

datanel:
Write Fare V1 if od_fares is  empty and prices not empty

### DIFF
--- a/src/ntfs/mod.rs
+++ b/src/ntfs/mod.rs
@@ -117,12 +117,13 @@ fn has_fares_v2(collections: &Collections) -> bool {
         && !collections.ticket_uses.is_empty()
         && !collections.ticket_use_perimeters.is_empty()
 }
+
 /// Checks if minimum FaresV1 collections are defined and not empty (fares_v1 is optional)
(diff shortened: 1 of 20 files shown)


src/ntfs/mod.rs
```diff
@@ -117,11 +117,6 @@ fn has_fares_v2(collections: &Collections) -> bool {
         && !collections.ticket_uses.is_empty()
         && !collections.ticket_use_perimeters.is_empty()
 }
-/// Checks if minimum FaresV1 collections are defined and not empty (fares_v1 is optional)
-/// See https://github.com/CanalTP/navitia/blob/dev/documentation/ntfs/ntfs_fare_extension_fr_deprecated.md
-fn has_fares_v1(collections: &Collections) -> bool {
```
woshilapin: I think it is not useless to have this function, even for one simple line:
- You can have the above documentation to describe what makes a collections FareV1 or not
- The name of the function itself serves as a documentation (`if has_fares_v1(...)` is clear about what it does, `if prices.is_empty()` is a little less clear)

PR #287, opened 2019-07-12
Add WGS84 trait for objects that can expose WGS84 coordinates

woshilapin:
This trait will be used in the coming implementation of NaPTAN parsing too so it's not a one-shot trait.

### DIFF
--- a/src/wgs84.rs
+++ b/src/wgs84.rs
@@ -1,0 +1,24 @@
+// Copyright 2017 Kisio Digital and/or its affiliates.
+//
+// This program is free software: you can redistribute it and/or
+// modify it under the terms of the GNU General Public License as
+// published by the Free Software Foundation, either version 3 of the
+// License, or (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful, but
+// WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+// General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see
+// <http://www.gnu.org/licenses/>.
+
+use crate::{objects::Coord, Result};
+
+/// Interface dealing with spatial coordinates in EPSG:4326 format aka WGS84
+/// See https://epsg.io/4326
+pub trait WGS84Coordinates {
+    /// Returns a WGS84 spatial coordinates
+    fn coord(&self) -> Result<Coord>;
+}
--- a/src/kv1/read.rs
+++ b/src/kv1/read.rs
@@ -20,13 +20,13 @@ use crate::{
     model::Collections,
     objects::*,
     read_utils::FileHandler,
-    Result,
+    Result, WGS84Coordinates,
 };
 use chrono::NaiveDate;
 use csv;
 use failure::{bail, format_err, ResultExt};
 use geo::algorithm::centroid::Centroid;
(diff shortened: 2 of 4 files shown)


woshilapin:
This is what I've been working on on another branch. Does it answer the questions you have in mind?

For Stop Points
https://github.com/woshilapin/transit_model/blob/naptan/src/transxchange/naptan.rs#L27-L48
and for StopAreas
https://github.com/woshilapin/transit_model/blob/naptan/src/transxchange/naptan.rs#L58-L86

Usually, we do not use `Coord` to store the incoming data (we use specific `struct` corresponding to the input CSV format). Therefore, we can almost assume that when you have an instance of `Coord` in our code, it is already WGS84 format.

So 2 ways of seeing the problem, either `Coord` always have WGS84 coordinates, in this case, we could implement the WGS84 the following way (not sure we do want it...)
```
impl WGS84 for Coord {
    fn wgs84_coordinates(&self) -> Result<Coord> {
        self
    }
}
```

or `Coord` is not necessary WGS84 coordinates and in this case, I understand totally your concern. We could make the `Coord` evolve into something like
```
struct Coord {
  proj_epsg: &str, // for example, "+proj=longlat +datum=WGS84 +no_defs"
  lon: f64,
  lat: f64,
}
```
and therefore the implementation would be something like
```
const EPSG_4326: &str = "+proj=longlat +datum=WGS84 +no_defs"; // See https://epsg.io/4326
#[cfg(feature = "proj")]
impl WGS84 for Point {
    fn wgs84_coordinates(&self) -> Result<Coord> {
        let epsg_28992_point = GeoPoint::new(self.lon, self.lat);
        let converter = Proj::new_known_crs(self.proj_epsg, &EPSG_4326, None).ok_or_else(|| {
            format_err!(
                "Proj cannot build a converter from '{}' to '{}'",
                self.proj_epsg,
                EPSG_4326
            )
        })?;
        converter.convert(epsg_28992_point).map(Coord::from)
    }
}
```

Another thing I've explored was to use `geo_types` crate for the definition of the `WGS84` trait.
```
pub trait WGS84 where F: geo_types::CoordinateType {
    fn wgs84_coordinates(&self) -> Result<geo_types::Point<F>>;
}
```
but that doesn't solve your question either (as long as the `struct` used to store the coordinates doesn't contain which EPSG it's in...)

datanel:
why not simply use `TryFrom`?
```rust
impl TryFrom<KV1Stop> for Coord
impl TryFrom<TXStop> for Coord
...
```

src/wgs84.rs
```diff
@@ -0,0 +1,25 @@
...
+
+use crate::{objects::Coord, Result};
+
+/// Interface dealing with spatial coordinates in EPSG:4326 format aka WGS84
+/// See https://epsg.io/4326
+pub trait WGS84 {
+    /// Returns a WGS84 spatial coordinates in a tuple containing respectively
+    /// the longitude and the latitude
+    fn wgs84_coordinates(&self) -> Result<Coord>;
+}
```
datanel: Coord struct already has Wsg84 coordinates (not documented but we should)
I suggest
```rust
pub trait WGS84Coordinates {
    /// Returns a Coord with WGS84 spatial coordinates
    fn coord(&self) -> Result<Coord>;
}
```

@ArnaudOggy @patochectp

woshilapin: I think I'll go with that for now and I'll add some documentation on the `struct Coord` mentioning WGS84.

PR #289, opened 2019-07-12
Move TryAttribute and TryOnlyChild into a reusable module

woshilapin:


src/minidom_utils/mod.rs
```diff
@@ -0,0 +1,23 @@
...
+// along with this program.  If not, see
+// <http://www.gnu.org/licenses/>.
+
+//! Some utilities to use Minidom and returns [Result](crate::Result) when
+//! parsing instead of [Option](Option)
+
+mod try_attribute;
+pub use try_attribute::TryAttribute;
+mod try_only_child;
+pub use try_only_child::TryOnlyChild;
```
datanel: why not put everything here?

woshilapin: Small modules, easy to maintain?  I took a look on the web and here is some quotes from a [Reddit thread](https://www.reddit.com/r/rust/comments/alsph9/rusts_modules_and_project_organization_best/) that I found useful.

> For example, a suggestion I found once was: one module per trait/type definition.

> A useful not Rust-specific idea is that _users_ of a module prefer a flat namespace, while _implementers_ prefer hierarchical namespace. So nested module structure plus a flat facade with re-exports as a crate's interface is usually a good idea

PR #290, opened 2019-07-15
Implement structures and skeleton for parsing NaPTAN

woshilapin:


src/transxchange/naptan.rs
```diff
@@ -0,0 +1,264 @@
...
+    name: String,
+    #[serde(rename = "Easting")]
+    easting: f64,
+    #[serde(rename = "Northing")]
+    northing: f64,
+}
+
+const EPSG_27700: &str = "EPSG:27700";
+// FIXME: String 'EPSG:4326' is failing at runtime (string below is equivalent but works)
+const EPSG_4326: &str = "+proj=longlat +datum=WGS84 +no_defs"; // See https://epsg.io/4326
```
datanel: what's the need to put this as const?

src/transxchange/naptan.rs
```diff
@@ -0,0 +1,264 @@
...
+                FileType::StopAreas => write!(f, "StopAreas"),
+                FileType::StopsInArea => write!(f, "StopsInArea"),
+                FileType::Stops => write!(f, "Stops"),
+            }
+        }
+    }
+    let zip_file = File::open(naptan_path)?;
+    let mut zip_archive = ZipArchive::new(zip_file)?;
+    let mut files = HashMap::new();
+    for index in 0..zip_archive.len() {
```
datanel: you can't use http://mvdnes.github.io/rust-docs/zip-rs/zip/read/struct.ZipArchive.html#method.by_name ?

woshilapin: Let's consider my solution a dumb one and try yours instead :stuck_out_tongue_closed_eyes:

src/transxchange/naptan.rs
woshilapin: It won't change so I thought `const` was a good idea... Note that it will avoid do declare the variable in the function every time we call the function `coord()` (even if the compiler probably do the optimization...)

src/transxchange/naptan.rs
woshilapin: Much simpler (like by a lot). Thanks (see 7b7fc368a7bd98b432f821d641f6e1186249e4b6).

PR #295, opened 2019-07-17
Allow faresV2 without price file + fix possibility of duplicate on faresV1 conversion

ArnaudOggy:


src/ntfs/mod.rs
```diff
@@ -109,11 +109,10 @@ fn default_visible() -> bool {
     true
 }
 
-/// Checks if minimum FaresV2 collections are defined and not empty (ticket_use_restrictions is optional)
+/// Checks if minimum FaresV2 collections are defined and not empty (ticket_use_restrictions and ticket_prices are optional)
```
woshilapin: So actually, all the fare files are optional, see https://github.com/CanalTP/ntfs-specification/pull/1. I'm not really sure how to modify this function... Maybe if any of the file is present?
```rust
!collections.tickets.is_empty()
	|| !collections.ticket_prices.is_empty()
	|| !collections.ticket_uses.is_empty()
	|| !collections.ticket_use_perimeters.is_empty()
	|| !collections.ticket_use_restrictions.is_empty()
```
Or just test for the presence of `collections.tickets` because if any of the other is present (and valid), they should reference a `ticket_id` or a `ticket_use_id` which would itself reference a `ticket_id`. So any of the other collections is linked to `ticket_id`.

ArnaudOggy: All files are optionnals because ntfs-fares is an extension, but when you use this extension, 3 files are required (ticket, ticket_use and ticket_perimeter). It described in specification and seen with Pascal and Alex

woshilapin: Yep, indeed, I missed that.

woshilapin:
I just saw in the two fixtures:
- `fixtures/merge-ntfs/input_farev2_conflicts`
- `fixtures/merge-ntfs/input_faresv2_without_euro_currency`

Might be nice to be coherent in the naming `farev2` vs `faresv2`.

ArnaudOggy:
> I just saw in the two fixtures:
> 
>     * `fixtures/merge-ntfs/input_farev2_conflicts`
> 
>     * `fixtures/merge-ntfs/input_faresv2_without_euro_currency`
> 
> 
> Might be nice to be coherent in the naming `farev2` vs `faresv2`.

I saw this. So, with or without a "s" ;)

woshilapin:
I'd say with an `s` but that's really just a matter of taste 😄

PR #296, opened 2019-07-19
Add 'platform_code' to StopPoint (NTFS 0.9.1)

woshilapin:


ArnaudOggy:
> there is no need to add platform_code on every input fixtures
> 
>     * ntfs with platform Some -> ntfs with platform code Some
> 
>     * ntfs with no platform (no header)  -> ntfs platform code header but no values
> 
>     * gtfs with platform code -> ntfs with platform code header but no value
> 
>     * gtfs with no platform code -> ntfs with platform code header but no value (we can skip this test?)
> 
>     * ntfs with platform code -> gtfs with platform code header but no value
> 
>     * ntfs with no platform code -> gtfs with platform code header but no value (we can skip this test?)

If you rollback stops.txt inputs, do it too on feed_infos.txt (to leave in 0.9[.0])

ArnaudOggy:
> @ArnaudOggy About switching to `0.9`, I disagree, even if the input NTFS is minimal, not having a `platform_code` doesn't mean it's not compatible with it. The field is optional so a `stops.txt` without a `platform_code` is compatible with `0.9.1`. So I think it'd be correct to update all input to `0.9.1`.

Indeed it is  absolutely valid ! 
It's just that it will never exist in reality (version 0.9.1 in feed_infos.txt and stops.txt without platform_code). It's just an insignificant detail :wink:

ArnaudOggy:
Wait for Tartare fixtures update to merge

fixtures/gtfs2ntfs/platforms/input/stops.txt
```diff
@@ -0,0 +1,16 @@
+stop_id,stop_name,stop_lat,stop_lon,location_type,parent_station,platform_code
```
datanel: it's a big fixture just for a platform code

ArnaudOggy: @datanel  
@woshilapin told me the same thing. I take the time to reduce it?

woshilapin: I just did it in 7f6aaf790546bdf5715101239c407dff176ba397.

ArnaudOggy: Nice :+1:

PR #297, opened 2019-07-22
Add contributors, datasets and feed_infos from TransXChange

woshilapin:


woshilapin:
@ArnaudOggy I just changed the naming of structures for less misleading about what is a `StopArea` and `Stop` in e0269aefd4b72b5909e8261f79169da143408b48.

woshilapin:
I changed the rule about datasets: validity periods will be updated for all the available datasets (see d1ff87771f422235a598169e25e9f38e49552155).

src/transxchange/read.rs
```diff
@@ -59,6 +124,19 @@ fn load_company(operator: &Element) -> Result<Company> {
...
+        .get_id_to_idx()
+        .values()
+        .cloned()
+        .collect();
+    dataset_indexes.into_iter().try_for_each(|index| {
+        update_validity_period(
+            &mut collections.datasets.index_mut(index),
+            &service_validity_period,
+        )
+    })?;
```
datanel: you could do "set min and max` + get_service_validity_period + update_validity_period in a same function

woshilapin: I'm not sure we can because the "set min/max" is done only once, and the rest is done on every TransXChange file. But I can take a look about merging `get_service_validity_period` and `update_validity_period`.

src/transxchange/read.rs
```diff
@@ -92,11 +170,26 @@ where
...
 
 /// Read TransXChange format into a Navitia Transit Model
-pub fn read<P>(transxchange_path: P, naptan_path: P) -> Result<Model>
+pub fn read<P>(transxchange_path: P, naptan_path: P, config_path: Option<P>) -> Result<Model>
 where
     P: AsRef<Path>,
 {
     let mut collections = Collections::default();
+    let (contributors, mut datasets, feed_infos) = crate::read_utils::read_config(config_path)?;
+    let dataset_indexes: Vec<_> = collections
```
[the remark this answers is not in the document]
woshilapin: Good hint. I changed it in 0f171e008ec61081fea3d2bce5c5ae5327f59568.

src/transxchange/read.rs
woshilapin: I extracted a function `update_validity_period_from_transxchange` for the logic of updating the datasets. Inside, I keep calling the existing functions `get_service_validity_period` and `update_validity_period` because each is really doing something different (easier to test too if they're independent). The new function is wrapping the calls to these functions (see 6245d6fc146fd845f6f572bd97778ac2752443b6).

PR #298, opened 2019-07-23
Parse Stops.csv file in NaPTAN

woshilapin:


src/transxchange/naptan.rs
```diff
@@ -150,10 +150,26 @@ where
         .delimiter(b',')
         .trim(csv::Trim::All)
         .from_reader(reader);
+    let mut stop_points = CollectionWithId::default();
     for record in reader.deserialize() {
-        let _stop: Stop = record.with_context(|_| "Error parsing the CSV record into a Stop")?;
+        let stop: Stop = record.with_context(|_| "Error parsing the CSV record into a Stop")?;
+        let stop_area_id = stops_in_area.get(&stop.atco_code).cloned().ok_or_else(|| {
+            format_err!(
+                "Failed to find the corresponding StopArea for the StopPoint '{}'",
```
datanel: you also can check that `stop_area_id` is in collections.stop_areas
but this could be done in `read_stops_in_area` instead of `validate_stops`

src/transxchange/naptan.rs
```diff
@@ -266,10 +282,11 @@ mod tests {
 
         #[test]
         #[should_panic]
-        fn missing_coords() {
-            let csv_content = r#""StopAreaCode","Name"
-"010G0001","Bristol Bus Station"
-"010G0002","Temple Meads""#;
+        fn duplicate_id() {
```
datanel: I wonder if this kind of test is useless. It is already tested in `collectionWithId::push` or `collectionWithId::new`

src/transxchange/naptan.rs
woshilapin: I prepared the validation in a following PR (see the diff here https://github.com/woshilapin/transit_model/compare/naptan-stops...woshilapin:naptan-validate?expand=1)

src/transxchange/naptan.rs
woshilapin: So we discussed about that test with @datanel and decided to keep it. The concept of testing for duplicate in the collection and failing when there is a duplicate is a valid use case to test (but edge case, I agree).

PR #300, opened 2019-07-25
read / write grid files

patochectp:


woshilapin:
It might be nice to add them to an integration test like `ntfs2ntfs` just to see if we parse them correctly, and write them correctly.

src/objects.rs
```diff
@@ -1440,6 +1440,62 @@ impl AddPrefix for TicketUseRestriction {
...
+        deserialize_with = "de_from_date_string",
+        serialize_with = "ser_from_naive_date"
+    )]
+    pub end_date: Date,
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct GridRelCalendarLine {
+    pub calendar_id: String,
+    #[serde(skip)]
```
woshilapin: Not sure I understand why this field is skipped?

patochectp: files in production don't have this field

patochectp:
I am waiting for the validation of the specifications

woshilapin:
I updated this PR, following the specifications. More to come on the topic of `grid_*.txt` files for:
- prefixes
- merge
- sanitize

PR #301, opened 2019-07-25
Validate that StopArea IDs are valid in NaPTAN

woshilapin:


woshilapin:
I changed the code to `warn` instead of `bail` when a Stop Area ID doesn't exist for a Stop Point.

woshilapin:
I just added some filtering on `Stop` and `StopArea` that don't have the status `act` (see c627fc8d5e6586e44c8cd76ba931c0b0af2ca245).

src/transxchange/naptan.rs
```diff
@@ -107,13 +125,25 @@ where
         .map(|record: csv::Result<NaPTANStopInArea>| {
             record.with_context(|_| "Error parsing the CSV record into a StopInArea")
         })
+        .filter(|record| {
```
datanel: I'm not fan of this. It's really difficult to understand, it seems tricky.
a for loop would increase the understanding imho

@ArnaudOggy  feel free to merge

woshilapin: I extracted a function instead of transforming into a `for` loop. It's not perfect but it helps the readability.

PR #302, opened 2019-07-26
 Implement AddPrefix for Collections, Collection and CollectionWithId

woshilapin:


src/model.rs
```diff
@@ -463,6 +466,37 @@ impl Collections {
     }
 }
 
+impl AddPrefix for Collections {
```
datanel: I wonder if this is the good place

src/transxchange/read.rs
```diff
@@ -92,13 +92,14 @@ where
...
 /// Read TransXChange format into a Navitia Transit Model
-pub fn read<P>(transxchange_path: P, naptan_path: P) -> Result<Model>
+pub fn read<P>(transxchange_path: P, naptan_path: P, prefix: Option<String>) -> Result<Model>
 where
     P: AsRef<Path>,
 {
     let mut collections = Collections::default();
     naptan::read_naptan(naptan_path, &mut collections)?;
     read_transxchange_archive(transxchange_path, &mut collections)?;
+    prefix.map(|p| collections.add_prefix(p.as_str()));
```
datanel: generally map is used to return non unit `()` value
you can do 
```rust
if let Some(prefix) = prefix {
    collections.add_prefix(p.as_str());
}
```

src/model.rs
woshilapin: Would you do it in `src/add_prefix.rs` as I did for `Collection` and `CollectionWithId`?

src/transxchange/read.rs
woshilapin: I knew I was doing something weird and yet my brain didn't provide the easy answer to that... Thank you. Fixed in 2250daeb61de84a7e15340e610c9b59db8220b6e.

woshilapin:
I added the blanket implementation for `add_prefix_with_sep` in fb0858d854a6075ad80dfc16aa95fdd9f76e0943.

src/model.rs
woshilapin: I moved the implementation of `AddPrefix` for `Collections` into `add_prefix.rs` in fb0858d854a6075ad80dfc16aa95fdd9f76e0943.

src/add_prefix.rs
```diff
@@ -0,0 +1,128 @@
...
+        }
+    }
+}
+
+impl<T> AddPrefix for CollectionWithId<T>
+where
+    T: Id<T> + AddPrefix,
+{
+    fn add_prefix(&mut self, prefix: &str) {
+        let indexes: Vec<_> = self.get_id_to_idx().values().cloned().collect();
```
datanel: I wonder if it's better like that

```suggestion
        let indexes: Vec<_> = self.iter().map(|(idx, _)| idx).collect();
```

woshilapin: Yes, I think it is for 2 reasons:
- No explicit clone (the copy of `Idx` happened in the `iter()` though)
- Not using this internal map `id_to_idx` which I'm not even sure why we expose it through the function `get_id_to_idx()`

Thank you :smiley:

PR #303, opened 2019-07-30
Allows Naptan in zip or folder format

ArnaudOggy:


ArnaudOggy:
> How does it handle the case when a non-zip file is sent? Does it panic on [`src/transxchange/naptan.rs#L226`](https://github.com/CanalTP/transit_model/pull/303/files#diff-9c970ed467ca42773ff2ca90a574e331R226) with a useful message?

It panic with the message "Invalid Zip archive".

PR #304, opened 2019-07-31
temporarily add read/write ntfs zone_id

datanel:
### :unamused:


src/ntfs/mod.rs
```diff
@@ -515,6 +516,7 @@ mod tests {
                 stop_area_id: "OIF:SA:8739322".to_string(),
                 timezone: Some("Europe/Paris".to_string()),
                 fare_zone_id: Some("1".to_string()),
+                zone_id: None,
```
woshilapin: Not necessary, the `..Default::default()` below will take care of that for you!

src/ntfs/mod.rs
```diff
@@ -762,6 +764,7 @@ mod tests {
                 equipment_id: Some("equipment_1".to_string()),
                 stop_area_id: "sa_1".to_string(),
                 fare_zone_id: Some("1".to_string()),
+                zone_id: None,
```
woshilapin: Not necessary, the `..Default::default()` below will take care of that for you!

PR #306, opened 2019-08-01
Create Stop Areas for Stop Points missing one

woshilapin:
Whenever a Stop Point is parsed from NaPTAN dataset which doesn't have a corresponding Stop Area (through `StopsInArea.csv`), we need to create a Stop Area (needed for NTFS) based on the Stop Point information (derived ID, name and coordinates).

### DIFF
--- a/src/transxchange/naptan.rs
+++ b/src/transxchange/naptan.rs
@@ -188,6 +188,7 @@ where
                 stop_areas.push(StopArea {
                     id: id.clone(),
                     name: stop.name.clone(),
+                    visible: true,
                     coord,
                     ..Default::default()
                 })?;
--- a/src/transxchange/naptan.rs
+++ b/src/transxchange/naptan.rs
@@ -154,10 +154,12 @@ where
         .collect()
 }
 
+// Create stop points and create missing stop areas for stop points without
+// a corresponding stop area in NaPTAN dataset
 fn read_stops<R>(
     reader: R,
     stops_in_area: &HashMap<String, String>,
-) -> Result<CollectionWithId<StopPoint>>
+) -> Result<(CollectionWithId<StopPoint>, CollectionWithId<StopArea>)>
 where
     R: Read,
 {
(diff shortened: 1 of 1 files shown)


src/transxchange/naptan.rs
```diff
@@ -356,28 +369,19 @@ mod tests {
...
+
             assert_eq!(stop_points.len(), 2);
             let stop_point = stop_points.get("0100053316").unwrap();
             assert_eq!(stop_point.name, "Broad Walk Shops");
             let stop_point = stop_points.get("0100053264").unwrap();
             assert_eq!(stop_point.name, "Alberton Road");
-        }
 
-        #[test]
-        fn no_stop_area() {
```
woshilapin: Note, I merged that test with the above by removing the stop area from the second stop point. And I assert that the resulting `CollectionWithId<StopArea>` contains the newly created `StopArea`.

src/transxchange/naptan.rs
```diff
@@ -390,19 +394,6 @@ mod tests {
             read_stops(csv_content.as_bytes(), &stop_in_area).unwrap();
         }
 
-        #[test]
-        fn no_id() {
```
woshilapin: It didn't feel like this test made sense anymore...

src/transxchange/naptan.rs
```diff
@@ -166,23 +168,32 @@ where
...
         let coord = Coord {
             lon: stop.longitude,
             lat: stop.latitude,
         };
+        let stop_area_id = match stops_in_area.get(&stop.atco_code).cloned() {
+            Some(stop_area_id) => stop_area_id,
+            None => {
+                // If the stop point don't have a corresponding stop area
+                // create the stop area based on stop point information
+                let id = format!("SA:{}", stop.atco_code);
```
datanel: we can use the same prefix like https://github.com/CanalTP/transit_model/blob/8f7f68b7ec65823278c571d1f6556bdbfbfc7c6f/src/ntfs/read.rs#L94

woshilapin: Yep, fixed.

PR #307, opened 2019-08-01
Allows TransXChange in zip or folder format

ArnaudOggy:


src/transxchange/read.rs
```diff
@@ -141,24 +142,54 @@ fn read_transxchange(transxchange: &Element, collections: &mut Collections) -> R
     unimplemented!()
 }
 
-fn read_transxchange_archive<P>(transxchange_path: P, collections: &mut Collections) -> Result<()>
+fn read_transxchange_from_zip<P>(transxchange_path: P, collections: &mut Collections) -> Result<()>
```
datanel: you are in transxchange module so `read_from_path` is enough

same for `read_transxchange_from_zip`

src/transxchange/read.rs
```diff
@@ -141,24 +142,54 @@ fn read_transxchange(transxchange: &Element, collections: &mut Collections) -> R
...
 {
     let zip_file = File::open(transxchange_path)?;
     let mut zip_archive = ZipArchive::new(zip_file)?;
     for index in 0..zip_archive.len() {
-        let mut zip_file = zip_archive.by_index(index)?;
-        match zip_file.sanitized_name().extension() {
+        let mut file = zip_archive.by_index(index)?;
+        let file_name = file.sanitized_name();
+        let file_extension = file_name.extension();
+        match file_extension {
```
datanel: `match file_name.extension()`

src/transxchange/read.rs
```diff
@@ -141,24 +142,54 @@ fn read_transxchange(transxchange: &Element, collections: &mut Collections) -> R
...
+{
+    for entry in WalkDir::new(transxchange_path)
+        .into_iter()
+        .filter_map(|e| e.ok())
+        .filter(|e| e.file_type().is_file())
+    {
+        let mut file = File::open(entry.path())?;
+        let file_name = entry.file_name();
+        let file_extension = entry.path().extension();
+        match file_extension {
```
datanel: `match entry.path().extension()`

src/transxchange/read.rs
```diff
@@ -141,24 +142,54 @@ fn read_transxchange(transxchange: &Element, collections: &mut Collections) -> R
...
+    Ok(())
+}
+
+fn read_transxchange_from_path<P>(transxchange_path: P, collections: &mut Collections) -> Result<()>
+where
+    P: AsRef<Path>,
+{
+    for entry in WalkDir::new(transxchange_path)
+        .into_iter()
+        .filter_map(|e| e.ok())
```
datanel: what happens if it is not ok?

ArnaudOggy: > Or, if you'd like to iterate over all entries and ignore any errors that may arise, use filter_map. (e.g., This code below will silently skip directories that the owner of the running process does not have permission to access.)

https://github.com/BurntSushi/walkdir

datanel: do we want to ignore bad files?

datanel: you should at least log what file is bad

PR #308, opened 2019-08-01
Specs TXC: adding routes

papailio:
Picking the specs TXC up where I left off, starting with routes.

### DIFF
--- a/src/documentation/transxchange_read.md
+++ b/src/documentation/transxchange_read.md
@@ -19,6 +19,12 @@ An additional data source is necessary in order to retrieve the information rela
 stops used in the TransXChange feed. The National Public Transport Access Nodes ([NaPTAN](http://naptan.app.dft.gov.uk/DataRequest/Naptan.ashx?format=csv)) database 
 is a UK nationwide system for uniquely identifying all the points of access to public transport in the UK.
 
+This version of the connector handles TranXChange files with a single *Service* 
+specifying one or more *Lines*. Multiple *Service*s will be possibly considered in a 
+later version. Also, on-demand-transport services as well as frequency-based trips 
+are not handled; therefore, any input feed that contains a *Service/FlexibleService* 
+or a *VehicleJourney/Frequency* will be ignored.
+
 ## Mapping between TransXChange elements and NTFS objects
 ### networks.txt
 
@@ -71,10 +77,11 @@ stop_lon | StopAreas.csv | *Easting* | The input coordinate system Easting/North
 location_type | | | Fixed value `1`.
 
 ### lines.txt
+A *Services/Service* in the input feed might specify one or more *Lines*.
 
 NTFS field | TransXChange element | Mapping rule/Comment
 --- | --- | ---
-line_id | *Services/Service/ServiceCode* | This field is prefixed.
+line_id | *Services/Service/ServiceCode*, *Services/Service/Line{id}* | This field is prefixed and formed by the concatenation of the two fields separated by a `:`. Ex. "\<prefix>:1_58_BC:SL1".
 line_code | *Services/Service/Lines/Line/LineName* |
 line_name | *Services/Service/Description* | If no *Description* is found, then this field is computed using the name of the first associated Route in the forward direction. If several forward routes exist, the one with the smallest `route_id` is used.
 forward_line_name | *Services/Service/StandardService/Destination* | 
@@ -103,10 +110,10 @@ A Route is created from a line and a direction of journey patterns (*StandardSer
 
 NTFS field | TransXChange element | Mapping rule/Comment
 --- | --- | ---
-route_id | *Services/Service/ServiceCode*, *StandardService/JourneyPattern/Direction* | This field is prefixed and formed by the concatenation of *Services/Service/ServiceCode* and *StandardService/JourneyPattern/Direction* separated by a `:`. Ex. "\<prefix>:1_58_BC:inbound".
+route_id | *Services/Service/ServiceCode*, *Services/Service/Line{id}* *StandardService/JourneyPattern/Direction* | This field is prefixed and formed by the concatenation of *Services/Service/ServiceCode*, *Services/Service/Line{id}* and *StandardService/JourneyPattern/Direction* separated by a `:`. Ex. "\<prefix>:1_58_BC:SL1:inbound".
 route_name |  | "[first stop of the first trip] - [last stop of the first trip]" (1)
 direction_type | *StandardService/JourneyPattern/Direction* | The value is set to `inbound` or `clockwise` when the specified value for the direction is `inboundAndOutbound` or `circular`, respectively.
-line_id | *Services/Service/ServiceCode* | This field is prefixed. Link to the file [lines.txt](#linestxt).
+line_id | *Services/Service/ServiceCode*, *Services/Service/Line{id}* | This field is prefixed. Link to the file [lines.txt](#linestxt).
 destination_id |  | `stop_id` of the stop_area of the last stop of the first trip (1). Link to the file [stops.txt](#stopstxt).
 
 (1) The first trip (in alphabetical order) of a route is the one with the smallest `trip_id` value.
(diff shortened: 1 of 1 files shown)


src/documentation/transxchange_read.md
```diff
@@ -71,11 +71,11 @@ NTFS field | TransXChange element | Mapping rule/Comment
 --- | --- | ---
 line_id | *Services/Service/ServiceCode* | This field is prefixed.
 line_code | *Services/Service/Lines/Line/LineName* |
-line_name | *Services/Service/Description* | If no *Description* is found, the value is computed as "backward_line_name - forward_line_name" 
+line_name | *Services/Service/Description* | If no *Description* is found, then this field is computed using the name of the first associated Route in the forward direction. If several forward routes exist, the one with the smallest `route_id` is used.
```
woshilapin: > The one with the smallest `route_id` is used

Is it about the length of String? What if there is 2 routes with a `route_id` of the same length?

It feels like it doesn't matter which one we're choosing if several forward routes are found, the only constraint might be that it is consistently the same one that is chosen, am I right?

src/documentation/transxchange_read.md
```diff
@@ -92,8 +92,17 @@ tram | Tramway | Tramway| Tramway| Tramway
...
+
+NTFS field | TransXChange element | Mapping rule/Comment
+--- | --- | ---
+route_id | *Services/Service/ServiceCode*, *StandardService/JourneyPattern/Direction* | This field is prefixed and formed by the concatenation of *Services/Service/ServiceCode* and *StandardService/JourneyPattern/Direction* separated by a `:`. Ex. "\<prefix>:1_58_BC:inbound".
+route_name |  | "[first stop of the first trip] - [last stop of the first trip]" (1)
+direction_type | *StandardService/JourneyPattern/Direction* | The value is set to `inbound` or `clockwise` when the specified value for the direction is `inboundAndOutbound` or `circular`, respectively.
+line_id | *Services/Service/ServiceCode* | This field is prefixed. Link to the file [lines.txt](#linestxt).
+destination_id | *???* | `stop_id` of the stop_area of the last stop of the first trip. (1)
+
+(1) The first trip of a route is the one with the smallest `trip_id` value.
```
woshilapin: Basically same remark than in [this comment](https://github.com/CanalTP/transit_model/pull/308/files#r310494349).

src/documentation/transxchange_read.md
```diff
@@ -92,8 +92,17 @@ tram | Tramway | Tramway| Tramway| Tramway
...
 ### routes.txt
-A Route is created from a line and a direction of journey patterns (*StandardService/JourneyPattern/Direction*/).
-Routes might not be specified at all in the xml --> reconstruct routes after trips?
+A Route is created from a line and a direction of journey patterns (*StandardService/JourneyPattern/Direction*).
+
+NTFS field | TransXChange element | Mapping rule/Comment
+--- | --- | ---
+route_id | *Services/Service/ServiceCode*, *StandardService/JourneyPattern/Direction* | This field is prefixed and formed by the concatenation of *Services/Service/ServiceCode* and *StandardService/JourneyPattern/Direction* separated by a `:`. Ex. "\<prefix>:1_58_BC:inbound".
+route_name |  | "[first stop of the first trip] - [last stop of the first trip]" (1)
+direction_type | *StandardService/JourneyPattern/Direction* | The value is set to `inbound` or `clockwise` when the specified value for the direction is `inboundAndOutbound` or `circular`, respectively.
```
woshilapin: What if it's another value? We just take the same value as `StandardService/JourneyPattern/Direction`?

src/documentation/transxchange_read.md
```diff
@@ -92,8 +92,17 @@ tram | Tramway | Tramway| Tramway| Tramway
 trolleyBus | Shuttle | Shuttle | Shuttle | Shuttle
 
 ### routes.txt
-A Route is created from a line and a direction of journey patterns (*StandardService/JourneyPattern/Direction*/).
-Routes might not be specified at all in the xml --> reconstruct routes after trips?
+A Route is created from a line and a direction of journey patterns (*StandardService/JourneyPattern/Direction*).
+
+NTFS field | TransXChange element | Mapping rule/Comment
+--- | --- | ---
+route_id | *Services/Service/ServiceCode*, *StandardService/JourneyPattern/Direction* | This field is prefixed and formed by the concatenation of *Services/Service/ServiceCode* and *StandardService/JourneyPattern/Direction* separated by a `:`. Ex. "\<prefix>:1_58_BC:inbound".
```
woshilapin: Do you think it would be correct to specify this as
```suggestion
route_id | line_id, *StandardService/JourneyPattern/Direction* | This field is prefixed and formed by the concatenation of the line ID and *StandardService/JourneyPattern/Direction* separated by a `:`. Ex. "\<prefix>:1_58_BC:inbound".
```
It feels that if we ever change what is going to define what is a line ID, then we're gonna need to adapt the route ID... unless we define them already based on the line ID. Would it be correct to say that?

src/documentation/transxchange_read.md
woshilapin: What if, in the same file, I find 2 `StandardService/JourneyPattern`, first one having direction `inbound` and the second one having the direction `inboundAndOutbound`? It will create the same ID for both routes hence a collision.

woshilapin: What if I have 2 `StandardService/JourneyPattern`, the first one having a direction `inbound` and the second one having a direction `inboundAndOutbound`. Is that fine that in the end, we'll have 2 routes, both `inbound` on the same line? The IDs won't collide because they will be `line_id:inbound` and `line_id:inboundAndOutbound`.

src/documentation/transxchange_read.md
papailio: It is the "smallest" in alphabetical order mostly, I will add a precision in the next commit. And yes, the idea is that the first trip is consistently chosen in the same way.

src/documentation/transxchange_read.md
papailio: For now, I would be ok with 2 distinct routes (different ids) with the same direction inbound (or circular).

src/documentation/transxchange_read.md
datanel: `line_id` is more related to the implementation. I would leave `Services/Service/ServiceCode` to not mix things coming from NTFS and transxchange

woshilapin: We discussed it with @papailio and `line_id` is not gonna be used.

src/documentation/transxchange_read.md
```diff
@@ -71,11 +71,11 @@ NTFS field | TransXChange element | Mapping rule/Comment
...
 line_id | *Services/Service/ServiceCode* | This field is prefixed.
 line_code | *Services/Service/Lines/Line/LineName* |
-line_name | *Services/Service/Description* | If no *Description* is found, the value is computed as "backward_line_name - forward_line_name" 
+line_name | *Services/Service/Description* | If no *Description* is found, then this field is computed using the name of the first associated Route in the forward direction. If several forward routes exist, the one with the smallest `route_id` is used.
 forward_line_name | *Services/Service/StandardService/Destination* | 
-forward_direction | *???* | 
+forward_direction |  | This field should have the same value as the `destination_id` of the first associated Route in the forward direction. If several forward routes exist, the one with the smallest `route_id` is used.
 backward_line_name | *Services/Service/StandardService/Origin* | 
-backward_direction | *???* | 
+backward_direction |  | This field should have the same value as the `destination_id` of the first associated Route in the backward direction. If several backward routes exist, the one with the smallest `route_id` is used.
```
ArnaudOggy: It must be origin_id I think

src/documentation/transxchange_read.md
```diff
@@ -91,14 +96,25 @@ rail | Train | Train | Train | Train
...
-A Route is created from a line and a direction of journey patterns (*StandardService/JourneyPattern/Direction*/).
-Routes might not be specified at all in the xml --> reconstruct routes after trips?
+A Route is created from a line and a direction of journey patterns (*StandardService/JourneyPattern/Direction*).
+
+NTFS field | TransXChange element | Mapping rule/Comment
+--- | --- | ---
+route_id | *Services/Service/ServiceCode*, *StandardService/JourneyPattern/Direction* | This field is prefixed and formed by the concatenation of *Services/Service/ServiceCode* and *StandardService/JourneyPattern/Direction* separated by a `:`. Ex. "\<prefix>:1_58_BC:inbound".
+route_name |  | "[first stop of the first trip] - [last stop of the first trip]" (1)
+direction_type | *StandardService/JourneyPattern/Direction* | The value is set to `inbound` or `clockwise` when the specified value for the direction is `inboundAndOutbound` or `circular`, respectively.
+line_id | *Services/Service/ServiceCode* | This field is prefixed. Link to the file [lines.txt](#linestxt).
```
woshilapin: I think this will need to use `LineRef` to find the correct line and then concatenate the `ServiceCode` + `Line/id`.

src/documentation/transxchange_read.md
```diff
@@ -19,16 +19,27 @@ An additional data source is necessary in order to retrieve the information rela
 stops used in the TransXChange feed. The National Public Transport Access Nodes ([NaPTAN](http://naptan.app.dft.gov.uk/DataRequest/Naptan.ashx?format=csv)) database 
 is a UK nationwide system for uniquely identifying all the points of access to public transport in the UK.
 
+This version of the connector handles TranXChange files with a single *Service* 
+specifying one or more *Lines*. Multiple *Service*s will be possibly considered in a
```
woshilapin:
```suggestion
specifying one or more *Lines*. Multiple *Services* will be possibly considered in a 
```

src/documentation/transxchange_read.md
```diff
@@ -107,16 +130,68 @@ Similarly, the element *BankHolidaysOperation* may be also be present, specifyin
...
+calculated from the cumulative sum of the current *JourneyPatternTimingLink* values 
+for all preceding stops in the journey link sequence as follows:
+- arrival_time at stoppoint<sub>n</sub> = departure_time from stop<sub>n-1</sub> + (*RunTime* for inbound link from stop<sub>n-1</sub>)
+- departure_time at stop<sub>n</sub> = arrival_time at stop<sub>n</sub> + *WaitTime* for destination end of inbound link from stop<sub>n-1</sub>) + *WaitTime* for origin of outbound link to stop<sub>n+1</sub>
+
+If *WaitTime* is not specified, the default waiting time at a stoppoint is considered `0`.
+
+Note that *RunTime* and *WaitTime* are given as [durations](https://en.wikipedia.org/wiki/ISO_8601#Durations).
+
+Except from the first stoppoint of the trip, the arrival end (*JourneyPatternTimingLink/To*)
```
ArnaudOggy: We reversed with the use of JourneyPatternTimingLink/From, so the except is on last stop.
No problem, the result is the same.

src/documentation/transxchange_read.md
```diff
@@ -91,14 +103,25 @@ rail | Train | Train | Train | Train
...
+route_name |  | "[first stop of the first trip] - [last stop of the first trip]" (1)
+direction_type | *StandardService/JourneyPattern/Direction* | The value is set to `inbound` or `clockwise` when the specified value for the direction is `inboundAndOutbound` or `circular`, respectively.
+line_id | *Services/Service/ServiceCode*, *Services/Service/Line{id}* | This field is prefixed. Link to the file [lines.txt](#linestxt).
+destination_id |  | `stop_id` of the stop_area of the last stop of the first trip (1). Link to the file [stops.txt](#stopstxt).
+
+(1) The first trip (in alphabetical order) of a route is the one with the smallest `trip_id` value.
 
 ### calendar_dates.txt
 The validity period of a service is stated in *Services/Service/OperatingPeriod*. In case the validity period is open ended (the *EndDate* is not specified), the default value [*StartDate* + 180 days] should be used.
```
woshilapin: Some of the data are showing `EndDate` like `9999-01-01`. The proposition in this case is to limit to `StartDate + 2 year`.

Note: It is mostly a technical reason due to the fact that the connector is creating a list of all the dates.

src/documentation/transxchange_read.md
prhod: @ArnaudOggy ot's OK like that, it's the destination of the backward route

src/documentation/transxchange_read.md
prhod: As the `StartDate` could be far in the past, we cannot use a `StartDate + 2Y`. Instead, will add a mandatory parameter specifying the EndDate of the loaded data to specify (or truncate) calendars.

PR #309, opened 2019-08-02
Support for multiple networks and companies

woshilapin:
This will get the operator reference from `<Service>` in order to generate the Network.

For companies, all `Operator` will generate a new company.

### DIFF
--- a/src/transxchange/read.rs
+++ b/src/transxchange/read.rs
@@ -26,6 +26,7 @@ use chrono::{
     naive::{MAX_DATE, MIN_DATE},
     Duration,
 };
+use failure::format_err;
 use log::info;
 use minidom::Element;
 use std::{fs::File, io::Read, path::Path};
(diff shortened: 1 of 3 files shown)


src/transxchange/read.rs
```diff
@@ -101,16 +102,45 @@ fn update_validity_period_from_transxchange(
     CollectionWithId::new(datasets)
 }
 
-fn get_operator(root: &Element) -> Result<&Element> {
-    root.try_only_child("Operators")?.try_only_child("Operator")
+fn get_operator<'a>(transxchange: &'a Element, operator_ref: &String) -> Result<&'a Element> {
```
datanel:
```suggestion
fn get_operator<'a>(transxchange: &'a Element, operator_ref: &str) -> Result<&'a Element> {
```

src/transxchange/read.rs
```diff
@@ -101,16 +102,45 @@ fn update_validity_period_from_transxchange(
...
+    let is_operator_ref = |operator: &&Element| {
+        operator
+            .attr("id")
+            .map(|id| id == operator_ref)
+            .unwrap_or(false)
```
datanel:
```rust
operator
            .attr("id")
            .filter(|id| id == &operator_ref)
            .is_some()
```

src/transxchange/read.rs
```diff
@@ -101,16 +102,45 @@ fn update_validity_period_from_transxchange(
...
+    };
+    transxchange
+        .try_only_child("Operators")?
+        .children()
+        .filter(is_operator_ref)
```
datanel: you can use `find` and remove `next`

src/transxchange/read.rs
woshilapin: Any reason for that? Technically, I know a `&String` is coming as an input argument. If I change the type to `&str`, it will force a deref from `&String` to `&str` which I believe is an unnecessary operation at this point (it will happen eventually below when I do the comparison `==` but I'd prefer to let the compiler dereference at the last moment). But this is the kind of comment I like because I might have completely miss something else, did I?

datanel: writing `&String` instead of `&str` involves a new object where a slice will do.

src/transxchange/read.rs
woshilapin: Done in 28c8cffa0c21ba886ef49e5b45af5b32ee5677bc.

src/transxchange/read.rs
woshilapin: Done in 680f032ff00836814abc44a95d87b02f8ec50dcc.

src/transxchange/read.rs
woshilapin: A slice is an object containing a pointer and a length ([slices](https://doc.rust-lang.org/stable/book/ch04-03-slices.html?highlight=slice#string-slices)), a reference is an object, probably containing only a pointer ([reference](https://doc.rust-lang.org/stable/book/ch04-02-references-and-borrowing.html#references-and-borrowing)). So technically, having a reference is cheaper than a slice. Because the performance impact is probably negligeable, I'll change the code but this is still a piece of Rust where I feel I need to dig more to understand more.
I've done the modification in 8b9c545be58c249d7611e5ebebe39cc33d1aa7b6.

datanel: these are just suggestions from clippy https://rust-lang.github.io/rust-clippy/master/index.html#ptr_arg

generally you never uses &String but &str, just if you want to mutate it with &mut String

I've never found a good explication between &String and &str so clippy seems a good referee to me.

PR #310, opened 2019-08-02
Name the Travis job for 'cargo fmt'

woshilapin:


.travis.yml
```diff
@@ -15,6 +15,7 @@ addons:
 matrix:
   include:
   - rust: stable
+    name: Format
```
datanel: it does not format but check the formatting

woshilapin: So `Format check`?

datanel: formatting check? or check of formatting?
format sounds like paper format

woshilapin: Changed to `Formatting check`.

PR #311, opened 2019-08-05
Create workspaces

datanel:
first step before moving everything in its crate
just moved **src, tests, examples, fixtures** to **transit_model** folder and created workpace

```
├── Cargo.lock
├── Cargo.toml
├── Dockerfile
├── LICENSE
├── model-builder
├── README.md
├── target
├── transit_model
└── transit_model_procmacro
```

we need a reflection on how to create git releases and how to deploy


woshilapin:
It is not necessary to move everything up a level. I was thinking more along the line where `transit_model` core library stay at the root level and we create sub-folders for the part we want to extract.

I've done an attempt in a branch `extract-kv1` and everything seems to build (and test) fine (see https://github.com/CanalTP/transit_model/compare/master...woshilapin:extract-kv1?expand=1).

A few notes:
- we have to use the `default-members` in this case from what I understand of the [documentation](https://doc.rust-lang.org/cargo/reference/manifest.html#package-selection)
- a lot of macros and functions have to be made public so a big part of the above diff is about exposing and/or documenting existing code (we might be able to make a first PR to do that first)

NB: I did that work so we can discuss the 2 options. I personally prefer to only extract what needs to be extract and do not move what can be left in place. It also makes sense to not have a `transit_model/transit_model/` folder. Last reason, it'll make our process of refactoring easier I believe (moving all the files will cause merging problems will all the current PR).

datanel:
Having tansit_model (with tests fixtures...) in its folder is cleaner
I was inspired by many projects like serde.

datanel:
I close this PR. we need more reflection about workspaces

PR #313, opened 2019-08-09
Make 'read_config' return unique contributor/dataset

woshilapin:
[`read_config`](https://github.com/CanalTP/transit_model/compare/master...woshilapin:read_config?expand=1#diff-a5ae51324be0fd74d0e1e9788b025f85R45) function is not designed to create multiple Contributors or Datasets from a `config.json` but only one of each. So there is no reason the signature of this function returns `CollectionWithId`.

NB: This modification will be useful in the TransXChange convertor implementation.

### DIFF
--- a/src/read_utils.rs
+++ b/src/read_utils.rs
@@ -45,8 +45,8 @@ struct Config {
 pub fn read_config<P: AsRef<path::Path>>(
     config_path: Option<P>,
 ) -> Result<(
-    CollectionWithId<objects::Contributor>,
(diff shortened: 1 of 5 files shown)


src/read_utils.rs
```diff
@@ -45,8 +45,8 @@ struct Config {
 pub fn read_config<P: AsRef<path::Path>>(
```
woshilapin: Modification to the signature of this function is the only important part of this PR. The rest of the modification is only a consequence.

PR #314, opened 2019-08-09
Implementation of From<T> for Collection<T> and CollectionWithId<T>

woshilapin:
Often, we want to create a `Collection` or a `CollectionWithId` out of only one element. Implementation of `From<T> for Collection<T>` and `From<T> for CollectionWithId<T>` should help with that.

### DIFF
--- a/src/collection.rs
+++ b/src/collection.rs
@@ -213,7 +213,7 @@ impl<T> Collection<T> {
     /// # #[derive(PartialEq, Debug)] struct Obj(&'static str);
     /// # impl Id<Obj> for Obj { fn id(&self) -> &str { self.0 }
     /// # fn set_id(&mut self, id: String) { unimplemented!(); }}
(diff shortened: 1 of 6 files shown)


src/collection.rs
```diff
@@ -361,6 +378,38 @@ pub struct CollectionWithId<T> {
     id_to_idx: HashMap<String, Idx<T>>,
 }
 
+/// Creates a `CollectionWithId` from one element.
+///
+/// This cannot fail since there will be a unique identifier in the collection
```
datanel: you are right, that's why you can use `CollectionWithId::new(vec![object]).unwrap()` without hesitation

woshilapin: Done in 1e0da0a40a1a1207612bbd045ae491625908c34f.

datanel:
yo have to rebase

woshilapin:
I just rebased, it should be good now.

PR #315, opened 2019-08-09
Use tempfile crate instead of deprecated tempdir crate

datanel:
https://github.com/rust-lang-deprecated/tempdir

### DIFF
--- a/src/ntfs/mod.rs
+++ b/src/ntfs/mod.rs
@@ -22,7 +22,6 @@ pub mod filter;
 mod read;
 mod write;
 
-use self::tempdir::TempDir;
 use crate::common_format;
 use crate::model::{Collections, Model};
 use crate::objects::*;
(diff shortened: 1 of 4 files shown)


src/ntfs/mod.rs
```diff
@@ -233,9 +232,10 @@ pub fn write_to_zip<P: AsRef<path::Path>>(
 ) -> Result<()> {
     let path = path.as_ref();
     info!("Writing NTFS to ZIP File {:?}", path);
-    let input_tmp_dir = TempDir::new("write_ntfs_for_zip")?;
+    let input_tmp_dir = tempdir()?;
     write(model, input_tmp_dir.path(), current_datetime)?;
     zip_to(input_tmp_dir.path(), path)?;
+    input_tmp_dir.close()?;
```
woshilapin: I'm not sure you need that because the `Drop` trait will take care of that. See [documentation](https://docs.rs/tempfile/3.1.0/tempfile/struct.TempDir.html#impl-Drop).

woshilapin: OK, after discussion, it's avoiding to fail silently when the clean up is not done. So I'm fine with that.

PR #316, opened 2019-08-09
removed binaries

datanel:
Binaries are (will be) now in https://github.com/CanalTP/tartare-tools

- [x] remove transXchange bin when it is completely implemented
- [x] delete jenkins job for building docker image


README.md
```diff
@@ -95,24 +89,6 @@ From the standard, some of the functionalities are not fully supported:
   `grid_exception_dates.txt`, `grid_periods.txt` and `grid_rel_calendar_line.txt`)
 - No support for Line Groups (files `line_groups.txt` and `line_group_links.txt`)
 
-## FAQ
```
woshilapin: Good catch, indeed, this section is not necessary anymore.

woshilapin:
Nice job about moving around the documentation. When looking, I see that you moved all the deprecated things about fares but file renaming seems inconsistent:
- `ntfs_deprecated_fares_reading.md`
- `deprecated_ntfs_fares_writing.md`

Also, in the first one, you edited the content to replace `fares` with `deprecated fares` (at least in the title).  Might be nice to do the same in the second file too.

datanel:
@woshilapin done

PR #317, opened 2019-08-09
Enhance TryOnlyChild with filters

woshilapin:
In transXChange, we're often gonna select one of the children based on it's `id` attribute. Giving this filtering ability is gonna simplify some node selection.

### DIFF
--- a/src/minidom_utils/try_only_child.rs
+++ b/src/minidom_utils/try_only_child.rs
@@ -26,12 +26,26 @@ use minidom::Element;
 /// fail)
 pub trait TryOnlyChild {
     /// Try to get an unique child from its name and return a [Result](crate::Result)
-    fn try_only_child<'a>(&'a self, child_name: &str) -> Result<&'a Element>;
+    /// A filter can be apply on the kind of children you want to select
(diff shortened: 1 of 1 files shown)


src/minidom_utils/try_only_child.rs
```diff
@@ -26,12 +26,26 @@ use minidom::Element;
...
 impl TryOnlyChild for Element {
-    fn try_only_child<'a>(&'a self, child_name: &str) -> Result<&'a Element> {
-        let mut child_iterator = self.children().filter(|child| child.name() == child_name);
+    fn try_only_child_with_filter<'a, P>(&'a self, child_name: &str, filter: P) -> Result<&'a Self>
+    where
+        P: Fn(&'a Self) -> bool,
+    {
+        let mut child_iterator = self
+            .children()
+            .filter(|child| child.name() == child_name)
```
woshilapin: I should also be able to do something like
```
.filter(|child| child.name() == child_name && filter(*child))
```
instead of chaining 2 `.filter()` operations.  Not sure which is best.

PR #323, opened 2019-08-22
Generate Calendar Dates for TransXChange

woshilapin:


src/transxchange/read.rs
```diff
@@ -277,15 +283,139 @@ fn create_route(_transxchange: &Element, _vehicle_journey: &Element) -> Result<R
...
 }
 
+struct NaiveDateRange {
+    current: NaiveDate,
+    end: NaiveDate,
+}
+
+impl NaiveDateRange {
+    fn new(start: NaiveDate, end: NaiveDate) -> NaiveDateRange {
+        if start <= end {
```
ArnaudOggy: Limit the endDate to Y+2 to avoid wtf dates (like 2099, 2500, 3019, 9999, etc.)

src/transxchange/read.rs
```diff
@@ -442,7 +562,11 @@ fn load_routes_vehicle_journeys_calendars(
         let line_ref = vehicle_journey.try_only_child("LineRef")?.text();
         let vehicle_journey_code = vehicle_journey.try_only_child("VehicleJourneyCode")?.text();
         let id = format!("{}:{}:{}", service_ref, line_ref, vehicle_journey_code);
-        let calendar = create_calendar_dates(transxchange, vehicle_journey)?;
+        let dates = create_calendar_dates(transxchange, vehicle_journey)?;
+        let calendar = Calendar {
+            id: format!("CD:{}", id),
```
ArnaudOggy: We keep CD or not ? :confused:

prhod: ok to keep it like this :)

ArnaudOggy: @woshilapin In my opinion we should skip the vj if there is no date (like for stop_times).

src/transxchange/read.rs
```diff
@@ -277,15 +283,139 @@ fn create_route(_transxchange: &Element, _vehicle_journey: &Element) -> Result<R
...
+                end: start,
+            }
+        }
+    }
+}
+
+impl Iterator for NaiveDateRange {
+    type Item = NaiveDate;
+
+    fn next(&mut self) -> Option<Self::Item> {
```
datanel: what's the need to have an iterator on only one `NaiveDateRange` ?
you can't do that in `new`?

woshilapin: I'm not sure I understand the comment here... Do you say we do not need the `NaiveDateRange` structure and could inline all the code?

datanel: why do you use an iterator for that?

woshilapin: I'm iterating over a list of dates between a start and an end date. Just felt like an iterator was logical for iterating.

src/transxchange/read.rs
woshilapin: Done in 6741ad303819e91e656e07f41c79a44ae4b9212f.

src/transxchange/read.rs
woshilapin: Done in e46aba8bff9fd53a0615f10e39f42d16bc180ad5.

woshilapin:
- I added a support for the case where no tag is present in `DaysOfWeek` (the specification mention that the seven days should be considered). see 8d3b9f42452e61a6fe612f1343f317f504cad14a
- In the case when no calendar date is generated, the whole vehicle journey should be ignore. see 6741ad303819e91e656e07f41c79a44ae4b9212f
- For the End Date not specified (or WTF-ish like `9999-01-01`), I'm now using the today's date + 2 years as an upper bound. see e46aba8bff9fd53a0615f10e39f42d16bc180ad5

src/transxchange/read.rs
datanel: ok

woshilapin:
I just handled also the case when `RegularDayType` contains `OnlyHolidays` instead of `DaysOfWeek`. see 8bf4bf2c5cf7356507d60f0b29764f7a8187feea

PR #328, opened 2019-08-27
Factorize the StopArea creation for orphan StopPoint

woshilapin:
Not sure about the solution is optimal but here is a first draft. I'm open to suggestion.

### DIFF
--- a/src/gtfs/read.rs
+++ b/src/gtfs/read.rs
@@ -561,14 +561,16 @@ where
         let equipment_id = get_equipment_id_and_populate_equipments(equipments, &stop);
         match stop.location_type {
             StopLocationType::StopPoint => {
-                if stop.parent_station.is_none() {
-                    let mut new_stop_area = stop.clone();
-                    new_stop_area.id = format!("Navitia:{}", new_stop_area.id);
-                    new_stop_area.code = None;
-                    stop.parent_station = Some(new_stop_area.id.clone());
-                    stop_areas.push(objects::StopArea::from(new_stop_area));
-                }
-                let mut stop_point = <objects::StopPoint>::from(stop);
+                let mut stop_point = if stop.parent_station.is_none() {
+                    stop.parent_station = Some(String::from("default_id"));
+                    let mut stop_point = objects::StopPoint::from(stop);
+                    let stop_area = objects::StopArea::from(stop_point.clone());
+                    stop_point.stop_area_id = stop_area.id.clone();
+                    stop_areas.push(stop_area);
+                    stop_point
+                } else {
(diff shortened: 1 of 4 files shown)


src/gtfs/read.rs
```diff
@@ -561,14 +561,16 @@ where
...
-                if stop.parent_station.is_none() {
-                    let mut new_stop_area = stop.clone();
-                    new_stop_area.id = format!("Navitia:{}", new_stop_area.id);
-                    new_stop_area.code = None;
-                    stop.parent_station = Some(new_stop_area.id.clone());
-                    stop_areas.push(objects::StopArea::from(new_stop_area));
-                }
-                let mut stop_point = <objects::StopPoint>::from(stop);
+                let mut stop_point = if stop.parent_station.is_none() {
+                    stop.parent_station = Some(String::from("default_id"));
```
datanel: I don't think this line is needed. and you can remove `mut` from `for mut stop`

src/ntfs/read.rs
```diff
@@ -88,14 +88,18 @@ pub fn manage_stops(collections: &mut Collections, path: &path::Path) -> Result<
...
-                if stop.parent_station.is_none() {
-                    let mut new_stop_area = stop.clone();
-                    new_stop_area.visible = area_visibility;
-                    new_stop_area.id = format!("Navitia:{}", new_stop_area.id);
-                    stop.parent_station = Some(new_stop_area.id.clone());
-                    stop_areas.push(StopArea::from(new_stop_area));
-                }
-                stop_points.push(StopPoint::from_with_type(stop, stop_type));
+                let stop_point = if stop.parent_station.is_none() {
+                    stop.parent_station = Some(String::from("default_id"));
```
datanel: same here

src/gtfs/read.rs
woshilapin: I thought I could too and I tried. But we're calling `object::StopPoint::from(stop)` just the line after and here is what this function is doing
```
  stop_area_id: stop.parent_station.unwrap(),
```
So `objects::StopPoint::from()` can't be used unless `parent_station` is `Some` (or it will panic). But at the same time, I cannot do `objects::StopArea::from(stop_point)` if I don't already have an existing `objects::StopPoint`. Kind of a chicken and egg problem.

So this is why I proposed the solution of initializing `parent_station` with a dumb value; and I know that immediately after, I'm updating the `stop_area_id` with the correct new value.

src/ntfs/read.rs
woshilapin: See [comment](https://github.com/CanalTP/transit_model/pull/328/files#r318464361).

PR #331, opened 2019-08-27
Add some stop locations enums

datanel:


woshilapin:
Do we plan to do something similar for GTFS?

datanel:
@woshilapin done https://github.com/CanalTP/transit_model/pull/331/commits/2f895602b9748e799a2d9be1ddf5ab5f3af38ff3

woshilapin:
Yipee, I was just asking and then you just did it, wonderful!

src/ntfs/read.rs
```diff
@@ -97,8 +98,16 @@ pub fn manage_stops(collections: &mut Collections, path: &path::Path) -> Result<
                 }
                 stop_points.push(StopPoint::from_with_type(stop, stop_type));
             }
-            1 => stop_areas.push(StopArea::from(stop)),
-            i => warn!("stop.location_type = {} not yet supported, skipping.", i),
+            StopLocationType::StopArea => stop_areas.push(StopArea::from(stop)),
+            StopLocationType::EntranceExit => {
```
woshilapin: I was wondering if something _like_ the following (the code below doesn't build) was possible but not really sure how and it's just a curiosity (I'm fine if we merge as it is).

```
stop_location_type => warn!(
  "stop.location_type = {} not yet supported, skipping.", 
  stop_location_type.serialize(?serializer?)
),
```

woshilapin:
I merged the PR even if the build was broken because:
- the only broken build was `cargo fmt`
- the broken job was broken because of the cache problem
- I checked locally that `cargo fmt` was just fine

PR #333, opened 2019-08-28
[Fix] Suffixing VehicleJourney Id

ArnaudOggy:


src/documentation/transxchange_read.md
```diff
@@ -141,7 +141,7 @@ NTFS field | TransXChange element | Mapping rule/Comment
 --- | --- | ---
 route_id | *Services/Service/ServiceCode*, *Services/Service/Line{id}*, *StandardService/JourneyPattern/Direction* | This field is prefixed and formed by the concatenation of the three elements separated by a `:`. Link to the file [routes.txt](#routestxt).
 service_id | *VehicleJourney/ServiceRef*, *VehicleJourney/LineRef*, *VehicleJourney/VehicleJourneyCode* | This field is prefixed and formed by the concatenation of the three elements separated by a `:` and a prefix `CD`. Ex: "<prefix>:CD:1_58_BC:SL1:VJ1". Link to the file [calendar_dates.txt](#calendar_datestxt). See above for the details about the services dates attached to a trip.
```
prhod: You've added a suffix for trip_id, but the service_id is set with the same value. This new suffix shoud'nt it be also in the service_id ?

ArnaudOggy: Yep :+1:

PR #338, opened 2019-08-29
Bump version 0.6.0

datanel:


model-builder/Cargo.toml
```diff
@@ -1,10 +1,10 @@
...
-version = "0.2.0"
+version = "0.3.0"
 authors = ["Antoine Desbordes <antoine.desbordes@gmail.com>"]
 edition = "2018"
 license = "AGPL-3.0"
 description = "A crate to easily build a transit_model::Model"
 
 [dependencies]
-transit_model = { path = "../", version = "0.5" }
+transit_model = { path = "../", version = "0.6" }
```
woshilapin: Do you think it would make sense to remove the `version` part here?

Cargo.toml
```diff
@@ -62,4 +62,4 @@ required-features = ["proj"]
 [dev-dependencies]
 approx = "0.3.2"
 rust_decimal_macros = "1.0.1"
-transit_model_builder= "0.1.0"
+transit_model_builder = { path = "./model-builder" }
```
woshilapin: How about doing the same for the dependency `transit_model_procmacro` from
```
transit_model_procmacro = { version = "0.1.0", path = "./transit_model_procmacro" }
```
to
```
transit_model_procmacro = { path = "./transit_model_procmacro" }
```

woshilapin: After discussion, only the `dev-dependencies` have been changed, the other, since we're not sure yet of the impact during release, we do not touch for now.

PR #325, opened 2019-08-26
removed read_url utility and reqwest crate 

datanel:
we think that requesting an url to get an archive can be done out of `transit_model`.
`read_url` uses `reqwest` that includes a lot of dependencies. Removing `reqwest` could lighten the build.

@antoine-de, @Tristramg do you use `read_url`?

reqwest dependencies 
```
├── reqwest v0.9.20
│   ├── base64 v0.10.1
│   │   └── byteorder v1.3.2 (*)
│   ├── bytes v0.4.12
│   │   ├── byteorder v1.3.2 (*)
│   │   ├── either v1.5.2
│   │   └── iovec v0.1.2
│   │       └── libc v0.2.60 (*)
│   ├── cookie v0.12.0
│   │   ├── time v0.1.42 (*)
│   │   └── url v1.7.2
│   │       ├── idna v0.1.5
│   │       │   ├── matches v0.1.8
│   │       │   ├── unicode-bidi v0.3.4
│   │       │   │   └── matches v0.1.8 (*)
│   │       │   └── unicode-normalization v0.1.8
│   │       │       └── smallvec v0.6.10
│   │       ├── matches v0.1.8 (*)
│   │       └── percent-encoding v1.0.1
│   ├── cookie_store v0.7.0
│   │   ├── cookie v0.12.0 (*)
│   │   ├── failure v0.1.5 (*)
│   │   ├── idna v0.1.5 (*)
│   │   ├── log v0.4.8 (*)
│   │   ├── publicsuffix v1.5.2
│   │   │   ├── error-chain v0.12.1
│   │   │   │   └── backtrace v0.3.34 (*)
│   │   │   │   [build-dependencies]
│   │   │   │   └── version_check v0.1.5
│   │   │   ├── idna v0.1.5 (*)
│   │   │   ├── lazy_static v1.4.0 (*)
│   │   │   ├── regex v1.2.1 (*)
│   │   │   └── url v1.7.2 (*)
│   │   ├── serde v1.0.99 (*)
│   │   ├── serde_json v1.0.40
│   │   │   ├── itoa v0.4.4 (*)
│   │   │   ├── ryu v1.0.0 (*)
│   │   │   └── serde v1.0.99 (*)
│   │   ├── time v0.1.42 (*)
│   │   ├── try_from v0.3.2
│   │   │   └── cfg-if v0.1.9 (*)
│   │   └── url v1.7.2 (*)
│   ├── encoding_rs v0.8.17 (*)
│   ├── flate2 v1.0.9
[...]

PR #342, opened 2019-09-02
Make GTFS read and FileHandler public

datanel:
`read_url` was removed in PR https://github.com/CanalTP/transit_model/pull/325/files which makes impossible to request and read a gtfs from the outside.

- add a new module with FileHandler trait and its implementations
- made `gtfs::read` public

another solution would be to add a feature (https://github.com/CanalTP/transit_model/pull/343)

### DIFF
--- a/src/common_format.rs
+++ b/src/common_format.rs
@@ -15,9 +15,9 @@
 // <http://www.gnu.org/licenses/>.
 
 use crate::collection::*;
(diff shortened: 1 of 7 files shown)


antoine-de:
I'm not sure a feature is required, I haven't tried using it, but this seems nice.

datanel:
I close this PR until someone needs it.

PR #343, opened 2019-09-02
Made reqwest optional

datanel:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,6 +37,7 @@ transit_model_procmacro = { version = "0.1.0", path = "./transit_model_procmacro
 walkdir = "2.1.4"
 wkt = "0.5.0"
 zip = "0.3.2"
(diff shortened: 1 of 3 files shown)


antoine-de:
that's also good for me, but honestly #342 is enough for our use case if you don't want the burden of a feature.

thanks for all this work!

woshilapin:
I'd be ok with only #342. I agree this would be a nice way to add it back but I really don't think it's necessary. I'm actually thinking that even supporting both ZIP and a folder is already a nice to have. Our API could be much more restrictive and accept only ZIP or only folders (which would result in the dump of either `walk_dir` or `zip` dependencies).

datanel:
+1 for simplicity
-1 for responsability (but finally I don't care if it's optional, it's a good feature)

datanel:
i close this PR until someone really needs it.

PR #344, opened 2019-09-03
Upgrade network/company name if possible

woshilapin:
I did a `trait` for `merge_with()` but I'm not yet sure it's the best option (this is why I have this PR as draft for now).  The idea was to do something similar to the [`Extend`](https://doc.rust-lang.org/std/iter/trait.Extend.html) trait. That said, `MergeWith` is implemented only for `CollectionWithId` so it could just be a new method of this `struct`, without any notion of a `trait` for now.

Another option would be to have the trait defined with a closure that would take only the two elements in conflict (instead of the full collection and the element to merge that is in conflict, see #345).


woshilapin:
I do prefer #345 too in the end. That's why I did it afterwards.

PR #345, opened 2019-09-03
Upgrade network/company name if possible

woshilapin:
This is an alternative to #344 where the `MergeWith` trait is using a closure that takes the two elements in conflict.

### DIFF
--- a/src/collection.rs
+++ b/src/collection.rs
@@ -650,6 +650,69 @@ impl<T: Id<T>> CollectionWithId<T> {
         }
     }
 
+    /// Merge all elements of an `Iterator` into the current `CollectionWithId`.
+    /// If any identifier of an inserted element is already in the collection,
+    /// the closure is called, with first parameter being the element with this
+    /// identifier already in the collection, and the second parameter is the
+    /// element to be inserted.
+    /// ```
+    /// # use transit_model::collection::*;
+    /// #[derive(Debug, Default)]
+    /// struct ObjectId<'a> {
+    ///    id: &'a str,
+    ///    name: &'a str,
+    /// }
+    /// impl Id<ObjectId<'_>> for ObjectId<'_> {
+    ///    fn id(&self) -> &str {
+    ///        self.id
+    ///    }
+    ///    fn set_id(&mut self, _id: String) {
+    ///        unimplemented!()
+    ///    }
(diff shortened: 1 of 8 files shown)


src/merge_with.rs
```diff
@@ -0,0 +1,116 @@
...
+        F: FnMut(&mut T, &T),
+        I: IntoIterator<Item = T>,
+    {
+        for e in iterator {
+            if let Some(mut source) = self.get_mut(e.id()) {
+                use std::ops::DerefMut;
+                f(source.deref_mut(), &e);
+                continue;
+            }
+            let _ = self.push(e);
```
datanel: you can `self.push(e).unwrap();` as you know that `e.id()` does not exist

src/merge_with.rs
```diff
@@ -0,0 +1,116 @@
+use crate::collection::{CollectionWithId, Id};
+
+pub trait MergeWith<T> {
```
datanel: This trait is useless if it only used by CollectionWithId
you can directly add your implementation in `collection.rs`

woshilapin: I was thinking first of `Collection` too as a possible implementation... and then I realized that inserting in `Collection` wouldn't create conflicts since it's essentially a `Vec`. So I'll move it as a simple method and dump the trait.

woshilapin:
I did remove the trait and move the method into the `CollectionWithId`'s. I switched this PR from draft to a real PR.

src/collection.rs
```diff
@@ -650,6 +650,69 @@ impl<T: Id<T>> CollectionWithId<T> {
...
+        F: FnMut(&mut T, &T),
+        I: IntoIterator<Item = T>,
+    {
+        for e in iterator {
+            if let Some(mut source) = self.get_mut(e.id()) {
+                use std::ops::DerefMut;
+                f(source.deref_mut(), &e);
+                continue;
+            }
+            let _ = self.push(e).unwrap();
```
datanel: no need to have `let _ = `

woshilapin: Oops, indeed. I'll take a look.

woshilapin: Done.

woshilapin:
The Pull Request has been rebased.

PR #351, opened 2019-09-05
moved collection, relations to its crate

datanel:
move `collection.rs`, `relations.rs` into its own crate.

The idea is to have in the future a new repository that will contain 3 worskspaces (relations, collection and transit_model_procmacro renamed) in order to be used by anyone that would like to manage relations between objects.


collection/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "collection"
```
woshilapin: I was wondering if we should call it `collection_with_id`. I know there is also the `Collection` structure but from what I understood, it's mostly an artifact used for `CollectionWithId` (even if we ended using both `Collection` and `CollectionWithId`).

collection/src/collection.rs
```diff
@@ -83,7 +88,7 @@ pub struct Collection<T> {
 /// # Examples
 ///
 /// ```
-/// # use transit_model::collection::*;
```
woshilapin: Technically, these tests should be considered like integration tests, as a user of the final packaged crate. Especially, it would be awkward to have a documentation that tells me
```
use crate::collection::*;
```
I know all of the modifications you made concerns invisible documentation but yet I feel that keeping `transit_model` is more correct.

collection/src/lib.rs
```diff
@@ -0,0 +1,28 @@
...
+//! Collections of objects with typed indices and buildin identifier
+//! support.
+
+/// The error type used by the crate.
+pub type Error = failure::Error;
+
+/// The corresponding result type used by the crate.
+pub type Result<T> = std::result::Result<T, Error>;
+
+pub mod collection;
```
woshilapin: Because of the line below, I think you don't have to make this one `pub`lic.
```suggestion
mod collection;
```

relations/src/lib.rs
```diff
@@ -0,0 +1,27 @@
...
+
+//! Modeling the relations between objects.
+
+/// The error type used by the crate.
+pub type Error = failure::Error;
+
+/// The corresponding result type used by the crate.
+pub type Result<T> = std::result::Result<T, Error>;
+
+pub mod relations;
```
woshilapin: Because of the line below, I think you don't have to make this one `pub`lic.
```suggestion
mod relations;
```

relations/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "relations"
+description = "Manage collection of objects"
```
woshilapin: Probably need another description than the package `collection`.

src/lib.rs
```diff
@@ -65,3 +64,5 @@ pub type Error = failure::Error;
 pub type Result<T> = std::result::Result<T, Error>;
 
 pub use crate::model::Model;
+
+pub use ::collection;
```
woshilapin: Do we need to expose that as `pub`?

collection/Cargo.toml
datanel: It doesn't matter just for now. The important is the name of the next repository. 
`collection` is already used in crates.io.
I would prefer something related to relations or graph. but let discuss that later

collection/src/collection.rs
datanel: I don't understand your point. This PR is to make collection indenpendant of transit_model

relations/Cargo.toml
datanel: yes it was a copy of collection's cargo.toml

src/lib.rs
datanel: probably no. I left like before. 
We could make it private once the new crate `collection` or whatever is published

collection/src/collection.rs
woshilapin: First of all, I misread something and indeed, I didn't meant `transit_model` but `collection`. So basically, having `use collection::collection::*` instead of `use crate::collection::*`.

Another way of seeing this is: documentation is supposed to contain code examples you should be able to copy/paste and build. If you put something like `crate::*` in the documentation, then whoever is gonna copy/paste that is gonna need to figure out what `crate` means in her/his context.

Code example of the documentation should be written as user of the final package. A user of the final package doesn't have access to `crate` but to `collection`.

src/lib.rs
woshilapin: Oh, right, I forgot collection was exposed so better keep that way for now.

collection/src/collection.rs
woshilapin: Maybe the `crate::Result` too could be transformed into `collection::Result`.

collection/src/collection.rs
```diff
@@ -447,9 +452,9 @@ impl<T: Id<T>> CollectionWithId<T> {
     /// # Examples
     ///
     /// ```
-    /// # use transit_model::collection::*;
+    /// # use collection::*;
     /// # use std::collections::HashMap;
-    /// # fn run() -> transit_model::Result<()> {
+    /// # fn run() -> Result<()> {
```
woshilapin: This one probably need `collection::Result` instead of just `Result`.

woshilapin: Mmmm, not true what I just said. The above line being `use collection::*;` so `Resultè is already in scope.

PR #358, opened 2019-09-10
Moved fixtures to tests folder

datanel:
fixtures are just for tests so I think its place is in `tests` folder


woshilapin:
Looks like a good idea. Just need a `cargo fmt` and we should be good to go.

datanel:
@woshilapin fmt done

PR #361, opened 2019-09-10
Overrides PickUp and DropOff

ArnaudOggy:
PickUp and DropOff can be overrided by a VehicleJourneyTimingLink section.


src/transxchange/read.rs
```diff
@@ -681,10 +716,15 @@ fn load_routes_vehicle_journeys_calendars(
             .try_only_child("DepartureTime")?
             .text()
             .parse());
+        let vehicle_journey_timing_links = vehicle_journey
+            .children()
+            .filter(|child| child.name() == "VehicleJourneyTimingLink")
+            .collect::<Vec<&Element>>();
```
woshilapin: Tip: You might be able to just do the following.
```suggestion
            .collect::<Vec<_>>();
```
`.coellct()` mostly need to know in which type of Collection object it needs to gather the items but from the iterator, it already knows what is the items' type.

src/transxchange/read.rs
```diff
@@ -1680,7 +1721,7 @@ mod tests {
                 </child>
             </root>"#;
             let root: Element = xml.parse().unwrap();
-            calculate_stop_times(&stop_points, &root, Time::new(0, 0, 0)).unwrap();
+            calculate_stop_times(&stop_points, &root, Time::new(0, 0, 0), &Vec::new()).unwrap();
```
woshilapin: It'd be nice to have a test when the `Vec` is not empty (when there is an actual override of the `JourneyPatternTimingLink`).

ArnaudOggy: That's why it's in "do not merge"  :wink:

src/transxchange/read.rs
```diff
@@ -549,7 +570,14 @@ fn calculate_stop_times(
...
         let run_time = get_duration_from(&journey_pattern_timing_link, "RunTime");
-        let (pickup_type, drop_off_type) = get_pickup_and_dropoff_types(&stop_point, "Activity");
+        let stop_point_activity = get_stop_point_activity(
+            &journey_pattern_timing_link,
+            &vehicle_journey_timing_links,
+            &stop_point,
+            "From",
+        );
+        let (pickup_type, drop_off_type) =
+            get_pickup_and_dropoff_types(&stop_point_activity, "Activity");
```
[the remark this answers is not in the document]
datanel: seen with @ArnaudOggy it will be called with `WaitTime`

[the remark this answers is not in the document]
woshilapin: Indeed, I don't think so. Because `Activity` is very specific to the pickup and dropoff concepts. I guess this parameter could be removed and hardcoded in the function `get_pickup_and_dropoff_types()`.

PR #362, opened 2019-09-11
Specs TXC: add a TODO list for a later version

papailio:


ArnaudOggy:
Wait for my return before merge it . Some special cases are missing.

PR #370, opened 2019-09-19
Netex IDF - Load networks and companies 

ArnaudOggy:


src/netex_idf/lines.rs
```diff
@@ -0,0 +1,169 @@
...
+    objects::{Company, Network},
+    Result,
+};
+use failure::{format_err, ResultExt};
+use log::info;
+use minidom::Element;
+use std::{fs::File, io::Read};
+use transit_model_collection::CollectionWithId;
+
+fn load_networks_companies(
```
datanel: you could split later in 2 functions `load_network` and `load_company`

src/netex_idf/lines.rs
```diff
@@ -0,0 +1,169 @@
...
+) -> Result<(CollectionWithId<Network>, CollectionWithId<Company>)> {
+    let mut networks = CollectionWithId::default();
+    let mut companies = CollectionWithId::default();
+    for frame in elem
+        .try_only_child("dataObjects")?
+        .try_only_child("CompositeFrame")?
+        .try_only_child("frames")?
+        .children()
+    {
+        if let Some(network) = frame.try_only_child("Network").ok() {
```
datanel:
```rust
        if let Ok(network) = frame.try_only_child("Network") {
```

src/netex_idf/lines.rs
```diff
@@ -0,0 +1,169 @@
...
+            let timezone = Some(String::from(EUROPE_PARIS_TIMEZONE));
+            let network = Network {
+                id,
+                name,
+                timezone,
+                ..Default::default()
+            };
+            networks.push(network)?;
+        }
+        if let Some(company) = frame
```
datanel: same here

src/netex_idf/lines.rs
ArnaudOggy: Ok. We'll see when we get the complete (and heavier) data set. For the moment we are staying on a loop for the 2

PR #374, opened 2019-09-23
Netex stops: split code in some functions

datanel:
just a refactoring

needs https://github.com/CanalTP/transit_model/pull/373

### DIFF
--- a/src/netex_idf/stops.rs
+++ b/src/netex_idf/stops.rs
@@ -28,23 +28,40 @@ use proj::Proj;
 use std::{collections::HashMap, fs::File, io::Read};
 use transit_model_collection::CollectionWithId;
 
-fn load_stops(elem: &Element) -> Result<(CollectionWithId<StopArea>, CollectionWithId<StopPoint>)> {
+fn load_stop_area(stop_place_elem: &Element, id: String) -> Result<StopArea> {
+    // LDA or ZDL without ParentSiteRef
+    let mut stop_area = StopArea {
+        id,
+        name: stop_place_elem
+            .try_only_child("Name")?
+            .text()
+            .trim()
+            .to_string(),
+        visible: true,
+        coord: Coord { lon: 0., lat: 0. },
+        ..Default::default()
+    };
+
+    // add object properties
+    let type_of_place_ref: String = stop_place_elem
+        .try_only_child("placeTypes")?
+        .try_only_child("TypeOfPlaceRef")?
+        .try_attribute("ref")?;
+    stop_area
+        .properties_mut()
+        .insert(("Netex_StopType".to_string(), type_of_place_ref));
+
+    Ok(stop_area)
+}
+
+// A stop area is a LDA or a ZDE without ParentSiteRef
+fn load_stop_areas<'a>(
+    members: &[&'a Element],
+    map_lda_zde: &mut HashMap<String, String>,
+    map_quay_lda: &mut HashMap<String, String>,
+) -> Result<CollectionWithId<StopArea>> {
     let mut stop_areas = CollectionWithId::default();
-    let mut stop_points = CollectionWithId::default();
-    let member_children: Vec<_> = elem
-        .try_only_child("dataObjects")?
-        .try_only_child("CompositeFrame")?
-        .try_only_child("frames")?
-        .children()
-        .flat_map(|e| e.children())
-        .filter(|e| e.name() == "members")
-        .flat_map(|e| e.children())
-        .collect();
-
-    let mut map_lda_zde: HashMap<String, String> = HashMap::default();
-    let mut map_quay_lda: HashMap<String, String> = HashMap::default();
-
-    for stop_place in member_children.iter().filter(|e| e.name() == "StopPlace") {
+    for stop_place in members.iter().filter(|e| e.name() == "StopPlace") {
         let id = stop_place.try_attribute("id")?;
 
         // ZDL with ParentSiteRef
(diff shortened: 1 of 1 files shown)


src/netex_idf/stops.rs
```diff
@@ -14,65 +14,89 @@
...
 
-    let mut map_lda_zde: HashMap<String, String> = HashMap::default();
-    for member in members {
-        let stop_places = member.children().filter(|e| e.name() == "StopPlace");
+    // add object properties
+    let type_of_place_ref: String = stop_place_elem
+        .try_only_child("placeTypes")?
+        .try_only_child("TypeOfPlaceRef")?
+        .try_attribute("ref")?;
+    stop_area
```
woshilapin: Maybe you can create the `KeyValues` instance first, then create the `StopArea`, the `stop_area` would not need to be `mut`.

datanel: done

src/netex_idf/stops.rs
```diff
@@ -83,8 +107,101 @@ fn load_stop_areas(elem: &Element) -> Result<CollectionWithId<StopArea>> {
...
+    let coords: Vec<&str> = gml_pos.split_whitespace().collect();
+    if coords.len() != 2 {
+        bail!("longitude and latitude not found");
+    }
+
+    Ok((coords[0].parse()?, coords[1].parse()?))
+}
+
+fn load_stop_points<'a>(
+    members: &[&'a Element],
```
woshilapin: This seems to be a list of `Quay` and `StopPlace`. Why not directly send the list of `Quay` here (and send the list of `StopPlace` to the function `load_stop_areas` above)?

src/netex_idf/stops.rs
```diff
@@ -83,8 +110,101 @@ fn load_stop_areas(elem: &Element) -> Result<CollectionWithId<StopArea>> {
             );
         }
     }
+}
 
-    Ok(stop_areas)
+fn load_coords(quay: &Element) -> Result<(f64, f64)> {
```
woshilapin: This function could return a `crate::objects::Coord` instead of a tuple.

woshilapin: Mmm, now that I think about it, I think we did mention `Coord` should be only WGS84... which is not the case for these coordinates. So forget about it.

src/netex_idf/stops.rs
datanel: done

src/netex_idf/stops.rs
```diff
@@ -83,8 +110,109 @@ fn load_stop_areas(elem: &Element) -> Result<CollectionWithId<StopArea>> {
...
+fn load_stops(elem: &Element) -> Result<(CollectionWithId<StopArea>, CollectionWithId<StopPoint>)> {
+    let member_children: Vec<_> = elem
+        .try_only_child("dataObjects")?
+        .try_only_child("CompositeFrame")?
+        .try_only_child("frames")?
+        .children()
+        .flat_map(|e| e.children())
+        .filter(|e| e.name() == "members")
+        .flat_map(|e| e.children())
+        .collect();
```
woshilapin: Does it need to be collected? Maybe you could keep the iterator.
```
let mut members = elem
        .try_only_child("dataObjects")?
        .try_only_child("CompositeFrame")?
        .try_only_child("frames")?
        .children()
        .flat_map(|e| e.children())
        .filter(|e| e.name() == "members")
        .flat_map(|e| e.children())
```
Then you could send an iterator to `load_stop_areas` and `load_stop_points`.
```
let mut members_stops_places = members.clone().filter(|e| e.name() == "StopPlace");
let mut stop_areas = load_stop_areas(members_stop_places, &mut map_lda_zde, &mut map_quay_lda)?;
...
let mut members_quays = members.clone().filter(|e| e.name() == "Quay");
let stop_points = load_stop_points(member_quays, &mut stop_areas, &mut map_quay_lda)?;
```

src/netex_idf/stops.rs
```diff
@@ -14,65 +14,92 @@
...
-        let stop_places = member.children().filter(|e| e.name() == "StopPlace");
+    Ok(StopArea {
+        id,
+        name: stop_place_elem
+            .try_only_child("Name")?
+            .text()
+            .trim()
+            .to_string(),
+        visible: true,
+        coord: Coord { lon: 0., lat: 0. },
```
woshilapin: `Coord` implements `Default` so you could probably remove this line.

src/netex_idf/stops.rs
```diff
@@ -14,65 +14,92 @@
...
+) -> Result<CollectionWithId<StopArea>> {
+    let mut stop_areas = CollectionWithId::default();
+    for stop_place in stop_places {
+        let id = stop_place.try_attribute("id")?;
 
-        // add stop areas
-        for stop_place in stop_places {
-            let id = stop_place.try_attribute("id")?;
+        // ZDL with ParentSiteRef
+        if let Ok(parent_site_ref) = stop_place.try_only_child("ParentSiteRef") {
```
woshilapin:
```suggestion
        if let Ok(parent_site_ref) = stop_place.try_only_child("ParentSiteRef").try_attribute::<String>("ref") {
```

src/netex_idf/stops.rs
```diff
@@ -108,6 +237,53 @@ pub fn from_path(path: &std::path::Path, collections: &mut Collections) -> Resul
...
+        </Location>
+    </Centroid>
+</Quay>"#;
+        let root: Element = xml.parse().unwrap();
+        load_coords(&root).unwrap();
+    }
+
+    #[test]
+    #[should_panic]
+    fn test_load_coords() {
```
woshilapin: I don't understand why this test should panic. Could you find a more explicit name for the function?

src/netex_idf/stops.rs
```diff
@@ -108,6 +237,53 @@ pub fn from_path(path: &std::path::Path, collections: &mut Collections) -> Resul
...
+</Quay>"#;
+        let root: Element = xml.parse().unwrap();
+        let coords = load_coords(&root).unwrap();
+
+        assert_eq!((666944.0, 6856019.0), coords);
+    }
+
     #[test]
     fn test_load_stop_areas() {
         let xml = r#"
```
woshilapin: Considering the size of this string, shouldn't we move this kind of tests to integration tests instead, with a fixture?

src/netex_idf/stops.rs
datanel: I don't see the difference between collecting and not collecting + clone

src/netex_idf/stops.rs
datanel: nothing to understand, just a bad copy paste

src/netex_idf/stops.rs
datanel: yes this will be done when we will more objects parsed

src/netex_idf/stops.rs
woshilapin: An iterator is probably much lighter in memory space than a `Vec`. And I did use the `.clone()` twice but technically, only one is needed (on the second use, you can just do `members.filter(...)`.

datanel: it seems I can't do that.
```
200 |     let members_stops_places = member_children
    |                                --------------- value moved here
...
207 |     let members_quays = member_children.cloned().filter(|e| e.name() == "Quay");
    |                         ^^^^^^^^^^^^^^^ value used here after move
```

woshilapin: I was talking about `.clone()` which would clone the iterator, not `cloned()` that would clone the items inside the iterator.

datanel: ok I thought it was a mistake because clone is not implemented by FlatMap

woshilapin: Hmm, it is implemented from what I can see here
https://doc.rust-lang.org/std/iter/struct.FlatMap.html#impl-Clone

datanel: you are right but `minidom::element::Children` should also  implement `Clone` but it does not.

PR #376, opened 2019-09-24
NetexIDF - Load lines

ArnaudOggy:


src/netex_idf/lines.rs
```diff
@@ -177,64 +337,30 @@ mod tests {
...
-                "TRANSDEV IDF RAMBOUILLET 2",
-                "TRANSDEV IDF RAMBOUILLET 3"
-            ],
-            companies_names
-        );
+        let (_networks, companies, map_line_network) = make_networks_companies(&root).unwrap();
+        let lines_netex_idf = load_netex_lines(&root, &map_line_network, &companies).unwrap();
+        let lines = make_lines(&lines_netex_idf).unwrap();
+        let lines_names: Vec<_> = lines.values().map(|l| &l.name).collect();
+        assert_eq!(vec!["Line 01", "Line 04",], lines_names);
```
woshilapin: Can you invert expected and calculated value to be consistent with the rest of the code.

```suggestion
        assert_eq!(lines_names, vec!["Line 01", "Line 04"]);
```

datanel: By convention and by experience I saw a lot of test framework recommending `assert_eq(expected, compute_value())`

this comes from 
`if  (666 == value)` is better than `if (value === 666)` because if you writes `if (666 = value)` you will have a compiler error

But it's true that `assert_eq(compute_value(), expected)` seems more natural

Personnally I usually use expected on the left.

woshilapin: You're right. I'm also thinking of the coloration for `pretty_assertions` that seems more adapted to what you mention (it would show a diff compared to the expected value). I was mostly mentioning it to homogeneize one the whole code. Do you think we should adopt this new pattern (expected on the left)? I'd be totally fine with it and I can help even convert what already exists to this new convention.

PR #378, opened 2019-09-24
[Netex IDF] compute stop area coords

datanel:
with centroid of related stop points


tests/fixtures/netexidf2ntfs/input/netex/arrets.xml
```diff
@@ -99,6 +99,33 @@
                                 <QuayRef ref="FR:91272:ZDE:50076651:STIF" version="732658"/>
                             </quays>
                         </StopPlace>
+                        <StopPlace dataSourceRef="STIF-REFLEX:Operator:59" version="902482" created="2015-02-20T04:02:10.0Z" changed="2017-12-19T10:12:38.0Z" id="will_be_removed_by_sanitize">
+                            <keyList>
+                                <KeyValue typeOfKey="OBJET_QUALIFIER">
```
woshilapin:
```suggestion
                                <KeyValue typeOfKey="OBJECT_QUALIFIER">
```
But in reality, it doesn't matter what this key is, right? So maybe we can just remove it entirely (and all the other instances above and below) so the fixture can be minimalist.
```suggestion
                                <KeyValue>
```

src/netex_idf/stops.rs
```diff
@@ -128,6 +128,37 @@ fn load_coords(quay: &Element) -> Result<(f64, f64)> {
...
 }
 
+use geo::algorithm::centroid::Centroid;
+use geo_types::MultiPoint;
+
+fn update_stop_area_coords(
+    stop_areas: &mut CollectionWithId<StopArea>,
+    stop_points: &CollectionWithId<StopPoint>,
+) {
+    let mut updated_stop_areas = stop_areas.take();
```
woshilapin: I think another possibility would be to `.collect()` the `Idx<StopArea>` in a `Vec` then iterate over this `Vec` and use `.get_mut` on the collection `stop_areas`. This means allocating space for a list of `Idx<StopArea>` but avoids to destroy the original `CollectionWithId`.

If I understand correctly, `.take()` is basically switching some pointers (which is cheap), but then at the end, you have to build again the entire map of the `CollectionWithId` (by calling `CollectionWithId::new()`). The good point though is that the inner `Vec` is never destroyed nor reallocated.

I'm not sure which one is best.

tests/fixtures/netexidf2ntfs/input/netex/arrets.xml
datanel: I just copied data from a real Netex file. I'm not sure if we have to remove xml nodes not used.
I prefer to have a minimal fixture but with a complete structure for integration tests
What do you think @ArnaudOggy ?

ArnaudOggy: @datanel 
Good question. I have no opinion really decided but I think like you; a minimal but complete fixture.

src/netex_idf/stops.rs
datanel: it would work but I think with `take` there are less operations allocation
I don't allocation memory with `collect` and I don't iteration on Idxs to find a stop area.

PR #379, opened 2019-09-25
[Netex IDF] add stop equipments

datanel:


src/netex_idf/stops.rs
```diff
@@ -236,15 +282,15 @@ fn load_stops(elem: &Element) -> Result<(CollectionWithId<StopArea>, CollectionW
...
+    let (stop_points, equipments) = load_stop_points(
         member_children.iter().filter(|e| e.name() == "Quay"),
         &mut stop_areas,
         &mut map_quay_lda,
     )?;
 
     update_stop_area_coords(&mut stop_areas, &stop_points);
 
-    Ok((stop_areas, stop_points))
+    Ok((stop_areas, stop_points, equipments))
```
datanel: should I mutate collections instead of returning N CollectionWithId?

src/netex_idf/stops.rs
```diff
@@ -159,13 +160,46 @@ fn update_stop_area_coords(
...
 fn load_stop_points<'a>(
     quays: impl Iterator<Item = &'a &'a Element>,
     stop_areas: &mut CollectionWithId<StopArea>,
     map_quay_lda: &mut HashMap<String, String>,
-) -> Result<CollectionWithId<StopPoint>> {
+) -> Result<(CollectionWithId<StopPoint>, CollectionWithId<Equipment>)> {
     let mut stop_points = CollectionWithId::default();
-    // add stop points
+    let mut equipments: HashMap<Availability, Equipment> = HashMap::new();
+    let mut id_incr: u8 = 0;
```
woshilapin: Or like this?
```suggestion
    let mut id_incr = 0u8;
```

src/netex_idf/stops.rs
```diff
@@ -207,13 +241,25 @@ fn load_stop_points<'a>(
...
 
+        let associated_equipment = get_or_create_equipment(quay, &mut equipments, &mut id_incr)?;
+        stop_point.equipment_id = Some(associated_equipment.id.clone());
+
         stop_points.push(stop_point)?;
     }
 
-    Ok(stop_points)
+    let mut equipments: Vec<_> = equipments.into_iter().map(|(_, e)| e).collect();
+    equipments.sort_unstable_by(|tp1, tp2| tp1.id.cmp(&tp2.id));
```
woshilapin: I don't understand, why do we need to sort?

src/netex_idf/stops.rs
woshilapin: I'm always struggling with this. I'm much more in favor in stateless functions in general but in most of our connectors, it really helps the readibility and make the code simpler to just send the `Collections` as a mutable reference.

src/netex_idf/stops.rs
datanel: for tests

src/netex_idf/stops.rs
datanel: ok we keep like that

PR #381, opened 2019-09-26
Add documentation about installing the binaries

woshilapin:


woshilapin:
Oops, that should be done on `tartare-tools` :sob: I'm closing it and I'll move it on `tartare-tools`.

PR #387, opened 2019-09-27
 Implement IntoIterator for FileHandler

woshilapin:
I added `IntoIterator` as a constraint for the trait `FileHandler`. This means implementing `IntoIterator` for both `PathFileHandler` and `ZipHandler`. It will allow us to list the files from the Handler and should help in the case of Netex IDFM.

I also added more genericity for the `get_file` and `get_file_if_exists` signature by accepting `AsRef<Path>` instead of simply `&str`.

Last note, this PR will break a use case we were supporting before. When a ZIP file was containing a sub-folder, the `ZipHandler` was able to go down a sub-folder to find the files... but with some drawbacks:
- if a GTFS has its files in different sub-folders, it would work... but not sure we would want to support such a case
- if the ZIP archive contains twice the same filename (for example, `communs.txt`), it won't work... this would cause problems for our Netex IDFM connector


woshilapin:
Some discussions have been launched about the possibility of having an `Iterator` that would emit files and folders (instead of `PathBuf` like in the current version of the code). The problem is with the `ZipHandler` implementation: it is not possible to have an `Iterator` on multiple `ZipFile` because mutability is needed on a `ZipArchive` (and therefore on the inner `ZipFile`). [`StreamIterator`] might be able to solve these kind of problems but they are not part of the standard library. I'm closing this PR.

[`StreamIterator`]: http://cglab.ca/~abeinges/talks/iter/#0

PR #388, opened 2019-09-30
Create netex_utils module with 'get_value_in_keylist'

woshilapin:
Starting to move some `utils` out of hellogo fares in a separate module so we can use them in Netex IDF. This PR is only a start (to keep the PR small), I'm also working on getting more in this `utils` module (like stuff about the frames).

### DIFF
--- a/src/hellogo_fares/read.rs
+++ b/src/hellogo_fares/read.rs
@@ -14,11 +14,11 @@
 // along with this program.  If not, see
 // <http://www.gnu.org/licenses/>.
 
-use super::utils;
-use super::utils::FrameType;
+use super::{utils, utils::FrameType};
 use crate::{
     minidom_utils::{TryAttribute, TryOnlyChild},
     model::Collections,
+    netex_utils,
     objects::*,
     AddPrefix, Result,
 };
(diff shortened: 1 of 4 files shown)


src/hellogo_fares/read.rs
```diff
@@ -169,7 +169,7 @@ fn get_line_id(fare_frame: &Element, service_frame: &Element) -> Result<String>
                     .map(|id| id == line_ref)
                     .unwrap_or(false)
             })
-            .map(|line| utils::get_value_in_keylist(line, "KV1PlanningLijnNummer"))
+            .map(|line| netex_utils::get_value_in_keylist(line, "KV1PlanningLijnNummer"))
```
ArnaudOggy: It's really too strange to have Netex and KV1 mixed up. 
It should better with a generic name (like TransModel or something else).

woshilapin: HelloGo Fares is Netex. KV1 is for the Public Transport data but the fares in HelloGo is using Netex format.

woshilapin:
@datanel The one function in a module is only temporary. I'm already working on moving some other stuff in it (as mentioned in my PR description).

src/hellogo_fares/read.rs
ArnaudOggy: Okay. My memory is lacking on this Monday morning :laughing:

PR #389, opened 2019-09-30
Add mergify config

datanel:
Automatically merge PR done by @CanalTP/coreteam that have

- at least 2 approved reviews
- no comment or request for change reviews
- all requested reviewer have approved
- all CI tests are ok
- no label wip, dot_not_merge
- PR are rebased on master if needed and tests are rerun before merge.
- Reviews are dismissed when a push occurs.

see https://doc.mergify.io/index.html for documentation

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -1,0 +1,25 @@
+pull_request_rules:
+  - name: automatic merge when CI passes and 2 reviews
+    conditions:
+      - "#approved-reviews-by>=2"
+      - "#review-requested=0"
+      - "#changes-requested-reviews-by=0"
+      - "#commented-reviews-by=0"
+      - status-success=continuous-integration/travis-ci/pr
+      - base=master
+      - label!=dot_not_merge
+      - label!=wip
+      - author=@CoreToolsTeam
+    actions:
+      merge:
+        method: merge
+        strict: true
+  - name: delete head branch after merge
+    conditions: []
+    actions:
+      delete_head_branch: {}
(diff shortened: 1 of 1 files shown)


.mergify.yml
```diff
@@ -0,0 +1,25 @@
...
+      - "#changes-requested-reviews-by=0"
+      - "#commented-reviews-by=0"
+      - status-success=continuous-integration/travis-ci/pr
+      - base=master
+      - label!=dot_not_merge
+      - label!=wip
+      - author=@CoreToolsTeam
+    actions:
+      merge:
+        method: merge
```
woshilapin: You mention in your PR's description that PR will be rebased on `master` but here you have `merge`. I guess this should be `rebase` instead but we might also need to define `rebase_fallback` too (probably to `null`?). But `rebase` will have a problem. It seems we can configure to rebase when merging the PR, in this case, this means the CI won't have built the rebased PR... or we tell Mergify to actually rebase the branch and run the CI (with `strict_method`), in this case, the branch will be changed (all SHAs too) so it means that the owner of the branch might have to force-push to update the PR if needed...

datanel: The PR is not rebased to master. It is merged (without rebase)
What I understand is that you have to rebase yourself your PR before it is merged in case there is a conflict.

I suggest to leave the configuration like that and change it if it does not work well.

woshilapin: I agree with that and we'll improve later if needed.

PR #391, opened 2019-10-01
Fix team's name for mergify and remove commented constraint

woshilapin:


.mergify.yml
```diff
@@ -4,12 +4,11 @@ pull_request_rules:
       - "#approved-reviews-by>=2"
       - "#review-requested=0"
       - "#changes-requested-reviews-by=0"
-      - "#commented-reviews-by=0"
```
woshilapin: @kinnou02 we have some problems with this parameter: it seems that as soon as there is at least one comment), this constraint is false.

datanel: this happens when you approve first and then resolve the comments

PR #392, opened 2019-10-01
Refactor of FrameType

woshilapin:
This PR is relatively big so we could do a live peer review so I can guide some of you to review it. Below is some details about what is done.

In HelloGo Fares, we had some utilities to parse Netex frames (`ServiceFrame`, `ResourceFrame`, `FareFrame`), classify them and use them. We have the same need in Netex IDF: this PR is mostly moving these utilities inside `netex_utils/mod.rs` and use these exposed methods in `netex_idf` parsing... with some adjustments :wink:

### DIFF
--- a/src/netex_idf/lines.rs
+++ b/src/netex_idf/lines.rs
@@ -19,6 +19,8 @@ use super::EUROPE_PARIS_TIMEZONE;
 use crate::{
     minidom_utils::{TryAttribute, TryOnlyChild},
     model::Collections,
+    netex_utils,
+    netex_utils::FrameType,
     objects::{CommercialMode, Company, Line, Network, PhysicalMode},
     Result,
 };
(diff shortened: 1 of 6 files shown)


src/netex_utils/mod.rs
```diff
@@ -32,6 +97,83 @@ where
...
+            parse_frames_by_type(&root).unwrap();
+        }
+    }
+
+    mod get_only_frame {
+        use super::*;
+        use pretty_assertions::assert_eq;
+
+        #[test]
+        fn one_unit_price_frame() {
```
ArnaudOggy: The notion of "price" here is no longer valid in your tests (on get_only_frame)

src/netex_idf/lines.rs
```diff
@@ -273,20 +272,30 @@ mod tests {
...
+            vec![&service_frame_networks_1, &service_frame_networks_2],
+        );
+        frames.insert(
+            FrameType::Resource,
+            vec![
+                &resource_frame_organisations_1,
+                &resource_frame_organisations_2,
+            ],
+        );
+        let (networks, companies, _) = make_networks_companies(&frames).unwrap();
```
ArnaudOggy: Unit or functional test here? that is the question to be discussed together.

src/netex_utils/mod.rs
woshilapin: Indeed, I forgot to change the name of the method. Thanks. Done in e99ea23f84f10c886ec18969ed9ad85b7965a1b9.

src/netex_idf/lines.rs
ArnaudOggy: Ok for me at the moment. Anyway these tests will change with the new Netex flows.

src/netex_idf/stops.rs
```diff
@@ -254,17 +256,16 @@ fn load_stop_points<'a>(
 }
 
 fn load_stops(
-    elem: &Element,
+    frames: &HashMap<FrameType, Vec<&Element>>,
```
datanel: you might create a new type for `HashMap<FrameType, Vec<&Element>>` 
it will be used a lot

woshilapin: Good idea, I'll take a look for another PR.

woshilapin:
I messed up and destroyed my branch (thinking it was merged). Thanks to Github, I restored it...

PR #396, opened 2019-10-03
Read, Write, Merge and Sanitize frequencies in NTFS

woshilapin:


tests/fixtures/ntfs/trips.txt
```diff
@@ -1,4 +1,6 @@
 route_id,service_id,trip_id,company_id,physical_mode_id,dataset_id
+M1F,Week,M1FF,TGC,Metro,TGDS
+M1B,Week,M1BF,TGC,Metro,TGDS
```
datanel: you could add frequencies on existing trips instead of creating others

woshilapin: OK, I changed it to modify the existing trips. I also changed the Stop Times since a best practice in the case of `frequencies.txt` is to start from midnight.

woshilapin:
I changed the name of the field `trip_id` into `vehicle_journey_id` in the object `Frequency` (see 3b8aaeeb99dc3338432d721f3229bfc886bc87b7).

PR #398, opened 2019-10-04
update GTFS parser according to the specifications

patochectp:


src/gtfs/read.rs
```diff
@@ -553,48 +670,112 @@ where
...
+    Ok((stopareas, stoppoints, stoplocations))
+}
+
+pub fn read_pathways<H>(file_handler: &mut H) -> Result<CollectionWithId<objects::Pathway>>
+where
+    for<'a> &'a mut H: FileHandler,
+{
+    let file = "pathways.txt";
+    let gtfs_stops = read_stops(file_handler)?;
+    let mut stops: HashMap<String, Stop> = HashMap::new();
```
woshilapin:
```suggestion
    let mut stops = HashMap::new();
```

src/gtfs/read.rs
```diff
@@ -553,48 +670,112 @@ where
...
+        }
+        Some(reader) => {
+            info!("Reading {}", file);
+            let mut rdr = csv::Reader::from_reader(reader);
+            let mut pathways = vec![];
+            for pathway in rdr.deserialize() {
+                let mut pathway: Pathway = skip_fail!(pathway.map_err(|e| format_err!("{}", e)));
+
+                let from_stop_point =
+                    skip_fail!(stops.get(&pathway.from_stop_id).ok_or_else(|| format_err!(
```
woshilapin: Should we do this kind of check on the read? The `sanitize` should take care of that eventually so I feel like we could just read the file `pathways.txt`, not caring about the Stop ID references (isn't it what we do already with other files?). On top of that, it would allow you to use `read_utils::read_opt_collections()` instead of creating this custom `read_pathways` function.

src/gtfs/write.rs
```diff
@@ -143,6 +143,38 @@ fn ntfs_stop_area_to_gtfs_stop(
...
+        .and_then(|eq_id| equipments.get(&eq_id))
+        .map(|eq| eq.wheelchair_boarding)
+        .unwrap_or_else(Availability::default);
+
+    let (lon, lat) = sl.coord.into();
+    Stop {
+        id: sl.id.clone(),
+        name: sl.name.clone(),
+        lat: lat,
+        lon: lon,
```
woshilapin: These lines can be written as
```rust
		lat,
		lon,
```

src/ntfs/mod.rs
```diff
@@ -74,6 +76,33 @@ enum StopLocationType {
     BoardingArea,
 }
 
+impl Into<StopType> for StopLocationType {
+    fn into(self) -> StopType {
+        let stop_type = match self {
```
woshilapin: You do not need the variable, just return the `match` block.

src/ntfs/mod.rs
```diff
@@ -74,6 +76,33 @@ enum StopLocationType {
...
+            StopLocationType::PathwayInterconnectionNode => StopType::GenericNode,
+            StopLocationType::BoardingArea => StopType::BoardingArea,
+        };
+        stop_type
+    }
+}
+
+impl From<StopType> for StopLocationType {
+    fn from(stop_type: StopType) -> StopLocationType {
+        let stop_location_type = match stop_type {
```
woshilapin: You do not need the variable, just return the `match` block.

src/ntfs/read.rs
```diff
@@ -31,90 +31,221 @@ use transit_model_collection::*;
 
 impl From<Stop> for StopArea {
     fn from(stop: Stop) -> StopArea {
+        if stop.name.is_empty() {
+            warn!("stop_id: {}: for station stop_name is required", stop.id);
```
woshilapin: Should we `bail!` instead (and/or possibly panic)? Same question below (at least 3 more times).

src/ntfs/read.rs
```diff
@@ -31,90 +31,221 @@ use transit_model_collection::*;
...
-            },
+            coord: coord,
             stop_area_id: stop.parent_station.unwrap(),
             timezone: stop.timezone,
             geometry_id: stop.geometry_id,
             equipment_id: stop.equipment_id,
             fare_zone_id: stop.fare_zone_id,
             zone_id: stop.zone_id,
-            stop_type,
+            stop_type: stop.location_type.into(),
```
woshilapin: This is why I was talking about implementing `From` instead of `Into` because they both can be used. You could write this instead.
```suggestion
            stop_type: StopLocationType::from(stop.location_type),
```
The `::from()` notation has one advantage (in my opinion), it is forcing you to tell which type you want to convert to, helps the readibility of the code.

src/ntfs/read.rs
```diff
@@ -31,90 +31,221 @@ use transit_model_collection::*;
...
+            platform_code: self.platform_code,
+            level_id: self.level_id,
+            ..Default::default()
+        };
+        Ok(stop_point)
+    }
+}
+
+impl StopLocation {
+    fn from_ntfs_stop(stop: Stop) -> Result<StopLocation> {
```
woshilapin: Could probably be a `TryFrom` implementation.

src/ntfs/read.rs
```diff
@@ -550,6 +681,50 @@ pub fn manage_companies_on_vj(collections: &mut Collections) -> Result<()> {
...
+    }
+
+    let mut pathways = vec![];
+    let mut rdr =
+        csv::Reader::from_path(&pathway_path).with_context(ctx_from_path!(pathway_path))?;
+
+    for pathway in rdr.deserialize() {
+        let mut pathway: Pathway = skip_fail!(pathway);
+        let from_stop_point =
+            skip_fail!(stops.get(&pathway.from_stop_id).ok_or_else(|| format_err!(
```
woshilapin: Same remark than in GTFS, why not let the `sanitize` do the job?

src/ntfs/write.rs
```diff
@@ -423,72 +422,37 @@ pub fn write_fares_v1(base_path: &path::Path, collections: &Collections) -> Resu
...
-        let location_type = match st.stop_type {
-            StopType::Point => StopLocationType::StopPoint,
-            StopType::Zone => StopLocationType::GeographicArea,
-        };
+        let location_type: StopLocationType;
+        if st.stop_type == StopType::Zone {
+            location_type = StopLocationType::GeographicArea;
+        } else {
+            location_type = StopLocationType::from(st.stop_type.clone());
+        }
```
woshilapin: You can do something like this.
```rust
        let location_type = if st.stop_type == StopType::Zone {
            StopLocationType::GeographicArea
        } else {
            StopLocationType::from(st.stop_type.clone())
        };
```

src/ntfs/write.rs
```diff
@@ -423,72 +422,37 @@ pub fn write_fares_v1(base_path: &path::Path, collections: &Collections) -> Resu
...
             visible: st.visible,
             name: st.name.clone(),
-            lat: st.coord.lat,
-            lon: st.coord.lon,
+            lat: st.coord.lat.to_string(),
+            lon: st.coord.lon.to_string(),
             fare_zone_id: st.fare_zone_id.clone(),
             zone_id: st.zone_id.clone(),
-            location_type,
+            location_type: location_type,
```
woshilapin:
```suggestion
            location_type,
```

src/objects.rs
```diff
@@ -1048,6 +1090,110 @@ impl GetObjectType for StopPoint {
...
+    pub comment_links: CommentLinksT,
+    pub visible: bool,
+    pub coord: Coord,
+    pub parent_id: Option<String>,
+    pub timezone: Option<String>,
+    pub geometry_id: Option<String>,
+    pub equipment_id: Option<String>,
+    pub level_id: Option<String>,
+    #[serde(skip)]
+    pub stop_type: StopType,
```
woshilapin: I'm not a fan of using `StopType` which is already used for `StopPoint`. If I understand well, `StopPoint` can only have the values `StopType::Point` and `StopType::Zone`. And `StopLocation` can only have the values `StopType::StopEntrance`, `StopType::GenericNode` or `StopType::BoardingArea`? If this is the case, why not creating the following `enum`, independent of `StopType`.
```rust
#[derive(Derivative, Debug, PartialEq, Clone)]
pub enum StopLocationType {
    StopEntrance,
    GenericNode,
    BoardingArea,
}
```

src/objects.rs
```diff
@@ -1048,6 +1090,110 @@ impl GetObjectType for StopPoint {
...
+    ExitGate,
+}
+
+#[derive(Serialize, Deserialize, Debug, PartialEq, Default, Clone)]
+pub struct Pathway {
+    #[serde(rename = "pathway_id")]
+    pub id: String,
+    pub from_stop_id: String,
+    #[serde(skip)]
+    pub from_stop_type: StopType,
```
woshilapin: Why do we need this value in the model. This value is already stored in the model if you navigate to the corresponding `StopPoint` with the `from_stop_id`.

src/objects.rs
```diff
@@ -1048,6 +1090,110 @@ impl GetObjectType for StopPoint {
...
+    #[serde(
+        deserialize_with = "de_from_u8_with_check",
+        serialize_with = "ser_from_bool"
+    )]
+    pub is_bidirectional: bool,
+    #[serde(default, deserialize_with = "de_option_positive_decimal")]
+    pub length: Option<Decimal>,
+    pub traversal_time: Option<u32>,
+    #[serde(default, deserialize_with = "de_option_non_null_integer")]
+    pub stair_count: Option<i16>,
```
woshilapin: Hmm, can we have a negative number of stairs?
```suggestion
    pub stair_count: Option<u16>,
```

src/utils.rs
```diff
@@ -69,6 +69,25 @@ where
     Ok(i != 0)
 }
 
+pub fn de_from_u8_with_check<'de, D>(deserializer: D) -> Result<bool, D::Error>
```
woshilapin: Why creating a new function? The function just above is doing exactly the same. You could improve the function just above?

src/objects.rs
patochectp: A positive stair_count implies that the rider walk up from from_stop_id to to_stop_id. And a negative stair_count implies that the rider walk down from from_stop_id to to_stop_id.

src/objects.rs
patochectp: Location at which the pathway begins. It contains a stop_id that identifies a platform, entrance/exit, generic node or boarding area from the stops.txt file.
I need the location type to sanitize the model.

src/gtfs/mod.rs
```diff
@@ -90,36 +90,61 @@ enum StopLocationType {
...
+
 #[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
 struct Stop {
     #[serde(rename = "stop_id", deserialize_with = "de_without_slashes")]
     id: String,
     #[serde(rename = "stop_code")]
     code: Option<String>,
-    #[serde(rename = "stop_name")]
+    #[serde(rename = "stop_name")] //Conditionally Required
     name: String,
```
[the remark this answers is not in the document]
patochectp: It was a note to myself. I forgot to withdraw it

src/gtfs/read.rs
patochectp: We already do this for the stop times and the sanitize don't log orphan reference.

src/kv1/read.rs
```diff
@@ -407,6 +407,7 @@ where
             timezone: None,
             geometry_id: None,
             equipment_id: None,
+            level_id: None,
```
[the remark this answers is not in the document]
patochectp: I don’t like it, we can discuss it

src/objects.rs
woshilapin: Yes, I understand that but you can look into `collections.stop_points` to get the stop and get the information. But it's always a balance between memory and performance: your proposition is using more memory, mine would need more processing.

src/objects.rs
woshilapin: Sorry for this one, I missed that in the specification.

woshilapin:
There is a function `.try_merge()` implemented in `Collections`. Some modifications are needed to merge `levels.txt` and `pathways.txt`.

src/gtfs/write.rs
```diff
@@ -143,6 +143,38 @@ fn ntfs_stop_area_to_gtfs_stop(
...
+    sl: &objects::StopLocation,
+    comments: &CollectionWithId<objects::Comment>,
+    equipments: &CollectionWithId<objects::Equipment>,
+) -> Stop {
+    let wheelchair = sl
+        .equipment_id
+        .clone()
+        .and_then(|eq_id| equipments.get(&eq_id))
+        .map(|eq| eq.wheelchair_boarding)
+        .unwrap_or_else(Availability::default);
```
datanel:
```suggestion
        .unwrap_or_default();
```

src/model.rs
```diff
@@ -308,6 +311,69 @@ impl Collections {
...
+                        {
+                            stop_area_ids_used.insert(stop_area_id);
+                        }
+                    }
+                }
+                if let Some(level_id) = &sl.level_id {
+                    level_id_used.insert(level_id.clone());
+                }
+                update_comments_used(&mut comments_used, &sl.comment_links, &self.comments);
+                return true;
```
datanel:
```suggestion
                true
```

src/ntfs/read.rs
```diff
@@ -31,90 +31,221 @@ use transit_model_collection::*;
...
+            }
+            if stop.name.is_empty() {
+                return Err(format_err!(
+                    "stop_id: {}: for entrances/exits stop_name is required",
+                    stop.id
+                ));
+            }
+        }
+        if stop.location_type == StopLocationType::PathwayInterconnectionNode {
+            if stop.parent_station.is_none() {
```
datanel:
```rust
if stop.location_type == StopLocationType::PathwayInterconnectionNode && if stop.parent_station.is_none() {
```

src/ntfs/read.rs
```diff
@@ -31,90 +31,221 @@ use transit_model_collection::*;
...
+                return Err(format_err!(
+                    "stop_id: {}: for generic node parent_station is required",
+                    stop.id
+                ));
+            }
+        }
+        if stop.location_type == StopLocationType::BoardingArea {
+            if stop.parent_station.is_none() {
```
datanel: same here

src/objects.rs
datanel: I agree with @woshilapin 
and may be StopLocation is not a good name

src/gtfs/read.rs
woshilapin: I made a PR to create some logging in the `sanitize` function (see https://github.com/CanalTP/transit_model/pull/399).

src/gtfs/read.rs
```diff
@@ -87,44 +87,170 @@ impl From<Stop> for objects::StopArea {
...
+            stop_codes.insert(("gtfs_stop_code".to_string(), c));
+        }
+
+        if self.name.is_empty() {
+            warn!("stop_id: {}: for station stop_name is required", self.id);
+        }
+
+        let coord: Coord = Coord::from((self.lon, self.lat));
+        if coord == Coord::default() {
+            warn!("stop_id: {}: for station coordinates are required", self.id);
```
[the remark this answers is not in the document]
patochectp: I'm waiting for the global specifications on errors

src/gtfs/read.rs
```diff
@@ -87,44 +87,170 @@ impl From<Stop> for objects::StopArea {
...
+            object_properties: KeysValues::default(),
+            comment_links: objects::CommentLinksT::default(),
+            coord: coord,
+            timezone: self.timezone,
+            visible: true,
+            geometry_id: None,
+            level_id: self.level_id,
+            equipment_id: None,
+        };
+        Ok(stop_area)
```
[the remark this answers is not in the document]
patochectp: same as before

src/gtfs/read.rs
```diff
@@ -443,7 +569,6 @@ pub fn read_agency<H>(
 where
     for<'a> &'a mut H: FileHandler,
 {
-    info!("Reading agency.txt");
```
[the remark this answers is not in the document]
patochectp: It is now managed by read_objects function

tests/fixtures/gtfs2ntfs/no_traffic/input/stops.txt
```diff
@@ -1,5 +1,5 @@
-stop_id,stop_name,stop_lat,stop_lon,location_type,parent_station
-stoparea:1,plop,48.844746,2.372987,1,
-stop:31,pouet,48.844746,2.372987,0,stoparea:1
-stop:32,pouet,48.844746,2.372987,0,stoparea:1
-stop:33,pouet,48.844746,2.372987,0,stoparea:1
+stop_id,stop_name,stop_lat,stop_lon,location_type,parent_station,level_id
```
ArnaudOggy: Maybe add files levels.txt in /input and /output. Otherwise gtfs and ntfs are not valid as well (orphan keys level_id).

woshilapin:
I created https://github.com/patochectp/navitia_model/pull/1 to transform the `Into` implementations into `From` implementation.

woshilapin:
I also updated some tests (to check for stops without geolocation) in https://github.com/patochectp/navitia_model/pull/2.

src/utils.rs
woshilapin: I created this PR to solve the problem.
https://github.com/patochectp/navitia_model/pull/3

PR #400, opened 2019-10-07
Add 'source' object_code for 'network'

woshilapin:


tests/fixtures/gtfs2ntfs/minimal/output/companies.txt
```diff
@@ -1,2 +1,2 @@
 company_id,company_name,company_address,company_url,company_mail,company_phone
-default_agency_id,mon agence,,http://kisio.org,,
+mon_agence,mon agence,,http://kisio.org,,
```
datanel: I don't like you change that. We should keep the test of default agency_id

by the way should the defautl `agency_id` be `default_agency_id` or `1` like in the specs https://github.com/CanalTP/transit_model/blob/master/src/documentation/gtfs_to_ntfs_specs.md#reading-agencytxt?

@papailio @prhod

src/gtfs/read.rs
```diff
@@ -55,10 +55,15 @@ fn get_agency_id(route: &Route, networks: &CollectionWithId<objects::Network>) -
 
 impl From<Agency> for objects::Network {
     fn from(agency: Agency) -> objects::Network {
+        let mut codes = KeysValues::default();
+        codes.insert((
+            "source".to_string(),
+            agency.id.as_ref().cloned().unwrap_or("1".to_string()),
```
datanel:
```suggestion
            agency.id.clone().unwrap_or("1".to_string()),
```

tests/fixtures/gtfs2ntfs/minimal/output/companies.txt
papailio: according to the specs, the default value of `agency_id` should be `1`

datanel: ok so we should change the function `default_agency_id` and the fixtures.

src/gtfs/read.rs
woshilapin: If the rule for the `agency_id` is the same than for the `source` `object_code`, then I can just reuse the `id` directly. This line will disappear.

tests/fixtures/gtfs2ntfs/minimal/output/companies.txt
woshilapin: Done in c950dca47798fbbb7c76fb3ca8b00475dae31e4d.

tests/fixtures/gtfs2ntfs/minimal/input/agency.txt
```diff
@@ -1,2 +1,2 @@
-agency_name,agency_url,agency_timezone
-mon agence,http://kisio.org,Europe/Paris
+agency_id,agency_name,agency_url,agency_timezone
+mon_agence,mon agence,http://kisio.org,Europe/Paris
```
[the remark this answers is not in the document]
woshilapin: I don't see how you can define this fixture as **minimal** since it got multiple `calendar_dates`, multiple `stops`, multiple `trips`, etc. This definition of **minimal** doesn't make any sense and seems different for everyone. That said, I moved this file back to its original version since we can still test there is a new entry in `object_codes.txt` (see 046052d24624f0ecbb5091d43359c15d46027e33).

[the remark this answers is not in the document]
datanel: minimal means just the minimum of data for reading a GTFS. no optional files. no extra columns..

PR #401, opened 2019-10-07
Add 'source' object code for 'stop_point'

woshilapin:


src/gtfs/read.rs
```diff
@@ -83,14 +83,15 @@ impl From<Agency> for objects::Company {
 
 impl From<Stop> for objects::StopArea {
     fn from(stop: Stop) -> objects::StopArea {
-        let mut stop_codes: KeysValues = BTreeSet::new();
+        let mut codes = KeysValues::default();
+        codes.insert(("source".to_string(), stop.id.clone()));
         if let Some(c) = stop.code {
-            stop_codes.insert(("gtfs_stop_code".to_string(), c));
+            codes.insert(("gtfs_stop_code".to_string(), c));
```
datanel: Where does `gtfs_stop_code` come from? I don't see it in the specs

woshilapin: This was already here before. I discussed with @papailio and she seemed aware of that `gtfs_stop_code` being added in the `object_code`. Maybe we should just update the specifications (I can do it as part of this PR).
@prhod @papailio

woshilapin:
I fixed the tests for this PR in 157f54c16aadac157488371b0013c49cb66c2eba.

src/gtfs/read.rs
ArnaudOggy: Specs updated in https://github.com/CanalTP/transit_model/pull/403

woshilapin: Thank you @ArnaudOggy

PR #406, opened 2019-10-09
Add clippy check to the CI

woshilapin:


woshilapin:
> I often use clippy but I don't want to force people to use it.
> 
> this change:
> 
>     * makes travis fail if there are clippy warnings
> 
>     * forces developers to use clippy and add instructions in the code to ignore warnings and can be a waste of time
> 
> 
> If everybody is ok @ArnaudOggy , @patochectp with that, I'm ok but we will need to create a `CONTRIBUTING.md` where we will explain which tools a developer needs to use. (fmt, clippy)
> 
> I leave @patoche to approve or reject

I added a `CONTRIBUTING.md` file, based on the one in Serde project. I also added all the sub-crates as part of the same workspace because so far, running `cargo clippy --all` (or `cargo fmt --all` or even `cargo test --all`) was not doing it in the sub-crates. Therefore, I fixed a few more clippy-warning in some of the sub-crates.

woshilapin:
The `--all` option will actually work only when the workspace is well-defined. https://github.com/CanalTP/transit_model/pull/434 is supposed to do that.

PR #407, opened 2019-10-11
Integrate parsing errors into the specification

woshilapin:
This is a first draft to introduce some of the error management into the specification of the GTFS -> NTFS connector. Below are a few more consideration that might be address as part of this PR or as another PR... to be discussed in this PR.
- there is no specification for the parsing of `shapes.txt`
- there is no specification for the parsing of `frequencies.txt`
- there is no specification for the parsing of `calendar.txt`
- there is no specification for the parsing of `calendar_dates.txt`
- the implementation is parsing a field `local_zone_id` in `stop_times.txt`... but this field doesn't exist in [GTFS specification](https://developers.google.com/transit/gtfs/reference)
- the specification for dataset and contributors is not sufficient (it's only mentioned we need a dataset and a contributor but nothing about the data and information they should contain)
- the conversion of `location_type` in `stops.txt` is incorrect/not specified (there is a mapping, we can't just keep the same integer)

### DIFF
--- a/src/documentation/gtfs_to_ntfs_specs.md
+++ b/src/documentation/gtfs_to_ntfs_specs.md
@@ -4,11 +4,18 @@
 This document aims to describe how the [GTFS format](https://developers.google.com/transit/gtfs/reference) is read in the Navitia Transit Model. To improve readability of this document, the specification will describe the transformation of a GTFS feed into a [NTFS feed](https://github.com/CanalTP/ntfs-specification/blob/master/ntfs_fr.md) (which is a bunch of csv files accordingly to the memory Navitia Transit Model).
 
 ## Introduction
+If at any time of the conversion, the GTFS is not conform to the [GTFS
+specification](https://developers.google.com/transit/gtfs/reference), the
+conversion should stop immediately with an error, unless otherwise specified.
+
+At the end of the conversion, a sanitizing operation is started on the final
+model. See [sanitizer.md](sanitizer.md) for more information.
+
 ### Prepending data
 The NTFS format introduce 2 objects to enable the manipulation of several datasets : contributors and datasets. Those two objects are not described here. The construction of NTFS objects IDs requires, for uniqueness purpose, that a unique prefix (specified for each source of data) needs to be included in every object's id.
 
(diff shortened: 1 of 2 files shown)


src/documentation/gtfs_to_ntfs_specs.md
```diff
@@ -46,29 +55,34 @@ A complementary `object_code` is added to each network with the following proper
...
+| object_codes.txt | object_code | Optional | stops.txt | stop_code | This GTFS property is stored as an associated code for this stop. See (2) for complementary properties. |
 | stops.txt | stop_name | Required | stops.txt | stop_name |  |
 | stops.txt | stop_lat | Required | stops.txt | stop_lat |  |
 | stops.txt | stop_lon | Required | stops.txt | stop_lon |  |
-| stops.txt | location_type | Optionnal | stops.txt | location_type |  |
-| stops.txt | parent_station | Optionnal | stops.txt | parent_station | (1) |
-| stops.txt | stop_timezone | Optionnal | stops.txt | stop_timezone |  |
-| comments.txt | comment_value | Optionnal | stops.txt | stop_desc | See (3) for additionnal properties |
-| equipments.txt | wheelchair_boarding | Optionnal | stops.txt | wheelchair_boarding | See (4) for detailed info. |
+| stops.txt | location_type | Optional | stops.txt | location_type | If value is not one of `0`, `1`, `2`, `3` or `4`, then set to `0` |
```
papailio: So what's the ntfs value when the value is `0` in the gtfs?
What about being explicit with the mapping: "The value is set to `0` if the input value is `0` or unspecified, `1` if the input value is `1`, `3` if the input value is `2`, `4` if the input value is `3` and `5` if the input value is `4`.

woshilapin: There is 2 things here:
1. what happen when the value is NOT valid
2. what is the mapping of `location_type` between GTFS -> NTFS

So to answer the question you're asking.

> So what's the ntfs value when the value is `0` in the gtfs?

I'm only addressing the case 1.: 'If the value is not valid, then default to `0`'. Therefore, if the value is `0` (which is valid), I don't prescribe to do anything.

For addressing 2., this is one of the question I'm asking in the PR description.

Now, your proposition would be a nice way to solve both 1. and 2. at the same time so that might be the way to go!

PR #411, opened 2019-10-14
Remove some useless crates

datanel:


src/ntfs/filter.rs
```diff
@@ -21,14 +21,11 @@
 use crate::{Model, Result};
 use failure::bail;
 use std::collections::HashSet;
-use structopt::clap::arg_enum;
 
-arg_enum! {
```
woshilapin: Will this removal cause any problem on `tartare-tools` side, especially `filter-ntfs` which has an argument of type `Action`?

woshilapin: OK, I just took a look into https://github.com/CanalTP/tartare-tools/pull/36 and you solved it. I'm resolving this conversation.

PR #418, opened 2019-10-17
Netex IDF: Load networks, companies, lines and modes

ArnaudOggy:
Additional data management will be done in another PR (like color, private code, comments, accessibility,...)

### DIFF
--- a/src/netex_idf/lines.rs
+++ b/src/netex_idf/lines.rs
@@ -27,7 +27,7 @@ use crate::{
 use failure::{bail, format_err, ResultExt};
 use log::{info, warn};
 use minidom::Element;
-use std::{collections::BTreeSet, collections::HashMap, fs::File, io::Read};
+use std::{collections::BTreeSet, fs::File, io::Read};
 use transit_model_collection::{CollectionWithId, Id};
 
 #[derive(Debug, Default)]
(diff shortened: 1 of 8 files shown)


src/netex_idf/lines.rs
```diff
@@ -56,12 +54,19 @@ fn load_netex_lines(
...
+                let network_id: String = line
+                    .try_only_child("RepresentedByGroupRef")?
+                    .try_attribute("ref")?;
+                let network_id = if let Some(_network) = networks.get(&network_id) {
+                    network_id
                 } else {
-                    warn!("Failed to find network for line {}", id);
+                    warn!("Failed to find network {} for line {}", network_id, id);
                     continue;
                 };
```
woshilapin: I think you can do something simpler like this.
```
if networks.get(&network_id).is_none() {
    warn!("Failed to find network {} for line {}", network_id, id);
    continue;
}
```

src/netex_idf/read.rs
```diff
@@ -42,7 +42,7 @@ where
     collections.feed_infos = feed_infos;
 
     let path = netex_idf_path.as_ref();
-    stops::from_path(&path.join("arrets.xml"), &mut collections)?;
+    //stops::from_path(&path.join("arrets.xml"), &mut collections)?;
```
woshilapin: Why not just removing the line?

src/netex_idf/lines.rs
datanel: same comment

PR #422, opened 2019-10-18
Implement sanitizer for 'grid_*.txt' files

woshilapin:


woshilapin:
@ArnaudOggy Nice catch for the edgy conditions. Here is the commit 966004a6027fcb2c0cf363b106250dff25c5bd5f.

src/documentation/sanitizer.md
```diff
@@ -49,3 +49,8 @@ anymore. This part of the process remove all of these objects:
 - commercial modes which are not referenced
 - trip properties which are not referenced
 - comments which are not referenced
+- grid calendar which refers a line which doesn't exist (through the
```
datanel: I think we should not write abbreviations in documentations `doesn't` -> `does not` for a better readability and homogeneity

woshilapin: `doesn't` is not really an abbreviation, more a contraction and is [acceptable in written English](https://en.wikipedia.org/wiki/English_auxiliaries_and_contractions#Contractions).

That said, you're right about homogeneity (the line just above is using `are not`). So I changed it for `does not`.

PR #427, opened 2019-10-21
netex_idfm_to_ntfs spec - improve schedule part of the spec

prhod:
Last part of spec update

### DIFF
--- a/src/documentation/netex_idfm_to_ntfs.md
+++ b/src/documentation/netex_idfm_to_ntfs.md
@@ -195,81 +195,86 @@ comment_id | Line/noticeAssignments/NoticeAssignment/NoticeRef | This field is p
 
 
 ## Reading of each folder
-In a **offre_** file, 2 **GeneralFrame** are expected in a **CompositeFrame/frames** node:
+In a **offre_*.xml** file, 2 **GeneralFrame** are expected in a **CompositeFrame/frames** node:
 * one with a `TypeOfFrameRef/@ref` containing the string `NETEX_STRUCTURE`
 * one with a `TypeOfFrameRef/@ref` containing the string `NETEX_HORAIRE`
 In the following chapters, the **CompositeFrame/frames/GeneralFrame/members** is ommitted.
 
 ### routes.txt
-`routes` are provided in each **offre_** file in the nodes **Route**. 
+`routes` are provided in each **offre_** file in the nodes **CompositeFrame/frames/GeneralFrame[]/members/Route[]**. The `GeneralFrame` to be used is the one with the `@ref` containing `NETEX_STRUCTURE`.
 
 NTFS field | Netex-IDFM element | Mapping rule/Comment
 --- | --- | ---
-route_id | *Route/@id* | This field is prefixed. 
-route_name | *Route/Name* | 
-direction_type | *Route/DirectionType* | The value of this field is used without transformation.
+route_id | Route/@id | This field is prefixed. 
+line_id | Route/LineRef/@ref | This field is prefixed. 
+route_name | Route/Name | 
+direction_type | Route/DirectionType | The value of this field is used without transformation.
 destination_id |  | The `DirectionRef` of the Route doesn't link to a stop (neither stop_point nor stop_area), thus its value is not used.
-line_id | *Route/LineRef/@ref* | This field is prefixed. 
 
 **ServiceJourneyPattern references**
-All ServiceJourneyPattern of a `route` are stored as complementary `object_codes`. ServiceJourneyPattern nodes are listed in the same parent node as Route nodes.
+
+All ServiceJourneyPattern of a `route` are stored as complementary `object_codes`. ServiceJourneyPattern nodes are listed in the nodes `CompositeFrame/frames/GeneralFrame[]/members/ServiceJourneyPattern[]`. The `GeneralFrame` to be used is the one with the `@ref` containing `NETEX_STRUCTURE`.
 
 NTFS field | Netex-IDFM element | Mapping rule/Comment
 --- | --- | ---
 object_type |  | fixed value `route` 
-object_id | *ServiceJourneyPattern/RouteRef/@ref* | This field is prefixed. 
+object_id | ServiceJourneyPattern/RouteRef/@ref | This field is prefixed. 
 object_system |  | fixed value `Netex_ServiceJourneyPattern`
-object_code | *ServiceJourneyPattern/@id* | The value of this field is used without transformation.
+object_code | ServiceJourneyPattern/@id | The value of this field is used without transformation.
 
 
 ### trips.txt
-`trips` are described in each **offre_** file in the nodes **ServiceJourney**. 
+`trips` are described in each **offre_*.xml** file in the nodes **CompositeFrame/frames/GeneralFrame[]/members/ServiceJourney[]**. 
+The `GeneralFrame` to be used is the one with the `@ref` containing `NETEX_STRUCTURE`.
 
 NTFS field | Netex-IDFM element | Mapping rule/Comment
 --- | --- | ---
-route_id | *ServiceJourney/JourneyPatternRef/@ref* | route_id of the Route containing the JourneyPatternRef as an object_code.
+route_id | ServiceJourney/JourneyPatternRef/@ref | route_id of the Route containing the `JourneyPatternRef` as an object_code.
 service_id |  | Defined using `ServiceJourney/DayTypeRef`, see (calendar.txt and calendar_dates.txt)[]
-trip_id | *ServiceJourney/@id* | This field is prefixed. 
-trip_headsign | *ServiceJourney/DestinationDisplayRef* | Content of the `DestinationDisplay/FrontText` node. If not available, the name of the `stop_point` of the last `stop_time` is used.
-trip_short_name | *ServiceJourney/DestinationDisplayRef* | Content of the `DestinationDisplay/PublicCode` node. If not available, this field is empty.
-company_id | *ServiceJourney/OperatorRef* | if *ServiceJourney/OperatorRef* is not defined, use *Line/OperatorRef* in [*lines.xml*](#linestxt) file. This field is prefixed.
-physical_mode_id | *Line/TransportMode* | see physical_modes definition
+trip_id | ServiceJourney/@id | This field is prefixed. 
+trip_headsign | ServiceJourney/DestinationDisplayRef | Content of the `DestinationDisplay/FrontText` node. If not available, the name of the `stop_point` of the last `stop_time` is used.
+trip_short_name | ServiceJourney/DestinationDisplayRef | Content of the `DestinationDisplay/PublicCode` node. If not available, this field is empty.
+company_id | ServiceJourney/OperatorRef | if `ServiceJourney/OperatorRef` is not defined, use `Line/OperatorRef` in [*lines.xml*](#linestxt) file. This field is prefixed.
+physical_mode_id | Line/TransportMode | see physical_modes definition
 trip_property_id |  | see [trip_properties.txt](#trip_propertiestxt)
 
 
 **comment_links for a trip**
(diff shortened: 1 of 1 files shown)


src/documentation/netex_idfm_to_ntfs.md
```diff
@@ -195,81 +195,86 @@ comment_id | Line/noticeAssignments/NoticeAssignment/NoticeRef | This field is p
 
 
 ## Reading of each folder
-In a **offre_** file, 2 **GeneralFrame** are expected in a **CompositeFrame/frames** node:
+In a **offre_*.xml** file, 2 **GeneralFrame** are expected in a **CompositeFrame/frames** node:
 * one with a `TypeOfFrameRef/@ref` containing the string `NETEX_STRUCTURE`
 * one with a `TypeOfFrameRef/@ref` containing the string `NETEX_HORAIRE`
 In the following chapters, the **CompositeFrame/frames/GeneralFrame/members** is ommitted.
```
woshilapin: First of all,
```suggestion
In the following chapters, the **CompositeFrame/frames/GeneralFrame[]/members** is ommitted.
```
But also, this sentence is not entirely true because only 2 lines below, you're using the full path...

src/documentation/netex_idfm_to_ntfs.md
```diff
@@ -195,81 +195,86 @@ comment_id | Line/noticeAssignments/NoticeAssignment/NoticeRef | This field is p
...
 NTFS field | Netex-IDFM element | Mapping rule/Comment
 --- | --- | ---
-route_id | *Route/@id* | This field is prefixed. 
-route_name | *Route/Name* | 
-direction_type | *Route/DirectionType* | The value of this field is used without transformation.
+route_id | Route/@id | This field is prefixed. 
+line_id | Route/LineRef/@ref | This field is prefixed. 
+route_name | Route/Name | 
+direction_type | Route/DirectionType | The value of this field is used without transformation.
 destination_id |  | The `DirectionRef` of the Route doesn't link to a stop (neither stop_point nor stop_area), thus its value is not used.
```
woshilapin: :smile: :man_facepalming:

src/documentation/netex_idfm_to_ntfs.md
```diff
@@ -195,81 +195,86 @@ comment_id | Line/noticeAssignments/NoticeAssignment/NoticeRef | This field is p
...
-All ServiceJourneyPattern of a `route` are stored as complementary `object_codes`. ServiceJourneyPattern nodes are listed in the same parent node as Route nodes.
+
+All ServiceJourneyPattern of a `route` are stored as complementary `object_codes`. ServiceJourneyPattern nodes are listed in the nodes `CompositeFrame/frames/GeneralFrame[]/members/ServiceJourneyPattern[]`. The `GeneralFrame` to be used is the one with the `@ref` containing `NETEX_STRUCTURE`.
 
 NTFS field | Netex-IDFM element | Mapping rule/Comment
 --- | --- | ---
 object_type |  | fixed value `route` 
-object_id | *ServiceJourneyPattern/RouteRef/@ref* | This field is prefixed. 
+object_id | ServiceJourneyPattern/RouteRef/@ref | This field is prefixed. 
 object_system |  | fixed value `Netex_ServiceJourneyPattern`
```
woshilapin: `Netex_ServiceJourneyPattern` or `NetexIDFM_ServiceJourneyPattern`?

src/documentation/netex_idfm_to_ntfs.md
```diff
@@ -195,81 +195,86 @@ comment_id | Line/noticeAssignments/NoticeAssignment/NoticeRef | This field is p
...
-object_code | *ServiceJourneyPattern/@id* | The value of this field is used without transformation.
+object_code | ServiceJourneyPattern/@id | The value of this field is used without transformation.
 
 
 ### trips.txt
-`trips` are described in each **offre_** file in the nodes **ServiceJourney**. 
+`trips` are described in each **offre_*.xml** file in the nodes **CompositeFrame/frames/GeneralFrame[]/members/ServiceJourney[]**. 
+The `GeneralFrame` to be used is the one with the `@ref` containing `NETEX_STRUCTURE`.
```
woshilapin: I believe this is the other frame.
```suggestion
The `GeneralFrame` to be used is the one with the `@ref` containing `NETEX_HORAIRE`.
```

src/documentation/netex_idfm_to_ntfs.md
```diff
@@ -195,81 +195,86 @@ comment_id | Line/noticeAssignments/NoticeAssignment/NoticeRef | This field is p
...
 
 NTFS field | Netex-IDFM element | Mapping rule/Comment
 --- | --- | ---
-route_id | *ServiceJourney/JourneyPatternRef/@ref* | route_id of the Route containing the JourneyPatternRef as an object_code.
+route_id | ServiceJourney/JourneyPatternRef/@ref | route_id of the Route containing the `JourneyPatternRef` as an object_code.
```
woshilapin: Do we keep the `object_code` only to facilitate the finding of this `route_id` or we do need to keep the `JourneyPatternRef` in `object_code` for other reasons?

src/documentation/netex_idfm_to_ntfs.md
```diff
@@ -195,81 +195,86 @@ comment_id | Line/noticeAssignments/NoticeAssignment/NoticeRef | This field is p
...
 service_id |  | Defined using `ServiceJourney/DayTypeRef`, see (calendar.txt and calendar_dates.txt)[]
-trip_id | *ServiceJourney/@id* | This field is prefixed. 
-trip_headsign | *ServiceJourney/DestinationDisplayRef* | Content of the `DestinationDisplay/FrontText` node. If not available, the name of the `stop_point` of the last `stop_time` is used.
-trip_short_name | *ServiceJourney/DestinationDisplayRef* | Content of the `DestinationDisplay/PublicCode` node. If not available, this field is empty.
-company_id | *ServiceJourney/OperatorRef* | if *ServiceJourney/OperatorRef* is not defined, use *Line/OperatorRef* in [*lines.xml*](#linestxt) file. This field is prefixed.
-physical_mode_id | *Line/TransportMode* | see physical_modes definition
+trip_id | ServiceJourney/@id | This field is prefixed. 
+trip_headsign | ServiceJourney/DestinationDisplayRef | Content of the `DestinationDisplay/FrontText` node. If not available, the name of the `stop_point` of the last `stop_time` is used.
```
woshilapin: In the data we have (which might be incomplete), I can see some of those `DestinationDisplay` node but they are in the `NETEX_STRUCTURE` frame, while we are parsing trips in the `NETEX_HORAIRE` frame. It might worth mentioning this if I'm correct.

On top of that, this line contains an indirection: we shouldn't use `ServiceJourney/DestinationDisplayRef` as a value but as a reference to jump somewhere else in the XML to get the value. There is the same problem below with `trip_short_name` et `company_id`. I don't know if would want to express that notion in a better way?

src/documentation/netex_idfm_to_ntfs.md
```diff
@@ -195,81 +195,86 @@ comment_id | Line/noticeAssignments/NoticeAssignment/NoticeRef | This field is p
...
+trip_short_name | ServiceJourney/DestinationDisplayRef | Content of the `DestinationDisplay/PublicCode` node. If not available, this field is empty.
+company_id | ServiceJourney/OperatorRef | if `ServiceJourney/OperatorRef` is not defined, use `Line/OperatorRef` in [*lines.xml*](#linestxt) file. This field is prefixed.
+physical_mode_id | Line/TransportMode | see physical_modes definition
```
woshilapin: Do we want to mention this field is **not** prefixed?

src/documentation/netex_idfm_to_ntfs.md
```diff
@@ -195,81 +195,86 @@ comment_id | Line/noticeAssignments/NoticeAssignment/NoticeRef | This field is p
...
 
 NTFS field | Netex-IDFM element | Mapping rule/Comment
 --- | --- | ---
-trip_id | *ServiceJourney/@id* | This field is prefixed. 
+trip_id | ServiceJourney/@id | This field is prefixed. 
 stop_sequence | | Auto-incremented field starting with `0` for the first stop_time
 stop_id | | See (1) below
-arrival_time | *TimetabledPassingTime/ArrivalTime* | If `TimetabledPassingTime/DepartureDayOffset` value is >0, arrival_time is incremented 24 hours for each day offset. 
-departure_time | *TimetabledPassingTime/DepartureTime* | If `TimetabledPassingTime/DepartureDayOffset` value is >0, departure_time is incremented 24 hours for each day offset. 
+arrival_time | TimetabledPassingTime/ArrivalTime | If `TimetabledPassingTime/DepartureDayOffset` value is >0, arrival_time is incremented 24 hours for each day offset.
```
woshilapin: Does this mean we can end up with `25:00` (if we have `1:00` in `ArrivalTime` and a `DepartureDayOffset` to `1`)? Even if no modulo operation is mentioned here, we have a serious tendency to automatically do it when it comes to hours... so it might worth making it explicit here.

src/documentation/netex_idfm_to_ntfs.md
prhod: Good question. I hope this code to be consistant in other netex profiles, and there is also a `Netex_PrivateCode` in Line object. Let's keep it like that.

src/documentation/netex_idfm_to_ntfs.md
prhod: :+1:

src/documentation/netex_idfm_to_ntfs.md
prhod: you're right, I changed my writing and forgot to remove this one ! thanks

src/documentation/netex_idfm_to_ntfs.md
prhod: I think having those codes is not very space consuming and can be usefull for debugging / data analyse purposes.

src/documentation/netex_idfm_to_ntfs.md
prhod: Nice catch for the first one, `DestinationDisplay` are in the `ServiceJourneyPattern` object (french spec was OK :-/ ). I've tried a clarification for the double use of ServiceJourneyPattern and ServiceJourney, hope it will be good :)

For the references, i did the best I could and I find the current specs understandable. As I try not to be too verbose, let's keep it like that if it's ok with you.

src/documentation/netex_idfm_to_ntfs.md
prhod: It's stated in the corresponding chapter, but I've added it here and at the line/commercial_mode_id level too as a reminder ;)

woshilapin: I wonder if these details should actually be left only for the file `common.md` (see #438).

src/documentation/netex_idfm_to_ntfs.md
prhod: ok, I've added it

src/documentation/netex_idfm_to_ntfs.md
prhod: the definition of the mapping between netex value and the physical_mode values should be in this file, because each format is different. Maybe I'm missing something ?

woshilapin: I was talking only about the prefix part. Obviously this section has to be there. But in order to know if you need (or not), to prefix something, this is actually described in #438 so I was thinking, maybe we can left out of this specification all details related to prefixes, except for a link to the `common.md`?

prhod: Ok, I think I understand what you mean. There is indeed a few duplicated information,  but in this case (prefixes), it's very important and having it at both locations seems relevant IMHO. What do you think ?

woshilapin: I'm fine with it if you think it's best to emphasize it.

prhod:
I've rebase the PR since auto-merge failed. @datanel and @woshilapin if you can approve again :pray:

PR #430, opened 2019-10-22
Parse route for Netex IDFM

woshilapin:


src/netex_idf/offers.rs
```diff
@@ -60,7 +111,8 @@ pub fn read_offer_folder(offer_folder: &Path, _collections: &mut Collections) ->
             .parse()
             .map_err(|_| format_err!("Failed to open {:?}", offer_path))?;
         info!("Reading {:?}", offer_entry.path());
-        parse_offer(&offer)?;
+        let routes = parse_offer(&offer, collections)?;
```
ArnaudOggy: let routes = skip_fail!(parse_offer(&offer, collections).map_err(|e| format_err!(
    "Skip file {}: {}",
    offer_path.display(),
    e
)));
Netex files will be dirty for a long time. 
It's better not to fail at the first invalid feed.

woshilapin:
@ArnaudOggy I renamed the fixtures in 8c04fe091d3e233f84bdc49f46a663266b7d25da ;-)

src/netex_idf/offers.rs
woshilapin: I now skip the invalid routes and I added some fixtures for it in 62243cb0d655c2de74036b90559ec6f7055503bf.

woshilapin:
@ArnaudOggy I refactored the code to avoid this chain of `skip_fail` as we discussed and making use of `TryFrom` trait instead.

PR #433, opened 2019-10-23
Improve version definition

woshilapin:
Add `version` and `path` as described in [documentation](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-path-dependencies).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,14 +41,12 @@ serde = { version = "1", features = ["derive"] }
 serde_json = "1"
 tempfile = "3"
 time-parse = "0.1"
-transit_model_procmacro = { version = "0.1", path = "./transit_model_procmacro" }
 walkdir = "2.1"
 wkt = "0.5"
 zip = "0.5"
-# transit_model_collection = "0.1"
-transit_model_collection = { path = "./collection" }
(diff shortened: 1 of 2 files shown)


woshilapin:
> if you want to compile tartare-tools with local transit_model, it won't work. you have to delete the path because you will have version conflicts.
> It's easier to comment and uncomment than to delete each time.
> And it seems I had a problem making a release
> I prefer to leave as before personally

So I added the following section in my `Cargo.toml` in `tartare-tools` and I had no problem to compile it.
```toml
[patch.crates-io]
transit_model = { version = "0.8.0", path = "../transit_model" }
transit_model_collection = { path = "../transit_model/collection" }
```

I don't expect `tartare-tools` to compile with a local `transit_model` without changing at least the `Cargo.toml` of `tartare-tools` because they're 2 different projects. They should refer to each other with a version number. Or to be more concrete, I don't expect something like the following lines to be in the `Cargo.toml` of `tartare-tools`.

```toml
[dependencies]
transit_model = { version = "0.9", features = ["proj"], path = "../transit_model"  }
...
```

datanel:
ok I will see if there is any problem with releases

Issue #366, opened 2019-09-12
Use cargo worskpace


datanel:
I propose a new  file structure based on cargo workspaces

```
.
├── Cargo.lock
├── Cargo.toml
├── README.md
├── examples
├── LICENSE
├── libs
│   ├── collection
│   ├── model-builder
│   ├── relations
│   └── transit_model_procmacro
├── src
└── tests
```

Cargo.toml
```toml
[package]
....


[workspace]
members = [
    ".",
    "./libs/collection",
    "./libs/relations",
    "./libs/transit_model_procmacro",
    "./libs/model-builder"
]

[dependencies]
...
transit_model_procmacro = { version = "0.1.0", path = "./libs/transit_model_procmacro" }
transit_model_collection = { path = "./libs/collection" }
transit_model_relations = { path = "./libs/relations" }

...
```
This will allow to:
- keep libraries in a separate folder
- run all tests with `cargo test --all`

datanel:
collection, relations and transit_model_procmacro are now external crates

PR #434, opened 2019-10-23
Improve workspace and version definition

woshilapin:
Define the workspace. Will be useful for CanalTP/transit_model#406 and the `--all` option.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,10 +20,7 @@ matrix:
   allow_failures:
   - rust: nightly
 script:
-- cargo test --verbose --features proj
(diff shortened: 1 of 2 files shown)


woshilapin:
> we can go further #366 but it can be done later

I like that, I didn't remember it. At the same time, it's not really how most of other multi-crates project are doing it as far as I can see. Let's discuss the pros and cons together and push that for another PR.

> you can also remove all the `cd ...` in [/.travis.yml@`master`#L24](https://github.com/CanalTP/transit_model/blob/master/.travis.yml#L24)

I also added `--all` so we still compile and test everything.

PR #435, opened 2019-10-24
Update licenses and add missing licenses

woshilapin:


woshilapin:
> I remember making a PR to remove 2019. slightly_frowning_face
> it was painful to think about changing every year

Honestly, I can go either way, I just wanted to have the same thing everywhere. But indeed you're right. I might just do that.

PR #436, opened 2019-10-24
Add SPDX license identifier

woshilapin:
In relation with https://spdx.org/, this is a proposition.

### DIFF
--- a/benches/apply_rules.rs
+++ b/benches/apply_rules.rs
@@ -1,3 +1,5 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+//
 // Copyright 2017 Kisio Digital and/or its affiliates.
 //
 // This program is free software: you can redistribute it and/or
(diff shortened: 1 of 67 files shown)


woshilapin:
Waiting for some resolution on the problems mentioned in https://github.com/CanalTP/transit_model/pull/437/

pascal-vs:
Why do you keep the following text:

```
// This program is free software: you can redistribute it and/or
// modify it under the terms of the GNU General Public License as
// published by the Free Software Foundation, either version 3 of the
// License, or (at your option) any later version.
//
// This program is distributed in the hope that it will be useful, but
// WITHOUT ANY WARRANTY; without even the implied warranty of
// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
// General Public License for more details.
//
// You should have received a copy of the GNU General Public License
// along with this program.  If not, see
// <http://www.gnu.org/licenses/>.
```
in each file?

woshilapin:
> Why do you keep the following text:
> 
> ```
> // ...
> ```
> 
> in each file?

This SPDX tag is mostly a structured tag that might be used by tools to give a status of the project. It doesn't prevent to have the header's license. But I realize we might never use such tools...

woshilapin:
It has been decided this modification is not really useful, neither will be used afterwards so I'm closing this PR.

PR #437, opened 2019-10-24
Fix licenses

woshilapin:
There is some inconsistencies in the published Licenses we are using in this project. Each file is referring to a `GPL-3.0-or-later` and yet our crates are published on `crates.io` as `AGPL-3.0`. Maybe they are compatible but I'd prefer to be sure. Who could check that?
@stifoon


woshilapin:
I discussed with @stifoon and the project had been planned to be `AGPL-3.0` everywhere. We can change the license everywhere and make it coherent with the approval of all external contributors. I believe @antoine-de and @Tristramg are the only one that might have contribute outside of their work at Kisio. Can you confirm this is the case, and would you be OK to homogeneize everything to a AGPL (header's license in source code files, and main `LICENSE` file in every crate)?

Tristramg:
tl;dr: I speak for antoine and I, we give up any copyright to kisio digital on any contribution to transit_model. So you can set the licence to GPL or AGPL as you prefer.

The longer version:

The change from GPL to AGPL won’t change much for us, but not in a good way.

So, this is a bit embarassing for us. We use transit_model in our validator which is under MIT licence. This means we were violating your licence as the contamination would obligate us to use the same licence.

Would you consider using a less restrictive licence in order to be able to use it as a library?
Otherwise we will switch our validator to GPL or AGPL once you setteled for the licence of your preference.

woshilapin:
> Would you consider using a less restrictive licence in order to be able to use it as a library?

I'm asking the people able to answer this kind of question at Kisio Digital and I (or someone else) will come back to you as soon as possible.

woshilapin:
> Would you consider using a less restrictive licence in order to be able to use it as a library?

We're gonna stick to the `AGPL-3.0-only` license. So please check on your side to verify which kind of license you can use. Thank you.

PR #441, opened 2019-10-30
Add object_code to route in Netex IDFM

woshilapin:
This is needed for the parsing of the trips.

### DIFF
--- a/src/netex_idf/offers.rs
+++ b/src/netex_idf/offers.rs
@@ -184,6 +184,18 @@ fn parse_common(_common: &Element) -> Result<()> {
     Ok(())
 }
 
+fn parse_service_journey_patterns<'a, I>(sjp_elements: I) -> HashMap<String, &'a Element>
+where
+    I: Iterator<Item = &'a Element>,
+{
+    sjp_elements
+        .filter_map(|sjp_element| {
+            let id: String = sjp_element.attribute("id")?;
+            Some((id, sjp_element))
+        })
+        .collect()
+}
+
 fn parse_routes<'a, I>(
     route_elements: I,
     collections: &Collections,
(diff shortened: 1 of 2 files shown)


src/netex_idf/offers.rs
```diff
@@ -218,13 +262,20 @@ fn parse_offer(offer: &Element, collections: &Collections) -> Result<CollectionW
                 NETEX_STRUCTURE
             )
         })?;
+    let map_journeypatterns = structure_frame
```
woshilapin: I know this map could be directly a map:
- key: `route_ref`
- value: `journey_pattern_ref`
But I already know I'm gonna need the map this way for the trips so I'm keeping this intermediate step here and create the new map inside `enhance_with_object_code`.

src/netex_idf/offers.rs
```diff
@@ -180,6 +180,18 @@ fn parse_common(_common: &Element) -> Result<()> {
...
 }
 
+fn parse_service_journey_patterns<'a, I>(sjp_elements: I) -> HashMap<String, &'a Element>
+where
+    I: Iterator<Item = &'a Element>,
+{
+    sjp_elements
+        .filter_map(|sjp_element| {
+            let id: String = sjp_element.attribute("id")?;
+            Some((id, sjp_element))
```
datanel: why you don't return `Some((route_ref, jp_id.clone()))` directly as you do in `enhance_with_object_code`?

woshilapin: Because as I said [here](https://github.com/CanalTP/transit_model/pull/441#discussion_r340664284), I'm going to need more than just the ID for the coming PR, so I'm returning the full `minidom::Element` which contains all the information.

PR #445, opened 2019-11-04
Add default CO2 emissions for physical modes

woshilapin:


src/documentation/common.md
```diff
@@ -0,0 +1,33 @@
+## CO2 emissions and fallback modes
+Physical modes may not contain CO2 emissions. If the value is missing, we are
+using default values (see below), mostly based on what is provided by
+[ADEME](https://www.ademe.fr).
+
+Physical Mode     | CO2 emission
```
prhod: maybe considering adding the unit `gCO2eq` (best writing I found)

src/documentation/common.md
```diff
@@ -0,0 +1,33 @@
...
+Shuttle           |  NC
+SuspendedCableCar |  NC
+Taxi              | 184
+Train             |  11.9
+Tramway           |   4
+
+The following fallback modes are also added to the model (they're usually not
+referenced by any Vehicle Journey).
+
+Physical Mode      | CO2 emission
```
prhod: same here (unit missing)

src/documentation/common.md
woshilapin: I looked a little bit and I also found sometimes a dash `-` to separate the `eq`. I could propose something like **gCO<sub>2</sub>-eq** (I found out that also putting the `2` as an indice might help too).

woshilapin: > I would propose gCO2-eq/km, to make it more explicit that it's grams of CO2 per km?

Thanks to @papailio, here is the final form, done in fc5c4ab224ae5190f3e898423c7e49570d272aa9.

PR #446, opened 2019-11-05
Add calendar to vehicle journey for Netex IDFM

woshilapin:


woshilapin:
There is a problem I didn't saw yesterday but the integration test depends on the date it has been run. CI had run yesterday and was fine. I launched a new build today and it break. I'll take a look.

woshilapin:
> There is a problem I didn't saw yesterday but the integration test depends on the date it has been run. CI had run yesterday and was fine. I launched a new build today and it break. I'll take a look.

I fixed the problem in the commit 59135084d0e3ea8fa5a7eaa89c6655890810403c.

PR #452, opened 2019-11-07
Doc: shared parts among connectors

papailio:
I grouped the common parts in a single file, including what was previously in the sanitizer file.
All feedback on more shared operations that I most probably forgot is welcome.


src/documentation/common.md
```diff
@@ -58,4 +58,75 @@ The files `contributors.txt` and `datasets.txt` provide additional information a
...
+The following rules apply to every converter, unless otherwise explicitly specified.
+- When one or more stop_points in the input data are not attached to a stop_area, 
+a stop_area is automatically created. The coordinates of the new stop_area are 
+computed as the barycenter of the associated stop_points. (The same rule applies 
+in case the stop_area exists in the input data, but the coordinates are not specified.)
+- Unless otherwise specified, dates of service are trasnformed into a list of active dates as if using a single NTFS file `calendar_dates.txt`.
+
+| NTFS file | NTFS field | Constraint | Note |
+| --- | --- | --- | --- |
+| calendar_dates.txt | service_id | Required | All slashes `/` are removed
```
datanel: are you sure all slashes are removed?
I know that slashes are removed in trip, stop time, stop in gtfs connector.

src/documentation/common.md
```diff
@@ -58,4 +58,75 @@ The files `contributors.txt` and `datasets.txt` provide additional information a
...
+| calendar_dates.txt | date | Required | 
+| calendar_dates.txt | exception_type | Required | Fixed value `1`.
+
+## Sanitizer
+The sanitizer checks for incoherences in the model and also cleans up all dangling
```
datanel: the function se call sanitize does not check incoherence. incoherences (2 same id) are check when a collection of objects is created

src/documentation/common.md
```diff
@@ -58,4 +58,75 @@ The files `contributors.txt` and `datasets.txt` provide additional information a
...
 | datasets.txt | contributor_id | contributor_id | Required | This field is prefixed.
 | datasets.txt | dataset_start_date |  |  | Smallest date of all the trips of the dataset.
-| datasets.txt | dataset_end_date |  |  | Greatest date of all the trips of the dataset.
\ No newline at end of file
+| datasets.txt | dataset_end_date |  |  | Greatest date of all the trips of the dataset.
+
+## Common practices
+The following rules apply to every converter, unless otherwise explicitly specified.
+- When one or more stop_points in the input data are not attached to a stop_area, 
+a stop_area is automatically created. The coordinates of the new stop_area are
```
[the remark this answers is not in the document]
woshilapin: I believe this should be considered a bug if it's not done in the other places. Eventually, it should be done for all created NTFS. So I'd be fine to leave like this in the file `common.md` and then fixed it when we see it in the code. What do you think?

src/documentation/common.md
woshilapin: Actually, it's interesting to note that slashes are only removed in the GTFS2NTFS connector. So to be precise, slashes are removed only on `stop.id` and `shape.id` only on the read of a GTFS file. Nothing is done (at least, from what I can see) on the NTFS side. That said, that might be considered a bug... not sure.

src/documentation/common.md
woshilapin: My bad, that's on me, @papailio just copied over the previous text I created in `sanitizer.md`... and you're right, no coherence check is done, only removing objects that are not referenced.

src/documentation/common.md
prhod: the problem has only been encounter with gtfs comming from osm (osm2gtfs tool). IMHO it's a good practice to centralize this kind of "can beak navitia api" data errors.

src/documentation/common.md
prhod: What about changing the name "sanitizer" with "Consistancy" ? this spec is not about where the checks are made, but what controls are effective.

PR #456, opened 2019-11-12
Netex IDF - Validity period calculation

ArnaudOggy:


src/netex_idf/offers.rs
```diff
@@ -234,6 +235,26 @@ where
     Ok(routes)
 }
 
+fn update_validity_period(
```
woshilapin: This function is basically a copy of what is done in the TransXChange connector. I think it would be nice to extract these functions into a common part instead of duplicating code.

src/read_utils.rs
```diff
@@ -102,6 +102,23 @@ pub fn set_dataset_validity_period(
     Ok(())
 }
 
+#[cfg(feature = "proj")]
+pub fn update_validity_period(
```
datanel: I'm not sure this is the good place
a "write" operation in s read module

ArnaudOggy: Will be refactored soon

PR #457, opened 2019-11-14
Exclude all useless files from the Cargo publishing process

woshilapin:
So far, all of our released crates contained a lot of useless files, particularly all the fixtures files. This `package.exclude` should remove all useless files during the publishing process.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,6 +7,21 @@ description = "Transit data management"
 repository = "https://github.com/CanalTP/transit_model"
 keywords = ["ntfs", "gtfs", "netex", "navitia", "transit"]
 edition = "2018"
+exclude = [
+	".gitignore",
+	".mergify.yml",
+	".travis.yml",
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -7,6 +7,22 @@ description = "Transit data management"
 repository = "https://github.com/CanalTP/transit_model"
 keywords = ["ntfs", "gtfs", "netex", "navitia", "transit"]
 edition = "2018"
+exclude = [
+	".gitignore",
+	".mergify.yml",
+	".travis.yml",
+	"Cargo.lock",
```
datanel: this file appears in the package

woshilapin: Indeed, I didn't saw that. I guess I can just remove them all... This PR is gonna be simpler :smile:

woshilapin: I've removed all the occurrences of `Cargo.lock`.

PR #460, opened 2019-11-14
Netex IDF: stop accessibility

datanel:


src/netex_idf/read.rs
```diff
@@ -66,7 +66,6 @@ where
     }
 
     collections.enhance_with_co2();
-    // TODO: uncomment once we have all netex  parsed
-    // collections.sanitize()?;
+    collections.sanitize()?;
```
woshilapin: :sweat_smile:

src/netex_idf/stops.rs
```diff
@@ -129,13 +130,73 @@ fn stop_point_parent_id(
         .map(|stop_area| stop_area.id.clone()))
 }
 
+#[derive(Eq, PartialEq, Hash, Clone, Copy)]
```
woshilapin: In some of the files of `transit_model`, the structures are defined at the very beginning of the file (just after the `use` section). I'm not sure where I stand about that. Do you have any insight on the question to where these should be defined?

src/netex_idf/stops.rs
```diff
@@ -188,10 +249,19 @@ fn load_stop_points<'a>(
...
+        {
+            stop_point.equipment_id = Some(associated_equipment.id.clone());
+        }
+
         stop_points.push(stop_point)?;
     }
 
-    Ok(stop_points)
+    let mut equipments: Vec<_> = equipments.into_iter().map(|(_, e)| e).collect();
+    equipments.sort_unstable_by(|tp1, tp2| tp1.id.cmp(&tp2.id));
```
woshilapin: It's really annoying we have to do that... even if I don't know how to do it differently in this case...

src/netex_idf/stops.rs
datanel: no religion. I just wanted to modify the struct without scrolling.
If @ArnaudOggy or @patochectp whant to put at the beginning. I'll do it

patochectp: I tend to agree with woshilapin

woshilapin: Funny, I don't really have a religion too :wink: and if I really had to choose, I think I would go with @datanel's idea. But I mentioned it because we already had this discussion with @ArnaudOggy :smile:

datanel: done

PR #461, opened 2019-11-14
Netex IDF: line colors

datanel:


src/netex_idf/lines.rs
```diff
@@ -104,6 +130,8 @@ fn make_lines(lines_netex_idf: &CollectionWithId<LineNetexIDF>) -> Result<Collec
             name: ln.name.clone(),
             code: ln.code.clone(),
             network_id: ln.network_id.clone(),
+            color: ln.color.clone().or_else(|| Some(DEFAULT_COLOR)),
```
woshilapin: I'm thinking it's not necessary to clone the entire `Option`, only the value inside which can be done with `.cloned()`. But in reality, I'm not even sure about the difference in term of performance.
```suggestion
            color: ln.color.cloned().or_else(|| Some(DEFAULT_COLOR)),
```

Same remark for the line below.

PR #464, opened 2019-11-15
NETEX IDF: complementary codes on lines

datanel:


src/netex_idf/lines.rs
```diff
@@ -125,6 +125,12 @@ fn make_lines(lines_netex_idf: &CollectionWithId<LineNetexIDF>) -> Result<Collec
             .get(ln.mode.as_str())
             .map(|m| { m.commercial_mode.0.to_string() })
             .ok_or_else(|| format_err!("{} not found", ln.mode)));
+
+        let codes: KeysValues = ln
+            .private_code
+            .clone()
+            .map(|pc| vec![("Netex_PrivateCode".into(), pc)].into_iter().collect())
```
woshilapin: I understand why you do that but isn't the following more easy to read?
```rust
|pc| {
    let mut codes = BTreeSet::new();
    codes.insert(("Netex_PrivateCode".into(), pc));
    codes
}
```

datanel: I did like this before but I finally I copied the code from a guy called Jean
https://github.com/CanalTP/transit_model/blame/master/src/netex_idf/offers.rs#L369
:wink:

woshilapin: lol :-)

PR #465, opened 2019-11-15
Parse communs.xml for Netex IDF

woshilapin:


src/netex_idf/common.rs
```diff
@@ -0,0 +1,89 @@
...
+    let general_frames_by_type = offers::parse_general_frame_by_type(general_frames)?;
+    let common_general_frame = general_frames_by_type
+        .get(&GeneralFrameType::Common)
+        .ok_or_else(|| format_err!("Failed to find the GeneralFrame of type {}", NETEX_COMMON))?;
+    let comments = common_general_frame
+        .only_child("members")
+        .iter()
+        .flat_map(|members| members.children())
+        .filter_map(|notice_element| {
+            println!("BEFORE");
```
datanel: why not but no :)
same for all the println below

woshilapin: Oops...

woshilapin: Fixed, and I also added one missing line which was merging the list of comments into the `Collections`.

PR #467, opened 2019-11-15
[feature] Netex IDF: Add comments for Lines

woshilapin:


src/netex_idf/lines.rs
```diff
@@ -42,7 +42,7 @@ const DEFAULT_TEXT_COLOR: Rgb = Rgb {
     blue: 255,
 };
 
-#[derive(Debug)]
+#[derive(Debug, Default)]
```
ArnaudOggy: Default or not Default, that is the question :laughing:

woshilapin: Oops, I probably pushed part of what I experimented on the side... It should be fixed now.

PR #469, opened 2019-11-15
Netex IDF: add trip properties

datanel:


woshilapin:
@datanel This PR needs a rebase.

src/netex_idf/stops.rs
```diff
@@ -137,49 +129,25 @@ fn stop_point_parent_id(
...
-    })
-}
-
-fn get_or_create_equipment<'a>(
+pub fn get_or_create_equipment<'a>(
     quay: &Element,
-    equipments: &'a mut HashMap<StopAccessibility, Equipment>,
+    equipments: &'a mut HashMap<Accessibility, Equipment>,
     id_incr: &mut u8,
 ) -> Option<&'a mut Equipment> {
```
woshilapin: I don't think you need to return a mutable reference here.
```suggestion
) -> Option<&'a Equipment> {
```

src/netex_idf/offers.rs
```diff
@@ -1089,6 +1091,7 @@ mod tests {
                 "Trip Short Name",
                 vehicle_journey.short_name.as_ref().unwrap()
             );
+            assert_eq!(Some("tp_id".to_string()), vehicle_journey.trip_property_id);
```
woshilapin:
```suggestion
            assert_eq!("tp_id", vehicle_journey.trip_property_id.as_ref().unwrap());
```

src/netex_idf/mod.rs
```diff
@@ -20,6 +20,7 @@ mod lines;
 mod modes;
 mod offers;
 mod read;
+mod share;
```
woshilapin: How about calling it `accessibility` instead of `share`?

src/netex_idf/stops.rs
```diff
@@ -137,49 +129,25 @@ fn stop_point_parent_id(
...
-    let accessibility = accessibility(quay)?;
+    let accessibility_node = quay.only_child("AccessibilityAssessment")?;
+    let accessibility = accessibility(accessibility_node)?;
 
-    let equipment = equipments.entry(accessibility).or_insert_with(|| {
-        let StopAccessibility {
-            wheelchair_boarding,
+    let equipment = equipments.entry(accessibility.clone()).or_insert_with(|| {
+        let Accessibility {
+            wheelchair,
```
woshilapin: You can also destructure with a different name for the variable.
```rust
let Accessibility {
            wheelchair: wheelchair_boarding,
            visual_announcement,
            audible_announcement,
        } = accessibility;
```
This way, you do not need to do the swap of names below on line 150. I know it's highly subjective but I do prefer to name correctly the variable early.

src/netex_idf/lines.rs
```diff
@@ -63,12 +68,43 @@ fn line_color(line: &Element, child_name: &str) -> Option<Rgb> {
...
+) -> Option<&'a mut TripProperty> {
+    let accessibility_node = line.only_child("AccessibilityAssessment")?;
+    let id: String = accessibility_node.attribute("id")?;
+    let accessibility = accessibility(accessibility_node)?;
+
+    let trip_property = trip_propertiess
+        .entry(accessibility.clone())
+        .or_insert_with(|| {
+            let Accessibility {
+                wheelchair,
```
woshilapin: You can destructure and rename the variable.
```suggestion
                wheelchair: wheelchair_accessible,
```

src/netex_idf/stops.rs
```diff
@@ -137,49 +129,25 @@ fn stop_point_parent_id(
...
             visual_announcement,
             audible_announcement,
+            ..
```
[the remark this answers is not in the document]
datanel: it's because I had 4 attributes in a previous version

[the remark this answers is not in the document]
woshilapin: I guessed :smile:

src/netex_idf/mod.rs
datanel: i you want. I was planning to add more things

src/netex_idf/offers.rs
datanel: why do you think it's better?

woshilapin: It's easier to read the expected part which to me is the important part.

datanel: I don't agree but as it's the same thing I will modify it

PR #470, opened 2019-11-19
[Feature] Update fixtures for RoutingConstraintZone

ArnaudOggy:
Add a 4th stopoint to the first ServiceJourneyPattern, 
and a RoutingConstraintZone on the 2nd and 3rd stopoint of this ServiceJourneyPattern


ArnaudOggy:
> Looks fine to me but I do not understand the need to add an additional `StopPoint` to create some `RoutingConstraintZone`. Wasn't it possible to create one from the existing `StopPoint`? That would keep the fixture smaller and therefore easier to maintain.

It is more interesting to have a constraint zone on 2 stops of a course (in the middle) containing 4

PR #472, opened 2019-11-19
[tech] Moved apply_rules functions into modules

datanel:
I just moved some functions into submodules because `appy_rules.rs` was big. This is the same code.

It will allow me to do my next PR on apply_rules

### DIFF
--- a/src/apply_rules/property_rule.rs
+++ b/src/apply_rules/property_rule.rs
@@ -12,30 +12,24 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program. If not, see <https://www.gnu.org/licenses/>
 
-//! See function apply_rules
-
-use crate::model::Collections;
-use crate::utils::{Report, ReportType};
-use crate::Result;
 use crate::{
-    objects::{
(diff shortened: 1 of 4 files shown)


src/apply_rules/property_rule.rs
```diff
@@ -47,8 +41,9 @@ enum ObjectType {
     StopPoint,
     StopArea,
 }
+
 impl ObjectType {
```
woshilapin: This structure is now duplicated, as well as the `as_str()` implementation (I can see it in `property_rules.rs` and in `complementary_code.rs`). Maybe this need to be kept in `apply_rules/mod.rs`?

datanel: I need to add physical modes in property rules.

If I keep only one object type, I add physical mode and I decide to add complementary code on an unknown type, the deserializer will tell me that only stop_point.... physical_mode are possible object types, which is wrong because we can't add codes on physical modes.

so each process (property and complementary codes) is applied on differents object types.

woshilapin: Ah OK, it makes sense now. So there is really 2 different concepts and it was just coincidental that the same `enum` was used before!

PR #473, opened 2019-11-20
[feature] Parse RoutingConstraintZone for Netex IDF

woshilapin:


woshilapin:
This PR is ready to be reviewed. Let's see together if we go with this kind of code or if are moving towards a more `String`-ish solution.

src/netex_idf/offers.rs
```diff
@@ -314,6 +370,73 @@ where
...
+        for routing_constraint_zone in routing_constraint_zone_list {
+            let mut current_node = &mut routing_constraint_zones;
+            current_node.depth = 0;
+            for (depth, stop_point_idx) in routing_constraint_zone.iter().enumerate() {
+                current_node = current_node
+                    .childrens
+                    .entry(*stop_point_idx)
+                    .or_insert_with(TreeNode::default);
+                current_node.depth = depth + 1;
+                if depth == routing_constraint_zone.len() - 1 {
```
patochectp: you can do it like this :
   current_node.is_leaf = depth == routing_constraint_zone.len() - 1;

woshilapin: But not exactly equivalent. Because if `current_node.depth` is already `true` and the `depth == routing_constraint_zone.len()-1` is false, your solution will erase it with `false` but mine will leave `true`.

woshilapin:
I'm closing this PR. An alternative implementation is provided in #478 which is simpler. #478 is slightly less performant but the readibility is much better and the algorithm much simpler to understand. The performance impacts only JourneyPatterns, so the performance impact should be almost negligeable on the global parsing.

PR #455, opened 2019-11-08
Add 'trip_short_name' into 'trips.txt' (NTFS v0.10.0)

woshilapin:
Note: I'm bumping version of `transit_model` to `0.11.0` since this is a breaking change.

PR #474, opened 2019-11-20
[tech] Bump version to catch up on our new strategy (see #458)

woshilapin:
We closed #458 since we want to let each contributor bump the version as they need to. But we need to catch up from the last version `0.11.0` (we've already produced code since and the version has not been bumped since).

Usually, I would have bumped to `0.11.1` but I believe the PR #455 is breaking change (we're changing from NTFS `v0.9.2` to NTFS `v0.10.0`.


woshilapin:
After discussion, #455 is not a breaking change, it's about adding an optional field in the NTFS norm.

PR #475, opened 2019-11-20
[feature] apply property rules on co2_emission

datanel:


datanel:
@woshilapin I don't understand your point.
I added rules in this file. the values are replaced

woshilapin:
> @woshilapin I don't understand your point.
> I added rules in this file. the values are replaced

Hmm, let's put that on the fact I was still sleepy when I did that comment :sleeping:

tests/fixtures/apply_rules/output_consolidation_with_object_code/physical_modes.txt
```diff
@@ -0,0 +1,4 @@
+physical_mode_id,physical_mode_name,co2_emission
+Bus,Bus,130.5
+Metro,Metro,90.0
+RapidTransit,Rapid Transit,
```
ArnaudOggy: I am not sure of the need for this file. There is no use of the property_rules.txt in the test.
The PR is ok otherwise.

tests/fixtures/apply_rules/output_consolidation_unvalid/physical_modes.txt
```diff
@@ -0,0 +1,4 @@
+physical_mode_id,physical_mode_name,co2_emission
```
ArnaudOggy: I am not sure of the need for this file. There is no use of the property_rules.txt in the test.
The PR is ok otherwise.

datanel: done

PR #476, opened 2019-11-20
[tech] Calendar deduplication

patochectp:


src/model.rs
```diff
@@ -777,6 +777,39 @@ impl Collections {
...
+                if c.dates == *dates {
+                    return Some(c);
+                }
+            }
+            None
+        }
+
+        let mut calendars_used: Vec<Calendar> = vec![];
+        let mut vehicle_journeys = self.vehicle_journeys.take();
+        vehicle_journeys.sort_unstable_by_key(|vj| vj.id.clone());
```
woshilapin: I'm not sure I understand why you need to sort the `vehicle_journeys` by `id`? `vehicle_journeys` and `calendars` are both stored in `CollectionWithId` which are already ordered and the order is not gonna be altered (only some elements are gonna be removed, but relative positions are gonna be the same). Or am I missing something?

If I am missing something and the sorting needs to stay, please consider the following solution to avoid the `.clone()` operation.
```suggestion
        vehicle_journeys.sort_unstable_by(|vj_a, vj_b| { 
            vj_a.service_id                              
                .partial_cmp(&vj_b.service_id)           
                .unwrap_or(std::cmp::Ordering::Equal)    
        });                                              
```

src/model.rs
```diff
@@ -777,6 +777,39 @@ impl Collections {
...
+        fn find_duplicate_calendar<'a>(
+            calendars: &'a [Calendar],
+            dates: &BTreeSet<NaiveDate>,
+        ) -> Option<&'a Calendar> {
+            for c in calendars {
+                if c.dates == *dates {
+                    return Some(c);
+                }
+            }
+            None
```
woshilapin: This part could be written with the following syntax too.
```rust
calendars.iter().filter(|c| c.dates == *dates).next()
```
This is shorter and might allow to remove the `find_duplicate_calendar` function and put the code directly inline. If you prefer to keep the `find_duplicate_calendar`, then maybe the `for`-loop is more readable?

tests/fixtures/kv12ntfs/output/calendar.txt
```diff
@@ -1,6 +1,4 @@
 service_id,monday,tuesday,wednesday,thursday,friday,saturday,sunday,start_date,end_date
 prefix:2029:1:1,0,0,0,0,0,0,1,20190428,20190530
-prefix:2029:10:10,0,0,0,0,0,1,0,20190427,20190504
-prefix:2165:6:6,0,0,0,0,1,0,0,20190913,20190913
```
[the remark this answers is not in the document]
patochectp: that is exactly what it is.

src/model.rs
datanel: why not `vehicle_journeys.sort_unstable_by(|vj1, vj2| vj1.service_id.cmp(&vj2.service_id));` ?

src/model.rs
datanel: `calendars.iter().find(|c| c.dates == *dates)`

src/model.rs
woshilapin: Yes, why not? That's obviously better than mine! No reason, just didn't look deep enough I guess.

src/model.rs
woshilapin: Ah yes, once again, you caught it. Maybe one day I'll think about it!

src/model.rs
```diff
@@ -1161,4 +1194,78 @@ mod tests {
...
+
+            let vehicle_journey = collections
+                .vehicle_journeys
+                .get("vehicle_journey_id_3")
+                .unwrap();
+            assert_eq!("service_3", vehicle_journey.service_id);
+
+            let calendar = collections.calendars.get("service_2");
+            assert_eq!(None, calendar);
+        }
```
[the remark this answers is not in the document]
patochectp: I don’t think it’s relevant because I don’t change the calendars

PR #477, opened 2019-11-21
[fix] Fix sanitize after filter-ntfs

datanel:


src/model.rs
```diff
@@ -588,12 +588,13 @@ impl Collections {
             vj_id_to_old_idx.iter().map(|(id, idx)| (idx, id)).collect();
         comments_used.extend(self.stop_time_comments.iter().filter_map(
             |((old_vj_idx, _), old_comment_idx)| {
-                let old_vj_id = vj_idx_to_old_id[old_vj_idx];
```
datanel: that line was failing because in `filter_ntfs` function we remove some vehicle journeys so `old_vj_idx` may not be contained in `vj_idx_to_old_id` anymore

tests/fixtures/filter_ntfs/output_extract/trips.txt
```diff
@@ -1,3 +1,3 @@
 trip_id,route_id,physical_mode_id,dataset_id,service_id,trip_headsign,trip_short_name,block_id,company_id,trip_property_id,geometry_id
```
woshilapin: I still don't understand why you do reorder all the file's lines since the test utility is checking for a `Set` of line anyway (so the order doesn't matter).

datanel: I don't reorder anything, the code do it, but I don't where. in sanitize a lot of things happen

datanel: ok I understand what you are saying. I'll change the fixtures

src/model.rs
datanel: I just executed the binary to see if there are changes

tests/fixtures/filter_ntfs/output_extract/trips.txt
datanel: I just executed the binary to see if there are changes

datanel: fixed now

PR #479, opened 2019-11-25
[fix] Netex - Skip line without network

ArnaudOggy:
Structure Line/RepresentedByGroupRef/@ref can not exist. It is not an obligatory field, and according to specifications the line should be skipped.

### DIFF
--- a/src/netex_idf/lines.rs
+++ b/src/netex_idf/lines.rs
@@ -342,6 +342,47 @@ mod tests {
     }
 
     #[test]
+    fn test_load_netex_lines_with_one_without_network() {
+        let xml = r#"
+            <ServiceFrame>
+               <lines>
+                  <Line id="FR1:Line:C00001">
+                     <Name>Line 01</Name>
+                     <ShortName>01</ShortName>
+                     <TransportMode>bus</TransportMode>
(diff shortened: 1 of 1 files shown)


woshilapin:
I thought we said we wouldn't fix this one? If we decide to fix it, why not also the `OperatorRef` and all the other missing fields?

src/netex_idf/lines.rs
```diff
@@ -116,9 +116,9 @@ fn load_netex_lines(
                     .map(Element::text)
                     .ok();
                 let private_code = line.only_child("PrivateCode").map(Element::text);
-                let network_id: String = line
-                    .try_only_child("RepresentedByGroupRef")?
-                    .try_attribute("ref")?;
+                let network_id: String = skip_fail!(line
```
woshilapin: How about adding a test for that one?

ArnaudOggy: I thought about doing it yesterday. I just wrote that before leaving.

ArnaudOggy:
> Are the other fields optional too? Like `OperatorRef`, `PublicCode`, `ShortName` or `TransportMode`?
OperatorRef, PublicCode and TransportMode are declared mandatory in their specifications.
ShortName is depreciated, but we use PublicCode primarily.

PR #480, opened 2019-11-26
[feature] NTFS: removed station_id from admin_stations

datanel:


tests/apply_rules.rs
```diff
@@ -37,9 +37,7 @@ fn test_apply_rules(
             "trips.txt",
             "stops.txt",
             "networks.txt",
-            "admin_stations.txt",
             "ticket_use_perimeters.txt",
-            "admin_stations.txt",
```
woshilapin: It looks like you removed all `admin_stations.txt` from the fixtures. But maybe we should keep at least one to check we're keeping this file and the information in it?

datanel: no need for this test

PR #481, opened 2019-11-26
[feature] NTFS: removed zone_id field from stops

datanel:


woshilapin:
Less code, these are the best PR :tada:

PR #483, opened 2019-11-27
Deprecate model-builder crate

woshilapin:
This Model Builder is only partial (it doesn't allow the creation of networks and lines for example). The problem is that we have ongoing work to sanitize automatically the `Model` on creation (in the `new()` method) which means that all the tests are breaking because a model that have `stop_times` which are not linked to a `line` are now sanitized...
Ping @antoine-de

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,7 +30,6 @@ travis-ci = { repository = "CanalTP/transit_model" }
 members = [
   "collection",
   "relations",
-  "model-builder",
   "transit_model_procmacro",
 ]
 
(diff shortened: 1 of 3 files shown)


antoine-de:
would it be really difficult to make it pass the sanitizer?

It's a bit of a shame for us to deprecate this, but frankly it's more for you. I think you underestimate the use of such tools for easy and maintainable integration tests. A similar tool has been **very** useful for Navitia.

woshilapin:
It would mean making some significant improvements to the `model-builder` to make it support more objects that it actually supports (`line` and `network` for example but that's not exhaustive).

Note however, we didn't remove the code. So it's still up to someone to make these improvements and reintegrate it into the CI and everything. We don't have yet plans to remove the source code, we just disconnected it completely from the parent project.

PR #486, opened 2019-11-29
[tech] Regroup all generic business rules into Model::new()

woshilapin:
This is the beginning of a long serie of regroupment of all the generic business rules that should be automatically applied on all models.


src/documentation/common.md
```diff
@@ -86,7 +86,7 @@ Train             |  11.9
 Tramway           |   4
 
 The following fallback modes are also added to the model (they're usually not
-referenced by any Vehicle Journey).
+referenced by any trip).
```
datanel: we definitively choose between vehicle journey or trip
@prhod @papailio ?

woshilapin: Not necessarily everywhere but I remember using the same word in the same document in our last (pair programming) PR with @prhod. This one must have been slipped out but the idea was to have the same word all over this document.

PR #489, opened 2019-11-29
[doc] Update doc on fares V2>V1

papailio:


src/documentation/ntfs_fare_v2_to_v1_specs.md
```diff
@@ -51,11 +51,11 @@ This file must be created without any data.
 ## Conversion of a flat fare on a specific network
 
 ### Description in V2 model
-An flat fare on a specific network is described in the V2 fare sytem as:
+A flat fare on a set of specific networks is described in the V2 fare sytem as:
 * A ticket in `tickets.txt` (with an ID) with:
   * A `ticket price` with a validity period, and only a euro currency
   * One `ticket use` without constrain on transfers, and empty bording and alighting times
-  * One `ticket use perimeter` with only the specified network (and no exclusion)
+  * One `ticket use perimeter` with the specified networks, all associated to the same `ticket_use` (and no exclusion)
```
woshilapin: I'm not sure I understand completely here. Is there multiple `ticket_use_perimeter` to define (one for each `network`)? If yes, maybe something like this

```suggestion
  * One `ticket use perimeter` for each specified networks, all associated to the same `ticket_use` (and no exclusion)
```

src/documentation/ntfs_fare_v2_to_v1_specs.md
```diff
@@ -51,11 +51,11 @@ This file must be created without any data.
 ## Conversion of a flat fare on a specific network
 
 ### Description in V2 model
-An flat fare on a specific network is described in the V2 fare sytem as:
+A flat fare on a set of specific networks is described in the V2 fare sytem as:
```
datanel:
```suggestion
A flat fare on a set of specific networks is described in the V2 fare system as:
```

src/documentation/ntfs_fare_v2_to_v1_specs.md
```diff
@@ -51,11 +51,11 @@ This file must be created without any data.
...
 * A ticket in `tickets.txt` (with an ID) with:
   * A `ticket price` with a validity period, and only a euro currency
   * One `ticket use` without constrain on transfers, and empty bording and alighting times
```
datanel:
```suggestion
  * One `ticket use` without constrain on transfers, and empty boarding and alighting times
```

src/documentation/ntfs_fare_v2_to_v1_specs.md
ArnaudOggy: > I'm not sure I understand completely here. Is there multiple `ticket_use_perimeter` to define (one for each `network`)? If yes, maybe something like this
:+1: 
Cardinalities are not clear (for an OD fare or flat fare).
One `ticket` may have
- multiple `ticket prices` (on Y, Y+1 for example)
- multiple `ticket use` (although usually only one)
- multiple `ticket use perimeter` for one `ticket use`
Same thing for `ticket use restriction`.
It's not a problem for v2 to v1 conversion.

prhod: When there is only one ticket_use_perimeter with several networks, the same ticket can be used on all networks and do connections between them.
If there is several  ticket_use_perimeter each containing a network, the ticket is usable in all networks, but you have to use a new one when having a transfer

ArnaudOggy: > When there is only one ticket_use_perimeter with several networks, the same ticket can be used on all networks and do connections between them.

You can only have one object (network here) by `ticket_use_perimeter`
https://github.com/CanalTP/ntfs-specification/blob/master/ntfs_fare_extension.md#ticket_use_perimeterstxt-optional

woshilapin:
I think I'm not competent enough to review this one. I'll leave others review and approve (ping @ArnaudOggy @prhod).

src/documentation/ntfs_fare_v2_to_v1_specs.md
prhod: @ArnaudOggy There could be multiple networks in one `ticket_use_perimeter`. It's said `at least one included line or network`. Am I missing something ?
The use case of a Train+Urbain ticket requires multiple networks.

ArnaudOggy: > There could be multiple networks in one `ticket_use_perimeter`. It's said `at least one included line or network`. Am I missing something ?
> The use case of a Train+Urbain ticket requires multiple networks.

Ok I understand what you mean.
You talk about file (ticket_use_perimeters.txt), I'm talking about objects (and csv lines in the file).

object_type | Text | Required | This field can be either line or network
object_id | Text | Required | This field refers to the ID of the considered line or network.

It's not a list. So one `ticket_use` must have n `ticket_use_perimeter` (not the file, but instances of the structure `ticket_use_perimeters`, so csv lines at final) to do Train+Urbain.

PR #491, opened 2019-12-02
[tech] Automatically generated route names if empty

woshilapin:


src/documentation/common.md
```diff
@@ -106,6 +106,13 @@ the same as the corresponding stop_point.
 - Any `/` character in an identifier of an object is removed.
 - If a trip doesn't have a `trip_headsign`, it is automatically generated based
   on the name of the last stop point of the trip
+- If a route doesn't have a name (or empty), it is automatically generated based
+  on the most frequent `stop_area` origin and most frequent `stop_area`
+  destination of all the associated trips (using `stop_area`'s' name): `[most
+  frequent origin] - [most frequence destination]`; in case of equal frequency,
+  the biggest `stop_area` (the most `stop_points`) is chosen; in case of
+  `stop_area` of equal size, the `stop_area` name are sorted alphabetically and
```
papailio: typo: `stop_area` name*s*

woshilapin: Done

src/model.rs
```diff
@@ -799,6 +799,150 @@ impl Collections {
...
+                    })
+            }
+            fn find_indexes_with_max_frequency(
+                frequencies: HashMap<Idx<StopArea>, usize>,
+            ) -> Vec<Idx<StopArea>> {
+                if frequencies.is_empty() {
+                    return Vec::new();
+                }
+                let first_key = *frequencies.keys().next().unwrap();
+                let mut max_frequency = *frequencies.get(&first_key).unwrap();
```
datanel: let mut max_frequency  = *frequencies.values().next().unwrap();

src/model.rs
```diff
@@ -799,6 +799,141 @@ impl Collections {
...
+    /// `stop_area` origin and `stop_area` destination of all the associated
+    /// trips.  The `stop_area` name is used to create the following `String`:
+    /// `[most frequent origin] - [most frequent destination]`
+    ///
+    /// If 2 stops have equal frequency, the biggest `stop_area` (biggest number
+    /// of `stop_point`) is chosen.
+    ///
+    /// If still equality between multiple `stop_area`, then alphabetical order
+    /// of `stop_area`'s name is used.
+    pub fn enhance_route_names(&mut self) {
```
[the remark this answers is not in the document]
woshilapin: I would have say the same 5 mn earlier. Then I tried to remove the specific `route.name` generation for `kv12ntfs`... Well, `Line` are generated based on existing `Route`, which means if the `Route` doesn't have a name, then the `Line` will not have a name either. So I will need to call this function directly in `kv12ntfs`. So I'll leave it public. But on principle, I agree with you, I'd be nice to keep them private.

src/model.rs
woshilapin: Hmm, historic reasons :stuck_out_tongue:

PR #492, opened 2019-12-03
[fix] Remove specific implementations for `route.name` generation in connectors

woshilapin:


tests/fixtures/kv12ntfs/output/lines.txt
```diff
@@ -1,4 +1,4 @@
 line_id,line_code,line_name,forward_line_name,backward_line_name,line_color,line_text_color,line_sort_order,network_id,commercial_mode_id,geometry_id,line_opening_time,line_closing_time
-prefix:2029,29,"Slagharen, Centrum - Slagharen, Attractiepark","Slagharen, Centrum - Slagharen, Attractiepark","Dedemsvaart, Busstation - Dedemsvaart, Adelaarweg",,,,prefix:SYNTUS,Bus,,,
+prefix:2029,29,"Hoogeveen, Willemskade - Hoogeveen, Willemskade","Hoogeveen, Willemskade - Hoogeveen, Willemskade","Dedemsvaart, Busstation - Dedemsvaart, Adelaarweg",,,,prefix:SYNTUS,Bus,,,
```
ArnaudOggy: We used the name of the stops previously, and now the name of the stopareas. 
Which gives this strange result, since the 2 stoppoints are in the same stoparea.
We will check on the production data set before, because it can be considered as a regression finally.

woshilapin: I wouldn't worry to check on production because we know this PR is a regression already: we have changed the way we calculate the `route.name`. I just happened to catch a bug on the way. But saying that, I realize I should update the `Cargo.toml` to `0.13.0`.

datanel: The api is not broken. To me it's a fix

ArnaudOggy: Just 2 cases in Production with the refactoring :
`Ede, Station Ede-Wageningen - Ede, Ziekenhuis Gelderse Vallei`
--> `Ede, Station Ede-Wageningen - Ede, Station Ede-Wageningen`
`Enschede, Centraal Station - Hengelo, Centraal Station`
--> `Hengelo, Centraal Station - Hengelo, Centraal Station`
Ok for me :+1: , we will see depending on possible returns.

woshilapin: @datanel, we're completely changing the way this field is generated so the output format is necessarily different from version `0.12.0`. You would consider this a non-breaking change? Should I move back to `0.12.1`?

datanel: We can consider that was a bug

woshilapin: I just changed it back to `0.12.1`.

src/documentation/netex_idfm_to_ntfs.md
```diff
@@ -210,9 +210,9 @@ NTFS field | Netex-IDFM element | Mapping rule/Comment
 --- | --- | ---
 route_id | Route/@id | This field is prefixed. 
 line_id | Route/LineRef/@ref | This field is prefixed. 
-route_name | Route/Name | 
+route_name | Route/Name | See [`common.md`](common.md#general-rules) to generate the `name`
```
prhod: To avoid an ambiguous state with a field indicating we use the `Route/Name` value and the text stating to see the an automatic generation of the name, what would you think of changing the text here to : 
`If Route/Name does not exists or is empty, it is generated automatically. See [...]`

woshilapin: Well, in this case, the Netex IDF is not giving any name so we know it is always gonna be empty. I guess what you're looking for is the generic section in [`common.md#general-rules`](https://github.com/CanalTP/transit_model/blob/master/src/documentation/common.md#general-rules) which indeed mention to do the modification only if the name is empty.

Does it answer your concern or did I missunderstood.

ArnaudOggy: @prhod , @woshilapin 
Go for merge guys ? It would be nice if it left tomorrow (in release 0.12.1)

prhod: Sorry for the late answer. This spec declares we read `Route/Name` as the Netex-IDFM element to be read.
@woshilapin I don't understand your point :)

ArnaudOggy: And actually this field is read, and with values ​​in the datasets provided.

prhod: Ok, let's keep it like this :)

woshilapin: My bad, I missed that one, I probably mixed things up with the KV1 connector. Good thing you didn't wait on me to merge. Thank you.

PR #498, opened 2019-12-09
Doc: Fare V2->V1 precisions and examples

prhod:
Following a discussion with @ArnaudOggy and @papailio, I'm proposing these modifications to indicate there could be unexpected behavior in navitia in some cases.

### DIFF
--- a/src/documentation/ntfs_fare_v2_to_v1_specs.md
+++ b/src/documentation/ntfs_fare_v2_to_v1_specs.md
@@ -50,7 +50,7 @@ Note that more that one `ticket uses` are allowed for the same `ticket` specifyi
 This file must be created without any data.
 
 
-## Conversion of a flat fare on a specific network
+## Conversion of a flat fare on specific networks
 
 ### Description in V2 model
 A flat fare on a set of specific networks is described in the V2 fare system as:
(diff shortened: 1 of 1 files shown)


src/documentation/ntfs_fare_v2_to_v1_specs.md
```diff
@@ -88,11 +90,40 @@ fares.csv field | Source file | Source field | Notes/Mapping rule
 "condition globale" |  |  | this field is empty
 "clef ticket" | tickets.txt | ticket_id |
 
-Then, another transition is defined allowing a transfer from a section of the given network to a section of each specified network (including the transition between two sections within the same network).
-All these transitions should use the same `ticket` (the field "clef ticket" should be left empty).
-In the end, all possible combinations of transitions among the specified networks should be defined.
-
-In case multiple `ticket uses` are specified for the same ticket defining different `ticket use perimeters`, the associated transitions for the `ticket` should be generated accordingly.
+Then, another transition is defined for each combinaison of all the networks referenced by the current `ticket_use` (including every network with itself). As transitions are directional, both combinasion (A->B and B->A) are created.
```
papailio: Then, another transition is defined for each combination of all the networks referenced by the current `ticket_use` (including every network with itself). As transitions are directional, both combination (A->B and B->A) are created.

src/documentation/ntfs_fare_v2_to_v1_specs.md
```diff
@@ -88,11 +90,40 @@ fares.csv field | Source file | Source field | Notes/Mapping rule
...
+Those now transitions have a `clef ticket` field empty to indicate the same ticket is to be used for the connection.
```
papailio: Those new transitions have a `clef ticket` field empty to indicate the same ticket is to be used for the connection.

PR #499, opened 2019-12-10
[tech] Gtfs2ntfs / Add test with a gtfs badly built (with hidden files)

ArnaudOggy:
Already functional, but an additional test on the case of a badly formatted gtfs is not a bad thing.


tests/gtfs2ntfs.rs
```diff
@@ -126,6 +126,16 @@ fn test_minimal_ziped_sub_dir_gtfs() {
     });
 }
 
+#[test]
+fn test_minimal_ziped_sub_dir_gtfs_with_hidden_files() {
+    test_in_tmp_dir(|path| {
+        let input = "./tests/fixtures/ziped_gtfs/sub_dir_gtfs_with_hidden_files.zip";
```
datanel: your zip contains a folder "input" at the root. This is what you want?

ArnaudOggy: > your zip contains a folder "input" at the root. This is what you want?

Yes. A subfolder + another system subfolder with hidden files. A dirty GTFS  :hankey:

PR #503, opened 2019-12-12
[tech] Apply 'sanitize' to 'merge_stop_areas'

woshilapin:


tests/fixtures/merge-stop-areas/output/stops.txt
```diff
@@ -9,5 +9,3 @@ SP:07,stop point name 7,1,,2.41,48.64,0,SA:11,Europe/Paris,,,,
 SP:08,stop point name 8,1,,2.41,48.64,0,SA:01,Europe/Paris,,,,
 SA:01,stop area name identical 3 SA,1,,2.38951,48.852245,1,,Europe/Paris,,,,
 SA:11,stop area name identical 2 SA,1,,2.390033,48.850663,1,,Europe/Paris,Geometry:3,,,
-SA:13,stop area name identical 2 SA not merged,1,,2.41,48.64,1,,Europe/Paris,,,,
-SA:14,stop area name identical 2 SA not merged,1,,2.41,48.65,1,,Europe/Paris,,,,
```
datanel: is it possible to keep this test?
those 2 stop areas have the same name but the distance between them is greater than 200m (see in the test) so  they are not merged.
It's an interesting case to keep

woshilapin: I can take a look to keep them, OK. Thanks for this really useful feedback as each time I'm modifying fixtures, I never completely know if I'm losing something in term of test coverage.

woshilapin: I added 2 `stop_point` (`SP:09` and `SP:10`) that respectively refers to `SA:13` and `SA:14` (the two `stop_area` you mentioned). Then I added additional `stop_time` in `trip:04` that refers to these 2 new `stop_point`. See the modification in 1dedcdc5a632ef712f3579d04168628a792095be.

datanel: :+1:

PR #504, opened 2019-12-13
[tech] apply sanitize when creating a model

datanel:


src/model.rs
```diff
@@ -999,22 +999,25 @@ impl Model {
...
     ///     real_min_transfer_time: None,
     ///     equipment_id: None,
     /// });
-    /// assert!(Model::new(collections).is_err());
+    /// assert!(Model::new(collections).is_ok());
     /// ```
     pub fn new(mut c: Collections) -> Result<Self> {
-        fn apply_generic_business_rules(collections: &mut Collections) {
+        fn apply_generic_business_rules(collections: &mut Collections) -> Result<()> {
+            collections.sanitize()?;
```
woshilapin: The ‘sanitize’ must be done in last. For example, ‘enhance_route_name’ might skip some lines if not able to generate a ‘route_name’. Therefore, some ‘trip’ might need to be sanitized.

datanel: ok. i might write in this order
route_name
sanitize
co2 and the other

to avoid useless processes objects that will removed

woshilapin: I wouldn't care so much for the order performance until it is really a problem. Let's prove it is a problem before doing such optimizations. For now, I would keep it simple and put it at the end.

datanel: done

src/model.rs
```diff
@@ -999,22 +999,25 @@ impl Model {
...
+    /// assert!(Model::new(collections).is_ok());
     /// ```
     pub fn new(mut c: Collections) -> Result<Self> {
-        fn apply_generic_business_rules(collections: &mut Collections) {
+        fn apply_generic_business_rules(collections: &mut Collections) -> Result<()> {
             collections.enhance_with_co2();
             collections.enhance_trip_headsign();
             collections.enhance_route_names();
+            collections.sanitize()?;
+            Ok(())
```
woshilapin: You can then just replace

```rust
collections.sanitize()?;
Ok(())
```

by

```rust
collections.sanitize()
```

PR #505, opened 2019-12-19
[fix] removed trips with empty stop times

datanel:
- remove trips with empty stop times
- warn if a trip has 1 stop time
- remove calendars not referenced by a trip

Not done: panic if there is no calendars. This breaks a lot a tests. We can discuss and do it later


datanel:
@Mergifyio refresh

PR #507, opened 2019-12-20
[tech] added github workflow

datanel:


woshilapin:
Do we want to change from TravisCI to Github Workflow? Any advantages compared to TravisCI which, so far, seems to do the job pretty well?

datanel:
It's just a test to see if it's better.and I wonder if it ca be used for private repositories.

woshilapin:
So I experimented a little bit more with Github Actions when converting [CanalTP/kirin](https://github.com/CanalTP/kirin) and I found a few more things.

For setting up Rust environment, it seems that there is the following actions already implemented that we could use:
https://github.com/actions-rs/toolchain

And more generally, it might be nice to take a look to the repository https://github.com/actions-rs?type=source.

.github/workflows/main.yml
```diff
@@ -0,0 +1,56 @@
...
+        rustup default stable
+        rustup component add rustfmt
+    - run: cargo fmt --all -- --check
+
+  clippy:
+    name: Clippy
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@master
+    - name: Install Rust
```
woshilapin: You should be able to replace this step with the following code.

```yaml
    - name: Install Rust
      uses: actions-rs/toolchain@v1
      with:
        toolchain: stable
        override: true
        default: true
        profile: minimal
        component: clippy
```

.github/workflows/main.yml
```diff
@@ -0,0 +1,56 @@
...
+      run: |
+        rustup update stable
+        rustup default stable
+        rustup component add clippy
+    - run: cargo clippy --workspace --all-features -- -D warnings
```
woshilapin: You should be able to use the action `actions-rs/cargo@v1` for this step.

```yaml
    - name: Run Clippy
      uses: actions-rs/cargo@v1
      with:
        command: clippy
        args: --workspace --all-features -- -D warnings
```

.github/workflows/main.yml
```diff
@@ -0,0 +1,56 @@
...
+    steps:
+    - name: Install proj
+      run: |
+        wget --quiet --output-document - "https://kisiodigital.jfrog.io/kisiodigital/api/gpg/key/public" | sudo apt-key add -
+        sudo echo "deb [arch=amd64] https://kisiodigital.jfrog.io/kisiodigital/debian-local stretch main" > /etc/apt/sources.list.d/kisio-digital.list
+        sudo apt update && sudo apt install --yes pkg-config libssl-dev clang proj=${PROJ_VERSION}
+    - uses: actions/checkout@master
+    - name: Install Rust (rustup)
+      run: |
+        rustup update ${{ matrix.rust }}
```
woshilapin: Nice use of self-defined variables in the `matrix` section.

woshilapin:
With the feature `xmllint`, you might also need 2 things if you want the tests to pass:
- Install `libxml2-utils` on the machine
- Pull the submodules (see below)

To pull the submodules, you can use the following step.

```yaml
            - name: Checkout Submodules
              uses: textbook/git-checkout-submodule-action@2.1.1
```

woshilapin:
Minor: the name of the file of the workflow is not really explicit about what it does `main.yml`. How about naming it `continuous-integration.yml`?

PR #508, opened 2019-12-20
[tech] centralize computation of stop areas coords

datanel:


src/model.rs
```diff
@@ -947,6 +949,35 @@ impl Collections {
...
+                .collect::<MultiPoint<_>>()
+                .centroid()
+                .map(|c| Coord {
+                    lon: c.x(),
+                    lat: c.y(),
+                })
+            {
+                stop_area.coord = coord;
+            } else {
+                log::warn!("failed to calculate a centroid of stop area {} because it does not refer to any corresponding stop point", stop_area.id)
```
woshilapin: Maybe a `use log::warn;` at the top of the file to stay consistent in the repo?

src/model.rs
```diff
@@ -1614,4 +1646,59 @@ mod tests {
...
+        fn collections(sa_id: &str) -> Collections {
+            let mut collections = Collections::default();
+            collections.stop_areas = stop_areas();
+            collections.stop_points = stop_points(sa_id);
+            collections
+        }
+
+        fn stop_areas() -> CollectionWithId<StopArea> {
+            CollectionWithId::from(StopArea {
+                id: "stop_area:1".into(),
```
woshilapin: It took me a minute to understand why this function doesn't take `sa_id` as a parameter and use it to create the `StopArea`. This is misleading to understand what is done in these test. I was thinking you might be able to parametrize instead the number of `StopPoint`.

```rust
fn collections(sp_amount: usize) -> Collections {
    collections.stop_areas = stop_areas();
    collections.stop_points = stop_points(sp_amount);
    collections
}

fn stop_areas() -> CollectionWithId<StopArea> {
    CollectionWithId::from(
        id: "stop_area:1".into(),
        name: "Stop Area 1".into(),
        coord: Coord::default(),
        ..Default::default()
    })
}

fn stop_points(sp_amount: &str) -> CollectionWithId<StopPoint> {
    CollectionWithId::new(
        (0..sp_amount)
            .map(|index| StopPoint {
                id: format!("stop_point:{}", index),
                stop_area_id: "stop_area:1".into(),
                coord: Coord {
                    lon: index as f64,
                    lat: index as f64,
                },
                ..Default::default()
            })
            .collect(),
    )
    .unwrap()
}
```

And then in your first test, you use `collections(5)` then in the second, `collections(0)`. What do you think?

datanel: I wanted stop points that don't reference a stop area, not 0 stop points.
The result is the same but I can change if it bothers you

woshilapin: Well, it's more the fact that everything seems to be parametrized for `stop_area_id` bit then the function `stop_areas` doesn't use this parameter that bother me. Maybe what I propose is not the only solution nor the best solution.

PR #509, opened 2020-01-08
[tech] Update conversion of fares v2 to fares v1

pbench:
ticket : https://jira.kisio.org/browse/NAVP-1465

TODO : 

- [x] merge the related updates of navitia https://github.com/CanalTP/navitia/pull/3106
- [x] update the fare specs in https://github.com/CanalTP/ntfs-specification/pull/12
- [x] add tests


woshilapin:
One more thing, it'd be nice that you also update the specifications as part of this PR.

src/ntfs/write.rs
```diff
@@ -32,6 +32,13 @@ use transit_model_collection::{Collection, CollectionWithId, Id, Idx};
 impl TryFrom<(&Ticket, &TicketPrice)> for PriceV1 {
     type Error = failure::Error;
     fn try_from((ticket, price): (&Ticket, &TicketPrice)) -> Result<Self> {
+        if price.currency != "Eur" {
```
woshilapin: As discussed, it might be slightly more complex than that, we need to check if the input model is multi-currency or not. If only one currency is used, we can convert to FareV1. If multi-currencies, we need to panic.

pbench: As discussed with @prhod and @papailio if the currency is not EUR, a warning is logged and the faulty price item is ignored

Cargo.toml
```diff
@@ -31,6 +31,7 @@ members = [
   "collection",
   "relations",
   "transit_model_procmacro",
+  "fare_conversion",
```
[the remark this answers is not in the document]
pbench: I removed the subcrate, will use tartare-tools instead

src/documentation/ntfs_fare_v2_to_v1_specs.md
```diff
@@ -1,129 +1,142 @@
...
+ticket_id   | ticket_price| ticket_currency| ticket_validity_start| ticket_validity_end
+---         | ---         | ---            | ---                  | ----
+my_ticket_id| 1.13        | EUR            | 20190101             | 20191231
+
+
+ticket_uses.txt
+
+ticket_use_id | ticket_id    | max_transfers | boarding_time_limit | alighting_time_limit
+---           | ---          | ---           | ---                 | ----
+my_use_id     | my_ticket_id |   3           |   60                | 90
```
ArnaudOggy:
```suggestion
my_use_id     | my_ticket_id |   2           |   60                | 90 
```
to avoid changing all nb_changes<4

src/ntfs/write.rs
```diff
@@ -224,181 +200,283 @@ struct Fares<'a> {
...
-}
-
-fn get_prices<'a>(
-    ticket_prices: &'a Collection<TicketPrice>,
-    ticket_id: &str,
-) -> Vec<&'a TicketPrice> {
-    ticket_prices
+// Returns Ok(()) if each ticket_use_id appears at most once in fares
+// returns an error otherwise
+fn check_uniqueness_of_ticket_use_ids(fares: &Fares) -> Result<()> {
```
ArnaudOggy: Uniqueness is guaranteed on ticket_uses (it's a CollectionWithId). 
I hardly see the relevance of this function. Did you have the case?

src/ntfs/write.rs
```diff
@@ -224,181 +200,283 @@ struct Fares<'a> {
...
-        if prices.is_empty() {
-            warn!("Failed to find TicketPrice for Ticket {:?}", ticket.id);
+fn extract_perimeter_for_ticket_use<'id, 'p>(
+    ticket_use_id: &'id str,
+    ticket_use_perimeters: &'p Collection<TicketUsePerimeter>,
+) -> Result<(Vec<&'p str>, Vec<&'p str>, Vec<&'p str>)> {
+    let mut included_networks = Vec::new();
+    let mut included_lines = Vec::new();
+    let mut excluded_lines = Vec::new();
+    for perimeter in ticket_use_perimeters.values() {
```
ArnaudOggy:
```suggestion
    for perimeter in ticket_use_perimeters.values().filter(|p| p.ticket_use_id == ticket_use_id) { 
```
More rustic

src/ntfs/write.rs
```diff
@@ -224,181 +200,283 @@ struct Fares<'a> {
...
         }
 
-        for price in prices {
-            prices_v1.insert(PriceV1::try_from((ticket, price))?);
-        }
+        // Now the restrictions for our ticket_use_id
+        let restrictions: Vec<&TicketUseRestriction> = fares
+            .ticket_use_restrictions
+            .values()
+            .filter(|restriction| restriction.ticket_use_id.as_str() == ticket_use.id)
```
ArnaudOggy:
```suggestion
            .filter(|restriction| restriction.ticket_use_id == ticket_use.id)
```
No need to convert in str. Both are Strings.

src/ntfs/write.rs
```diff
@@ -239,181 +215,283 @@ struct Fares<'a> {
...
+        // we find all prices with id ticket.id
+        // and for each we create a price_v1 with id ticket_use_id (as ticket_use_id of fare v2 plays the role of ticket_id in fare v1)
+        for price in fares
+            .ticket_prices
             .values()
-            .filter(|p| p.object_type == ObjectType::Network)
-            .filter(|p| p.perimeter_action == PerimeterAction::Included)
-            .filter(|p| p.ticket_use_id == ticket_use_perimeter.ticket_use_id)
-            .filter(|p| p.object_id != ticket_use_perimeter.object_id)
+            .filter(|&ticket_price| ticket_price.ticket_id == ticket.id)
```
ArnaudOggy:
```suggestion
            .filter(|ticket_price| ticket_price.ticket_id == ticket.id)
```

src/ntfs/write.rs
```diff
@@ -239,181 +215,283 @@ struct Fares<'a> {
...
+            //   and it's unclear how to provide this information (which evolves over time)
+            if price.currency != "EUR" {
+                warn!(
+                    "The price {:?} is ignored as it has an unsupported currency : {}. \
+                     Only EUR currency supported in conversion from fare v2 to fare v1.",
+                    price, price.currency
+                );
+                continue;
+            }
+            let price_v1 = build_price_v1(&ticket_use.id, ticket, price)?;
```
ArnaudOggy:
```suggestion
            let price_v1 = build_price_v1(&ticket_use.id, &ticket, &price)?;
```

src/ntfs/write.rs
```diff
@@ -239,181 +215,283 @@ struct Fares<'a> {
...
+                .chain(
+                    included_lines
+                        .iter()
+                        .map(|line| format!("line=line:{}", line)),
+                );
+
+            // will yield a sequence of String
+            // each  corresponds to a start_trip condition
+            //  in FareV1
+            // these conditions must appears in all transitions (i.e. lines of fares.txt)
```
ArnaudOggy:
```suggestion
            // these conditions must appears in all transitions (i.e. lines of fares.csv)
```

src/ntfs/write.rs
```diff
@@ -239,181 +215,283 @@ struct Fares<'a> {
...
+        //now fares_v1
+        {
+            let states = included_networks
+                .iter()
+                .map(|network| format!("network=network:{}", network))
+                .chain(
+                    included_lines
+                        .iter()
+                        .map(|line| format!("line=line:{}", line)),
+                );
```
ArnaudOggy:
```suggestion
                )
                .collect::<Vec<String>>();
```
if you collect here you can avoid an extra clone below (and just using a reference).
```&states``` instead of ```states.clone()```

File: src/documentation/ntfs_fare_v2_to_v1_specs.md (hunk 3 above)
Comment (pbench) [reply src/documentation/ntfs_fare_v2_to_v1_specs.md]: well spotted ! fixed

File: src/ntfs/write.rs (hunk 4 above)
Comment (pbench) [reply src/ntfs/write.rs]: you're right ! I removed the function

File: src/ntfs/write.rs (hunk 5 above)
Comment (pbench) [reply src/ntfs/write.rs]: :+1:

File: src/ntfs/write.rs (hunk 10 above)
Comment (pbench) [reply src/ntfs/write.rs]: but then it will allocate a vector and all the strings inside it, whereas for now it just clones the iterator, right ?

Comment (ArnaudOggy) [reply src/ntfs/write.rs]: @pbench Right. But you can avoid multiple clones afterwards (and therefore memory allocation).

Comment (pbench) [reply src/ntfs/write.rs]: allright ! updated

**Comment by ArnaudOggy** [comment]
@pbench 
Just to confirm, Navitia accepts (without flinching) tickets without fare ? (existing case in Production and well managed in your PR)

File: src/documentation/ntfs_fare_v2_to_v1_specs.md [hunk 11]
```diff
@@ -1,129 +1,142 @@
...
+fares.csv
+
+| avant changement | apres changement   | debut trajet                                      | fin trajet    | condition globale | clef ticket |
+| ---              |    ---             |  ---                                              | ---           |   ---             |  ---        |
+|                  | network=my_network | line!=excluded_line & duration<61 & nb_changes<3  | duration<91   |                   | my_use_id   |
+|                  | line=my_line       | line!=excluded_line & duration<61 & nb_changes<3  | duration<91   |                   | my_use_id   |
+
+Each line allows entering one of the two included perimeters : the network my_network and the line my_line.
+Note that the exluded perimeter (excluded line) is added as a condition "debut trajet".
+The conditition "debut trajet" also encodes the contraints on the boarding duration and the number of transfers.
```
woshilapin:
```suggestion
The condition "debut trajet" also encodes the constraints on the boarding duration and the number of transfers.
```

src/ntfs/write.rs
```diff
@@ -239,181 +215,250 @@ struct Fares<'a> {
...
-
-fn get_prices<'a>(
-    ticket_prices: &'a Collection<TicketPrice>,
-    ticket_id: &str,
-) -> Vec<&'a TicketPrice> {
-    ticket_prices
+fn extract_perimeter_for_ticket_use<'id, 'p>(
+    ticket_use_id: &'id str,
+    ticket_use_perimeters: &'p Collection<TicketUsePerimeter>,
+) -> Result<(Vec<&'p str>, Vec<&'p str>, Vec<&'p str>)> {
```
woshilapin: Just wondering how to make this return value more explicit (a tuple of 3 `Vec` of references on `str` does not make it very clear about what it is).

I'm wondering if it would be possible to return something like `Result<Vec<Perimeter>>` where `Perimeter` would be something like.

```rust
enum Perimeter {
  IncludeNetwork(String),
  IncludeLine(String),
  ExcludeLine(String),
}
```

Another more simple trick but that helps self-documenting would be to do something like

```rust
type IncludedNetworks<'p> = Vec<&'p str>;
type IncludedLines<'p> = Vec<&'p str>;
type ExcludedLines<'p> = Vec<&'p str>;
```
and then your function would return a type of more explicit types. This second solution is only aliases on names.

src/ntfs/write.rs
```diff
@@ -239,181 +215,250 @@ struct Fares<'a> {
...
-                })
-        });
-    for (ticket, ticket_use_perimeter) in ticket_use_perimeters {
-        let prices = get_prices(fares.ticket_prices, &ticket.id);
-        if prices.is_empty() {
-            warn!("Failed to find TicketPrice for Ticket {:?}", ticket.id);
-            continue;
-        }
+        // Now the restrictions for our ticket_use_id
+        let restrictions: Vec<&TicketUseRestriction> = fares
```
woshilapin: On each ticket, the iteration is realized on the entirety of `fares.ticket_use_restriction`. There might be a way to be more efficient by storing these relations in the `Model` and then use `get_corresponding()` instead of the current iteration. We can discuss it if you want.

ArnaudOggy:
As seen with @prhod we skip a ticket without -valid- price (in v1 ouput). 
A PR is waiting for you here https://github.com/pbench/transit_model/pull/3

src/ntfs/write.rs
datanel: @woshilapin be careful not to create short circuits by adding relations.
You can break everything

ArnaudOggy: Yes, fares are not in graph

datanel: I mean if we add more relations, get_corresponding may not work as before. the paths between nodes in the graph may change and result a bad behaviour

ArnaudOggy: There is still a lot of work to do on fares v2 (checking cardinalities and relations between the objects in the 5 collections themselves, and others, stopareas, lines, networks, etc.). We have a ticket in the backlog.

ArnaudOggy: In fact, we already had this debate there. hence the existence of the ticket in the backlog :grinning:

src/documentation/ntfs_fare_v2_to_v1_specs.md
pbench: well spotted, thanks !

src/ntfs/write.rs
pbench: Good idea, I added this struct as a return type 
```rust
struct Perimeter<'p> {
    included_networks : Vec<&'p str>,
    included_lines : Vec<&'p str>,
    excluded_lines : Vec<&'p str>
}
```

src/ntfs/write.rs
pbench: @woshilapin if you go down this way then, for each `ticket_use` :

- the whole `fares.ticket_use_perimeters` is scanned in `extract_perimeters_for_ticket_use(..)`
- the whole `fares.ticket_use_restrictions` is scanned as you point
- the whole `fares.ticket_prices`  is also scanned

Correct me if I'm wrong, by it seems to me that using `get_corresponding` is doing a `filter_map`, which is more or less equivalent to what happens here (?)

I think it may be possible to scan each file only once, and incrementally build a data structure as below, and then generate the output. 
Not sure if it is worth the effort though !

```rust 

struct MyFaresV2 {
    ticket_uses : HashMap<String,TicketUse> // maps a ticket_use_id to its data
    ticket_prices : HashMap<String, Vec<Price> > // maps a ticket_id to its prices
}

struct Price {
    currency : String,
    price : Decimal,
    validity_start : Date,
    validity_end : Date, 
}

struct TicketUse {
    id : String,
    ticket_id : String,
    ticket_name : String,
    ticket_comment : String,
    max_transfers : i64,
    boarding_time_limit : i64,
    alighting_time_limit : i64,
    included_networks : Vec<String>,
    included_lines : Vec<String>,
    excluded_lines : Vec<String>,
    restrictions : Vec<Restriction>,
}

enum Restriction {
    Zone(String, String),
    OD(String, String)
}

```

woshilapin: What you propose is pretty much what is doing `get_corresponding` when creating the object `Model`. When `Model` is built, all relations are created once.

That said, we discussed it in daily this morning and we might address this refactorization later. For now, let's not put more complexity to the current PR.

PR #510, opened 2020-01-10
[tech] Skeleton for the Netex France exporter

woshilapin:
Just a skeleton for the exporter. Trying something new in term of architecture with a `struct` (instead of a function) that will own the `Model` and be responsible for the export. The reason is that we discussed a lot of possible parameters to pass to the binary; the `Exporter` is the structure that would also have these parameters.

Note: Putting that PR as a Draft for now. Please tell me what you think of it.

### DIFF
--- a/src/netex_france/exporter.rs
+++ b/src/netex_france/exporter.rs
@@ -1,0 +1,116 @@
+// Copyright (C) 2017 Kisio Digital and/or its affiliates.
+//
+// This program is free software: you can redistribute it and/or modify it
+// under the terms of the GNU Affero General Public License as published by the
+// Free Software Foundation, version 3.
+
+// This program is distributed in the hope that it will be useful, but WITHOUT
+// ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+// FOR A PARTICULAR PURPOSE. See the GNU Affero General Public License for more
+// details.
+
+// You should have received a copy of the GNU Affero General Public License
+// along with this program. If not, see <https://www.gnu.org/licenses/>
+
+//! Exporter for Netex France profile
+use crate::{model::Model, Result};
+use chrono::prelude::*;
+use minidom::Element;
+use std::{
+    convert::AsRef,
+    fmt::{self, Display, Formatter},
+    fs::File,
+    path::Path,
+};
+
+const NETEX_FRANCE_STOPS_FILENAME: &str = "arrets.xml";
+
+enum VersionType {
+    Stops,
+}
+
+impl Display for VersionType {
+    fn fmt(&self, fmt: &mut Formatter) -> std::result::Result<(), fmt::Error> {
+        use VersionType::*;
+        match self {
+            Stops => write!(fmt, "ARRET"),
+        }
+    }
+}
+
+/// Struct that can write an export of Netex France profile from a Model
+pub struct Exporter {
+    _model: Model,
+    participant_ref: String,
+    timestamp: NaiveDateTime,
+}
+
+// Publicly exposed methods
+impl Exporter {
+    /// Build a Netex France profile exporter from the model.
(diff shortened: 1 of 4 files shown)


src/netex_france/exporter.rs
```diff
@@ -0,0 +1,43 @@
...
+{
+    /// Build a Netex France profile exporter from the model.
+    /// `path` is the expected output Path where the Netex France is going to be
+    /// written. It should be a folder that already exists.
+    pub fn new(model: Model, path: P) -> Self {
+        Exporter { model, path }
+    }
+
+    /// Actually write `model` into `path` as a Netex France profile.
+    pub fn write(&self) -> Result<()> {
```
datanel: I prefer `write(&self, path: P)` if path is not used elsewhere

woshilapin: Good point!

woshilapin:
In last commit 742b1a4afd4313f5451a96ab2283634f59499087, I added a few things:
- A new parameter to `Exporter` object which will contain the `ParticipantRef` value (an input parameter of the program)
- A method `write_stops` which returns an `Element` which should be a `GeneralFrame`
- A method to wrap the frame (see above) inside a entire NeTEx block.

src/netex_france/exporter.rs
```diff
@@ -0,0 +1,113 @@
...
+use chrono::prelude::*;
+use minidom::Element;
+use std::{
+    convert::AsRef,
+    fmt::{self, Display, Formatter},
+    fs::File,
+    path::Path,
+};
+
+const NETEX_FRANCE_STOPS_FILENAME: &'static str = "arrets.xml";
```
datanel: clippy -> `&str`

src/netex_france/exporter.rs
```diff
@@ -0,0 +1,113 @@
...
+            .ns("http://www.netex.org.uk/netex/")
+            .append(frame)
+            .build();
+        let root = Element::builder("PublicationDelivery")
+            .ns("http://www.netex.org.uk/netex/")
+            .attr("xmlns:xsi", "http://www.w3.org/2001/XMLSchema-instance")
+            .attr("version", format!("1.09:FR-NETEX_{}-2.1-1.0", version_type))
+            .append(publication_timestamp)
+            .append(participant_ref)
+            .append(data_objects)
```
datanel: I'm not sure but this node is not wrapped by `<CompositeFrame><frames></frames></CompositeFrame>`?

woshilapin: This is still something to discuss with @prhod. From what I understood of NeTEx specification (well, the XSD files), you need to wrap in a `CompositeFrame` only if there is more than one frame. In the fixtures we do have for IDFM, they have 2 `GeneralFrame` but only one contains the stops. The other one seems useless and we might not produce it for NeTEx France. So if we do have only one to produce, we might just skip this extra level of `CompositeFrame`. But as I said, still to be discussed with @prhod.

src/netex_france/exporter.rs
woshilapin: Yep, I just pushed that after I saw the CI complaining :smile:

woshilapin:
One more thing, it'd be nice that you also update the specifications as part of this PR.

src/netex_france/exporter.rs
woshilapin: So, decision has been taken to go for the simplest solution for now (without a `CompositeFrame`). We're not 100% sure it's completely correct yet but it should fairly easy to wrap it in a `CompositeFrame` later if needed.

woshilapin:
I just added this commit 855e57cb6fb6efd0af1baa5d5b8809473190a2f9 to change the way were gonna write timestamp in the file.

datanel:
@Mergifyio refresh

ArnaudOggy:
@Mergifyio refresh

PR #516, opened 2020-01-20
[feature] Support of 'stop_code' in 'stops.txt'

woshilapin:


src/documentation/common.md
```diff
@@ -98,9 +98,12 @@ Car                | 184
 The following rules apply to every converter, unless otherwise explicitly specified.
 
 ### General rules
-- When one or more stop_points in the input data are not attached to a stop_area, 
-a stop_area is automatically created for each one. The coordinates and the name of the new `stop_area` are
-the same as the corresponding stop_point.
+- When one or more stop_points in the input data are not attached to a
+  stop_area, a stop_area is automatically created for each one. The name, the
+  coordinates, the code, the visibility, the timezone, the equipment, the level
```
prhod: You felt in the trap of the current model :wink:  . A stop_area should not have stop_code, equipment, level. The stop_point doesn't have a geometry, so keeping it doesn't make sense

woshilapin: As discussed, `StopArea` won't support `stop_code`, only `StopPoint` and `StopLocation` (see changes 3af9142b658f43dadc51fdff5f99895bdc1ad533). A known effect is: if a GTFS uses `stop_code` for a `stop_area`, a conversion GTFS > NTFS > GTFS would loose the `stop_code` information on every `stop_area` (but not the other types of stops).

I updated the specifications in `common.md`. I also removed `equipment_id`, `level_id` and `geometry_id` for the creation of a `StopArea` from a `StopPoint`.

prhod:
Good for me at functional level. I'm not approving to let 2 code reviews :wink:

woshilapin:
I just bumped the version to 0.14.0 in dca6e3e2d5e90baa8579a8df4658110e4c48cee1 since we just released (actually, I should already have done that).

woshilapin:
> you need to rebase

Indeed, thanks. I just did it.

PR #526, opened 2020-01-29
[feature] Export Lines for NeTEx France

woshilapin:


src/netex_france/exporter.rs
```diff
@@ -183,6 +186,20 @@ impl Exporter<'_> {
         Ok(frames)
     }
 
+    // Returns a list of 'ServiceFrame' each containing a 'Network'
```
ArnaudOggy: // Returns a list of 'Line'

woshilapin: Is it so obvious I used copy/paste to often? :roll_eyes:

woshilapin: Fixed in 8d0bd72034c3f87d460c28ccd9dbcbace0216c5c.

PR #534, opened 2020-02-04
[feature] Add ValidBetween for calendriers.xml in NeTEx France

woshilapin:


src/netex_france/exporter.rs
```diff
@@ -258,8 +260,36 @@ impl Exporter<'_> {
...
             .attr("version", "any")
+            .append(valid_between)
             .append(members)
             .build();
         Ok(frame)
     }
+
+    fn create_valid_between(&self) -> Result<Element> {
+        let get_date = |feed_info_key, hour, minute, second| -> Result<String> {
+            let ntfs_string = self.model.feed_infos.get(feed_info_key).ok_or_else(|| {
```
ArnaudOggy: You cannot use the feed_infos. In the case of an input gtfs (**our initial request**), the file is optional (and is also not read). The collection is empty. However datasets collection is necessarily filled.
Change your recovery method using
```
model.datasets.values().min_by_key(|d| d.start_date) 
model.datasets.values().max_by_key(|d| d.end_date)
```

woshilapin: Nice catch. However, I'm not convinced of the solution. From the [NTFS Specification of `feed_infos.txt`](https://github.com/CanalTP/ntfs-specification/blob/master/ntfs_fr.md#description-du-contenu-du-fichier), `feed_start_date` and `feed_end_date` are mandatory. So I believe that is a bug of the `gtfs::read()` method. This should be fixed on the input, not on the output. What do you think?

woshilapin: I pushed a modification to solve this in 5bd92c3c3ee65991e8fd4de16fa8f2f5ac7f7529. However, it depends on another PR to be merged first #538.

PR #538, opened 2020-02-05
[tech] Calculate Validity period is done in Collections

woshilapin:


src/model.rs
```diff
@@ -988,6 +988,24 @@ impl Collections {
...
+    /// The calculation is based on the minimum start date and the maximum end
+    /// date of all the datasets.
+    /// If no dataset is found, an error is returned.
+    pub fn calculate_validity_period(&self) -> Result<(Date, Date)> {
+        let start_date = self
+            .datasets
+            .values()
+            .map(|dataset| dataset.start_date)
+            .min();
+        let end_date = self.datasets.values().map(|dataset| dataset.end_date).max();
```
datanel: is it better than `max_by_key` like before?

woshilapin: `max_by_key` returns the `Dataset` but I only care of the `Date`.

PR #539, opened 2020-02-05
[feature] FilterNtfs: filter by line_code

datanel:
It works but it's ugly.
something like https://github.com/CanalTP/transit_model/compare/master...datanel:wrap_new_filter_ntfs?expand=1 would be better but it does not compile because model that is moved.

### DIFF
--- a/tests/filter_ntfs.rs
+++ b/tests/filter_ntfs.rs
@@ -21,12 +21,10 @@ fn test_extract_network() {
     test_in_tmp_dir(|path| {
         let input_dir = "tests/fixtures/filter_ntfs/input";
 
-        let model = filter::filter(
-            transit_model::ntfs::read(input_dir).unwrap(),
-            filter::Action::Extract,
-            vec!["network1".into()],
-        )
-        .unwrap();
+        let mut filter = filter::Filter::new(filter::Action::Extract);
+        filter.add(filter::ObjectType::Network, "network_id", "network1");
+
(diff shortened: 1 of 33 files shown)


woshilapin:
I did a big refactor on this PR. The nice thing coming out of it is that it should easy now to add support for new properties and new object (like Route). Please have a look.

woshilapin:
As discussed with @ArnaudOggy, we removed completely the notion of `calendar` from the `filter()` function. The `sanitize()` already does a good job at filtering them. See fbb2dd6588fb1bc25f6f59b9939b3c0976538fb4.

tests/filter_ntfs.rs
```diff
@@ -41,37 +39,50 @@ fn test_remove_network() {
...
-    .unwrap();
+    let mut filter = filter::Filter::new(filter::Action::Remove);
+    filter.add(filter::ObjectType::Network, "network_id", "network1");
+    filter.add(filter::ObjectType::Network, "network_id", "network2");
+    filter.add(filter::ObjectType::Network, "network_id", "network3");
+    filter::filter(transit_model::ntfs::read(input_dir).unwrap(), &filter).unwrap();
+}
+
+#[test]
+fn test_remove_line_by_line_code() {
```
ArnaudOggy: Can you a test ```test_extract_lines_by_line_code``` with route1 and route2 ?

woshilapin: I added this test with `route1` and `route3`. See 0a8014bd7e7ba1b184cd45586bc3744e4b8dafa8.

PR #543, opened 2020-02-11
[feature] Check coherence of Geometries in the model and fix it

woshilapin:


src/model.rs
```diff
@@ -989,6 +989,41 @@ impl Collections {
...
+    /// doesn't exist.
+    ///
+    /// This function checks that all objects points to existing `Geometry` and,
+    /// in the case it doesn't, fix the model by removing this pointer.
+    fn check_geometries_coherence(&mut self) {
+        macro_rules! check_and_fix_object_geometries {
+            ($collection:expr) => {
+                let objects_to_fix: Vec<String> = $collection
+                    .values()
+                    .filter(|line| {
```
datanel: I suggest to use obj instead of line as it also concerns others objects

tests/fixtures/apply_rules/output_report/report_apply_property.json
```diff
@@ -62,8 +62,8 @@
       "message": "object_type=route, object_id=RERAB:2, property_name=route_geometry: property_old_value does not match the value found in the data"
     },
     {
-      "category": "ObjectNotFound",
-      "message": "object_type=route, object_id=RERAF: geometry unknown not found"
+      "category": "OldPropertyValueDoesNotMatch",
+      "message": "object_type=route, object_id=RERAF, property_name=route_geometry: property_old_value does not match the value found in the data"
```
datanel: there is now dead code. https://github.com/CanalTP/transit_model/blob/master/src/apply_rules/property_rule.rs#L397

you may remove the code and add a comment that explains why this case should never happen now.

src/model.rs
woshilapin: Good point. Done in eaef491376a016ac4aec5be7d4f99b3116174621.

src/model.rs
```diff
@@ -989,6 +989,41 @@ impl Collections {
         self.stop_areas = CollectionWithId::new(updated_stop_areas).unwrap();
     }
 
+    /// Check that all references to geometries actually points towards existing
+    /// `Geometry`. This is a common problem where a `geometries.txt` is read in
+    /// NTFS, a line of this file is not a valid WKT format, then the `Geometry`
+    /// is not created. However, the object that references this `Geometry` will
+    /// very likely be created (for example, a `Route` from `routes.txt`). This
+    /// creates a incoherent model where a `Route` points to a `Geometry` which
```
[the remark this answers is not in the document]
woshilapin: Done in 282c36e0bfe4a02de88a1e6493340c83cb49817d.

tests/fixtures/apply_rules/output_report/report_apply_property.json
woshilapin: I don't think it is dead code. The `geo_id` is given by an end-user when configuring an `apply-rules` process. So it is still possible to end up in this code with an incorrect ID I believe.

datanel: This case in the link above means
an object has a geometry_id and there is no geometries in the collection of geometries with id = geometry_id
but this case never happen now with your change.

If I'm wrong you should write the test for this case

woshilapin: I pushed a [unit test](https://github.com/CanalTP/transit_model/pull/543/files#diff-e3d3b1c0a718fe3ab0336f3ef26df712R1080-R1111) for testing this case.

In term of public API, we indeed accept a `Model` so the case you're mentioning should never happen. But there is also internal API, like `crate::apply_rules::property_rules::update_geometry()`, or `crate::apply_rules::property_rules::apply_rules()` (which accept a `Collections`, not a `Model`) that a future developer might use at some point.

ArnaudOggy:
@Mergifyio refresh

PR #550, opened 2020-02-21
Doc ntfs2gtfs: Update location_type according to the code

papailio:


src/documentation/ntfs_to_gtfs_specs.md
```diff
@@ -82,7 +82,7 @@ stop_name | yes | stops.txt | stop_name
 stop_lat | yes | stops.txt | stop_lat
 stop_lon | yes | stops.txt | stop_lon
 zone_id | no | stops.txt | fare_zone_id
-location_type | no | stops.txt | location_type
+location_type | no | stops.txt | location_type | The value is set to `0` if the input value is `0` or invalid or unspecified, `1` if the input value is `1`, `2` if the input value is `3`, `3` if the input value is `4` and `4` if the input value is `5`.
```
woshilapin: What about if the input value is `2`?

woshilapin: Some more information. From the current implementation, this is what is done:
- Remove all `stop`s of type `2`
- In all `trip`s, remove all `stop_time` whose `parent_station` was of type `2`
- Remove `trip` that don't contains anymore `stop_time`

Apart from the fact that this choice of removing only parts of a `trip` is weird, is very possible we also end up with `route`s that don't contain any `trip` anymore.

prhod: @woshilapin you're right about this "need for sanitize", we'll do it soon.

PR #553, opened 2020-02-26
[tech] Mergify: add condition on github actions

datanel:


woshilapin:
Since you're modifying the mergify rules, we talked a lot already about getting back to a ONE approve instead of TWO. Should we do it now?

datanel:
the ONE approve was not just for NetexFrance?

PR #558, opened 2020-02-27
[tech] Add code coverage in the build

woshilapin:


woshilapin:
I completely rewrote the code coverage publishing:
- used `cargo-tarpaulin` instead of `grcov`
- used CodeCov instead of Coveralls

woshilapin:
I updated the PR with a flat-styl-ish badge and removed the `do_not_merge` tag (I didn't realize it was still here).

woshilapin:
Hmm, that's annoying. So `tarpaulin` is failing with a segfault and I've been able to reproduce locally... but this is not deterministic.  Launching the same command over and over finally get through locally. I'm not sure how to deal with this. Any idea?

woshilapin:
I've reported an issue on `tarpaulin` project.
https://github.com/xd009642/tarpaulin/issues/461

woshilapin:
I updated the PR and it seems to work fine.

woshilapin:
This ticket adds a Github Action task to produce the Code Coverage and send it to CodeCov. This means we now have some visibility on code coverage. CodeCov will even produce a comment on each PR to show the evolution of coverage (see [Kirin repository](https://github.com/CanalTP/kirin/pull/317) for example).

pbougue (approved):
OK, we will also get a shield on top of readme.
And it should be quite easy to use codecov to spot where tests are missing.

PR #567, opened 2020-03-06
[feature] NeTEx France - Add entrances/exits in arrets.xml

ArnaudOggy:


src/netex_france/stops.rs
```diff
@@ -389,6 +396,51 @@ impl<'a> StopExporter<'a> {
...
 
+    fn generate_entrances(&self, stop_area_id: &'a str) -> Result<Option<Element>> {
+        let stop_place_entrances = self
+            .model
+            .stop_locations
+            .values()
+            .filter(|sl| sl.parent_id == Some(stop_area_id.to_string()))
+            .filter(|sl| sl.stop_type == StopType::StopEntrance)
+            .map(|sl| self.generate_stop_place_entrance(sl))
+            .collect::<Result<Vec<Element>>>()?;
```
woshilapin: This is filtering over all `stop_location` for every `stop_area`. Maybe we can avoid all this processing? You could create a new field `stop_entrances` in `StopExporter`:
```rust
stop_entrances: HashMap<&'a str, Vec<&'a str>>,
```

This new field would store information once and for all at construction time of `StopExporter` about all `stop_entrance` per `stop_area` (key is a `stop_area` identifier, value is a list of `stop_entrance` identifiers).

PR #571, opened 2020-03-13
[tech] Fix some new clippy errors

woshilapin:


woshilapin:
> I wonder if we always must support the last version of Rust or we can have a minimum version as it is done in mimirsbrunn [CanalTP/mimirsbrunn:.travis.yml@`master`#L4](https://github.com/CanalTP/mimirsbrunn/blob/master/.travis.yml#L4)

That wouldn't change our need to make these kind of PR to update the code to the latest stable. But yes, that's something we could think about.

PR #576, opened 2020-03-16
[feature] NeTEx France - Add TransportMode on Line section

ArnaudOggy:


src/netex_france/lines.rs
```diff
@@ -39,11 +50,45 @@ impl<'a> LineExporter<'a> {
...
+                    .map(move |netex_mode| (vehicle_journey, netex_mode))
+            })
+            .map(|(vehicle_journey, netex_mode)| {
+                let line_id = model
+                    .routes
+                    .get(&vehicle_journey.route_id)
+                    .map(|route| &route.line_id)
+                    .unwrap();
+                (line_id, netex_mode)
+            })
```
woshilapin: To avoid the `.unwrap()` (which should be OK but who knows...) you can write something like that I believe.

```suggestion
            .filter_map(|vehicle_journey| {                                        
                NetexMode::from_physical_mode_id(&vehicle_journey.physical_mode_id)
                    .map(move |netex_mode| (vehicle_journey, netex_mode))          
                    .and_then(|(vehicle_journey, netex_mode)| {                    
                        model                                                      
                            .routes                                                
                            .get(&vehicle_journey.route_id)                        
                            .map(|route| &route.line_id)                           
                            .map(|line_id| (line_id, netex_mode))                  
                    })                                                             
            })                                                                     
```
Note however that if no `line` is found for this `vehicle_journey`, then it's just ignored without a log message. Might be able to make even better?

ArnaudOggy: Ok for me. Done in https://github.com/CanalTP/transit_model/pull/576/commits/773f4d9db0805cab9adca8d07b1493bb2a018462.
We are after sanitize, therefore a vj necessarily has a line.

PR #579, opened 2020-03-19
[feature] Compact identifiers

woshilapin:


src/netex_idf/lines.rs
```diff
@@ -188,8 +214,15 @@ fn make_lines(lines_netex_idf: &CollectionWithId<LineNetexIDF>) -> Result<Collec
...
-            .unwrap_or_else(BTreeSet::new);
+            .map(|pc| {
+                vec![
+                    ("source".into(), ln.source_code.clone()),
+                    ("Netex_PrivateCode".into(), pc),
+                ]
+                .into_iter()
+                .collect()
+            })
+            .unwrap_or_else(KeysValues::default);
```
ArnaudOggy: You cannot make the insertion of source_code (necessarily existing) dependent by private_code (optional)

woshilapin: Nice catch indeed. Something like the following should be more correct I believe.

```rust
let mut codes = KeysValues::default();                       
codes.insert(("source".into(), ln.source_code.clone()));     
if let Some(private_code) = ln.private_code.clone() {        
    codes.insert(("Netex_PrivateCode".into(), private_code));
}                                                            
```

woshilapin: Done in 023fca7b89a26e414ed850d20752eba0fc0513eb.

ArnaudOggy:
offers.rs / line 504
```
let codes = vec![(
    String::from("Netex_ServiceJourneyPattern"),
    journey_pattern_ref.clone(),
)]
.into_iter()
.collect::<BTreeSet<_>>();
let mut route = route.clone();
route.codes.extend(codes);
```
otherwise you lose the source (initial id) of your route

woshilapin:
> offers.rs / line 504
> 
> ```
> let codes = vec![(
>     String::from("Netex_ServiceJourneyPattern"),
>     journey_pattern_ref.clone(),
> )]
> .into_iter()
> .collect::<BTreeSet<_>>();
> let mut route = route.clone();
> route.codes.extend(codes);
> ```
> 
> otherwise you lose the source (initial id) of your route

Hmm, indeed I missed the `object_code` for routes... Thanks. Done in 9dd08ec6383b6bb255a04de78508fe216e0fb734.

PR #580, opened 2020-03-19
[tech] Use 'skip_error' crate instead of custom macro

woshilapin:


woshilapin:
I'm putting `do_not_merge` since there is discussion about a change of name for the crate `skip-fail`.

woshilapin:
> I'm putting `do_not_merge` since there is discussion about a change of name for the crate `skip-fail`.

Waiting for the release of the crate [`skip_error`](https://github.com/CanalTP/skip_error).

PR #588, opened 2020-03-24
remove collection and use typed_index_collection

datanel:


Cargo.toml
```diff
@@ -95,6 +93,6 @@ name = "write_netex_france"
 path = "tests/write_netex_france.rs"
 required-features = ["proj"]
 
-[dev-dependencies]
+[dev_dependencies]
```
woshilapin: Hmm, official documentation mentioned with a dash `-`.

https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#development-dependencies

src/gtfs/write.rs
```diff
@@ -537,7 +537,7 @@ mod tests {
     use pretty_assertions::assert_eq;
     use std::{collections::BTreeSet, fs::File, io::Read};
     use tempfile::tempdir;
-    use transit_model_collection::CollectionWithId;
+    use typed_index_collection::CollectionWithId;
```
woshilapin: This line can probably be removed? It's also at the top of the file.

tests/merge_ntfs.rs
```diff
@@ -23,20 +23,22 @@ use transit_model::objects::{Comment, CommentLinks, StopPoint, VehicleJourney};
 use transit_model::test_utils::*;
 use transit_model::transfers;
 use transit_model::transfers::TransfersMode;
-use transit_model_collection::CollectionWithId;
-use transit_model_collection::Idx;
+use typed_index_collection::CollectionWithId;
+use typed_index_collection::Idx;
```
woshilapin: Maybe they can be merged?
```suggestion
use typed_index_collection::{CollectionWithId, Idx};
```

Cargo.toml
datanel: I know, I didn't change that, weird

tests/merge_ntfs.rs
datanel: maybe :)

PR #590, opened 2020-03-26
[tech] NeTEx France - Fix AccessibilityAssessment uniqueness

ArnaudOggy:
Fix https://github.com/CanalTP/tartare-tools/issues/59

### DIFF
--- a/src/netex_france/stops.rs
+++ b/src/netex_france/stops.rs
@@ -192,13 +192,12 @@ impl<'a> StopExporter<'a> {
                 element_builder
             };
 
-        let element_builder = if let Some(accessibility_element) =
-            self.generate_quay_accessibility(&stop_point.id, stop_point.equipment_id.as_ref())
-        {
-            element_builder.append(accessibility_element)
-        } else {
-            element_builder
-        };
+        let element_builder =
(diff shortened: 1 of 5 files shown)


src/netex_france/stops.rs
```diff
@@ -350,14 +350,21 @@ impl<'a> StopExporter<'a> {
         None
     }
 
-    fn generate_quay_accessibility(&self, equipment_id: Option<&'a String>) -> Option<Element> {
+    fn generate_quay_accessibility(
+        &self,
+        stop_point_id: &'a str,
+        equipment_id: Option<&'a String>,
+    ) -> Option<Element> {
```
woshilapin: Send the reference to whole `StopPoint` instead since you need 2 fields out of it...
```suggestion
    fn generate_quay_accessibility(
        &self,
        stop_point: &'a StopPoint,
    ) -> Option<Element> {
```

ArnaudOggy: Ah, well seen! I missed the idea :+1:

ArnaudOggy: Fix in d95424d438267811d3f4b3185f71f22e924d90f0

PR #594, opened 2020-03-27
[tech] Remove TravisCI

woshilapin:
We've been using Github Actions for a while now with no major problem. How about we remove TravisCI which seems to have some problems lately?


woshilapin:
> don't forget to remove the badge in `README.md`

Indeed! Done in c529731fe9638eecf81927bb739ac4dfef3fc29d.

PR #597, opened 2020-03-31
[feature] add filter NTFS by line id

patochectp:


patochectp:
> 
> 
> It would be nice to test this new feature. I think you can take inspiration on [this test](https://github.com/CanalTP/transit_model/blob/master/tests/filter_ntfs.rs#L72-L88).
> 
> Or maybe even better, since we probably don't want an integration test for each new property (which is expansive in compilation and execution time), let's introduce some unit tests directly in the file `src/filter.rs`, to test the functions only? I know they don't exist at the moment but that seems the right level to test them.

PR #604, opened 2020-04-16
remove some clippy warnings

datanel:


src/model.rs
```diff
@@ -1299,7 +1298,7 @@ mod tests {
                 .unwrap();
             collections.try_merge(collections_to_merge).unwrap();
             let bus_mode = collections.physical_modes.get(BUS_PHYSICAL_MODE).unwrap();
-            assert_eq!(42f32, bus_mode.co2_emission.unwrap());
+            assert!((bus_mode.co2_emission.unwrap() - 42f32).abs() < std::f32::EPSILON);
```
pbougue, line 1301: Can't you use `relactive_eq!` like object.rs here ?
(same for all `EPSILON` occurences I guess)

datanel: yes, I did that before using `approx`  for the test I was blocked on but I didn't have the courage to change that

.github/workflows/ci.yml
```diff
@@ -43,7 +43,7 @@ jobs:
       uses: actions-rs/cargo@v1
       with:
         command: clippy
-        args: --workspace --all-features -- -D warnings
+        args: --workspace --all-features --lib --bins --tests --examples -- -D warnings
```
woshilapin, line 46: `--all-targets` would do the same if I understand correctly what we assume about these options (except for `--benches` but you could just remove them as part of the PR since we decided to move forward with this or just leave them since you seem to have fixed them).

benches/read_kv1.rs
```diff
@@ -21,5 +21,7 @@ use transit_model::kv1;
 
 #[bench]
 fn read_kv1(bencher: &mut Bencher) {
-    bencher.iter(|| kv1::read_from_path("./tests/fixtures/kv12ntfs/input", None, None).unwrap());
+    bencher.iter(|| {
+        kv1::read_from_path("./tests/fixtures/kv12ntfs/input", None::<&str>, None).unwrap()
```
woshilapin, line 25: Interesting, I would have thought about something like `Option::<&str>::None` but didn't know about your more concise solution.

src/objects.rs
```diff
@@ -1848,51 +1849,28 @@ mod tests {
...
+        lon: 2.374_377,
+        lat: 48.844_304,
     };
 
     #[test]
     fn orthodromic_distance() {
-        assert!(nearly_equal(COORD1.distance_to(&COORD1), 0.0, TOLERANCE));
-        assert!(nearly_equal(COORD1.distance_to(&COORD2), 357.64, TOLERANCE));
-        assert!(nearly_equal(COORD2.distance_to(&COORD1), 357.64, TOLERANCE));
+        relative_eq!(COORD1.distance_to(&COORD1), 0.0);
```
woshilapin, line 1865: I believe it should be `assert_relative_eq!` because `relative_eq!` only returns a `bool` (which will never fail).

src/model.rs
datanel: done

benches/read_kv1.rs
datanel: yes it's not natural

src/objects.rs
datanel: done

src/model.rs
```diff
@@ -1807,17 +1806,17 @@ mod tests {
             let mut collections = collections(3);
             collections.update_stop_area_coords();
             let stop_area = collections.stop_areas.get("stop_area:1").unwrap();
-            assert_eq!(2.0, stop_area.coord.lon);
-            assert_eq!(2.0, stop_area.coord.lat);
+            assert!((stop_area.coord.lon - 2.0).abs() < std::f64::EPSILON);
```
woshilapin, line 1809: Not using `assert_relative_eq!` on these one, any reason?

src/objects.rs
woshilapin: You might have missed one, see https://github.com/CanalTP/transit_model/pull/604#discussion_r409545688.

src/model.rs
datanel: bad find and replace by regex . Thx

.github/workflows/ci.yml
datanel: benches removed and `--lib --bins --tests --examples` replaced by `-all-targets`

PR #616, opened 2020-04-24
[lint] Fix some clippy warnings on Cargo manifests

woshilapin:
`clippy::cargo` is not enabled by default but can provide some useful hints on  `Cargo.toml` manifests. However, there is no easy way to solve the problem of multiple versions of dependencies so I specifically deactivated this one.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,7 +43,7 @@ jobs:
       uses: actions-rs/cargo@v1
       with:
         command: clippy
-        args: --workspace --all-features --all-targets -- -D warnings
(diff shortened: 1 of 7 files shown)


woshilapin:
> Now that you applied all the recommandations from `cargo_common_metadata`, is it useful to always check `clippy::cargo` ?

I would say yes, for the same reasons that we didn't remove `clippy` in general even if we already fixed all the warnings and errors (I agree changes on `Cargo.toml` are much less frequent but still).

> May be if we add a new "sub-crate". I don't see well the benefice.

A few reasons to keep it, in my opinion:
- new developers might come in, not necessarily aware of all our practices
- new sub-crates might be created
- new `clippy::cargo` rules might appears (today, only 3 but more might be created)

PR #617, opened 2020-04-28
[tech] Add Dockerfile in TransitModel

ArnaudOggy:


.dockerignore
```diff
@@ -0,0 +1,5 @@
+./examples
+./target
+LICENSE
```
woshilapin, line 3: Might need to keep the `LICENSE` file in the packaging...

Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+RUN echo "deb [arch=amd64] https://kisiodigital.jfrog.io/kisiodigital/debian-local stretch main" > /etc/apt/sources.list.d/kisio-digital.list
+RUN apt update && apt install --yes pkg-config libssl-dev clang proj=${PROJ_VERSION}
+
+COPY gtfs2netexfr gtfs2netexfr
+COPY gtfs2ntfs gtfs2ntfs
+COPY ntfs2gtfs ntfs2gtfs
+COPY ntfs2netexfr ntfs2netexfr
+COPY ntfs2ntfs ntfs2ntfs
+COPY Cargo.toml .
+COPY src src
```
woshilapin, line 19: Why not `COPY . .`? I think because you defined the `.dockerignore`, it will copy only what's not ignored (see https://docs.docker.com/engine/reference/builder/#dockerignore-file).

Dockerfile
```diff
@@ -0,0 +1,40 @@
+ARG PROJ_VERSION="6.3.0"
+
+FROM rust:1-slim-stretch as builder
+ARG PROJ_VERSION
+ENV PROJ_DEB "proj_${PROJ_VERSION}_amd64.deb"
+ENV GPG_KEY "C60D758F807A525534C5DFD57B639E3638A8112A"
+WORKDIR /usr/src/app
```
woshilapin, line 7: How about moving this line below, just above the `COPY` operation since they're related.

.dockerignore
ArnaudOggy: Ok. Done in https://github.com/CanalTP/transit_model/pull/617/commits/91c57e7a5d2e985469bf019544df22dbfc32a174

Dockerfile
ArnaudOggy: Ok. Besides, it was done like this previously.

woshilapin: Yes, I know, incremental improvements ;-) Whoever did that previously was improving something else at the time I guess!

PR #618, opened 2020-04-28
[feature] add odt comment

patochectp:


patochectp:
> You can already prepare the reception normally
> 
> ```
> /// OnDemandTransport comment
> #[structopt(long)]
> odt_comment: Option<String>,
> ```
> 
> In gtfs2ntfs/src/main.rs
> and gtfs2netexfr/src/main.rs

Done

src/gtfs/mod.rs
```diff
@@ -261,13 +261,13 @@ fn read<H>(
     config_path: Option<impl AsRef<Path>>,
     prefix: Option<String>,
     on_demand_transport: bool,
+    on_demand_transport_comment: Option<String>,
```
woshilapin, line 264: Would it be possible to regroup all this parameters into on structure to avoid getting more and more parameters to this function?

Maybe something like (not tested)

```rust
struct Configuration<P: AsRef<Path>> {
  config_file: P,
  prefix: Option<String>,
  on_demand_transport: bool,
  on_demand_transport_comment: Option<String>,
}
```

src/gtfs/read.rs
```diff
@@ -584,6 +604,34 @@ fn manage_comment_from_stop(
     comment_links
 }
 
+fn manage_odt_comment_from_stop_time(
+    comments: &mut CollectionWithId<objects::Comment>,
+    companies: &mut CollectionWithId<objects::Company>,
+    on_demand_transport_comment: Option<String>,
```
woshilapin, line 610: This whole function is useless to call if this parameter is `None`. It feels that this function should be conditionally called. Therefore, this parameter should be `String`, not `Option<String>`. Same comment could be made for companies. I'm wondering if this function should be more like this.

```rust
fn manage_odt_comment_from_stop_time(
    comments: &mut CollectionWithId<objects::Comment>,
    on_demand_transport_comment: String,
    company: &Company,
) -> Option<String> {
```

src/gtfs/read.rs
```diff
@@ -584,6 +604,34 @@ fn manage_comment_from_stop(
...
+    company_id: &str,
+) -> Option<String> {
+    if let Some(message) = on_demand_transport_comment {
+        if let Some(company) = companies.get(company_id) {
+            let comment_id = "ODT:".to_string() + &company.id;
+            let comment = comments.get(&comment_id);
+            if comment.is_none() {
```
woshilapin, line 617: `comment` variable is not used after. You could squash these 2 lines.
```suggestion
            if comments.get(&comment_id).is_none() {
```

src/gtfs/read.rs
```diff
@@ -584,6 +604,34 @@ fn manage_comment_from_stop(
...
 
+fn manage_odt_comment_from_stop_time(
+    comments: &mut CollectionWithId<objects::Comment>,
+    companies: &mut CollectionWithId<objects::Company>,
+    on_demand_transport_comment: Option<String>,
+    company_id: &str,
+) -> Option<String> {
+    if let Some(message) = on_demand_transport_comment {
+        if let Some(company) = companies.get(company_id) {
+            let comment_id = "ODT:".to_string() + &company.id;
```
woshilapin, line 615: Maybe using `format!` would be more idiomatic.
```suggestion
            let comment_id = `format!("ODT:{}", company.id);
```

src/gtfs/read.rs
```diff
@@ -441,7 +441,27 @@ where
...
+                        &stop_time.stop_sequence.to_string()
+                    );
+                    let comment_idx = collections.comments.get_idx(&comment_id).unwrap();
+                    collections
+                        .stop_time_ids
+                        .insert((vj_idx, stop_time.stop_sequence), stop_time_id);
+                    collections
+                        .stop_time_comments
+                        .insert((vj_idx, stop_time.stop_sequence), comment_idx);
+                }
```
woshilapin, line 463: I believe this whole block should be part of the `manage_odt_comment_from_stop_time` function for multiple reasons:
- first of all, this `manage_stop_times` function is already pretty big
- the current `manage_odt_comment_from_stop_time` function is creating a `Comment` where you could get back directly the `Idx<>`. But instead, the function push the `Comment` on the `CollectionWithId` and then only after, you search again for it with a `get_idx`.

src/gtfs/read.rs
```diff
@@ -584,6 +604,34 @@ fn manage_comment_from_stop(
...
+                    comment_type: objects::CommentType::OnDemandTransport,
+                    label: None,
+                    name: message
+                        .replace("{agency_name}", &company.name)
+                        .replace("{agency_phone}", &company.phone.clone().unwrap_or_default()),
+                    url: None,
+                };
+                comments.push(comment).unwrap();
+            }
+            return Some(comment_id);
```
woshilapin, line 629: You never need the `comment_id` (`String`), what you really need is the `Idx<Comment>`. So I believe you could modify this function to return `Option<Idx<Comment>>`. And 2 lines above, instead of doing `comments.push(comment).unwrap()` gives you the `Idx<Comment>` you're looking for.

src/gtfs/mod.rs
ArnaudOggy: clippy limit for Result (3 or 4 max), but not for inputs :thinking:

woshilapin: For function, the limit for `clippy` seems to be 7
https://rust-lang.github.io/rust-clippy/master/index.html#too_many_arguments

For returned values, it's always been one (you can't return more). But I think you're talking about big tuples in `Result`? This must have been a different rule about type complexity
https://rust-lang.github.io/rust-clippy/master/index.html#type_complexity

ArnaudOggy: Limit is 3 for tuple in ```Result``` yes

src/gtfs/read.rs
patochectp: For this purpose, i need to 2 more parameters for manage_odt_comment_from_stop_time (collections.stop_time_ids and collections.stop_time_comment) and lots of parameters are considered bad style and reduce readability (Clippy Lints)

woshilapin: If you need multiple fields out of `collections`, then why not sending the an entire mutable reference on `collections`?

patochectp: collections.vehicle_journeys is borrowing before manage_odt_comment_from_stop_time (collections.vehicle_journeys.index_mut(vj_idx).

src/gtfs/mod.rs
```diff
@@ -256,22 +256,30 @@ struct Shape {
...
+    pub config_path: Option<P>,
+    /// used to prefix objects
+    pub prefix: Option<String>,
+    /// stop time precision management
+    pub on_demand_transport: bool,
+    /// on demand transport comment template
+    pub on_demand_transport_comment: Option<String>,
+}
+
+fn read<H>(file_handler: &mut H, configuration: Configuration<impl AsRef<Path>>) -> Result<Model>
```
woshilapin, line 272: Interesting, any specific reason that made you choose
```
fn read<H>(file_handler: &mut H, configuration: Configuration<impl AsRef<Path>>) -> Result<Model>
```

over something like
```
fn read<H, P: AsRef<Path>>(file_handler: &mut H, configuration: Configuration<P>) -> Result<Model>
```

datanel: apparently it's the same with a small difference https://stackoverflow.com/a/47515540

woshilapin: Excellent link, thank you.

patochectp:
> Two minor comments, apart from that, nice refactorization.

Thanks for your help !

PR #620, opened 2020-05-04
[tech] minor refactoring

patochectp:


src/gtfs/mod.rs
```diff
@@ -269,7 +269,7 @@ pub struct Configuration<P: AsRef<Path>> {
     pub on_demand_transport_comment: Option<String>,
 }
 
-fn read<H>(file_handler: &mut H, configuration: Configuration<impl AsRef<Path>>) -> Result<Model>
+fn read<H, P: AsRef<Path>>(file_handler: &mut H, configuration: Configuration<P>) -> Result<Model>
```
woshilapin, line 272: I was pretty happy with your proposition of using `impl AsRef<Path>` actually. The only difference between the 2 propositions being that one is giving a name to the generic (`P` here). This could be useful if you have to use it somewhere else but in this case, you don't. So I find `impl AsRef<Path>` simpler because it avoid the complexity of a new variable name. But whichever you choose to keep, I'm fine with it.

PR #640, opened 2020-06-01
[feature] Expose Report and ReportType

woshilapin:
For moving some part of `transit_model` like `apply_rules` out of `transit_model`, we need to expose `Report` and `ReportType`.  However, some variant enums of `ReportType` are really related to `apply_rules` itself, not to `transit_model`. So I changed `ReportType` into a trait/contract/interface instead of an `enum`. User needs to implement a type that conforms to this contract. I also moved `Report` and `ReportType` into their own module `report`.

See https://github.com/CanalTP/tartare-tools/pull/77 for how this new trait will be used.  In the current PR, one implementation of the new trait `ReportType` is `TransitModelReportType` which is nothing else than the old `enum ReportType`.

### DIFF
--- a/src/report.rs
+++ b/src/report.rs
@@ -1,0 +1,96 @@
+// Copyright (C) 2017 Kisio Digital and/or its affiliates.
+//
+// This program is free software: you can redistribute it and/or modify it
+// under the terms of the GNU Affero General Public License as published by the
+// Free Software Foundation, version 3.
+
+// This program is distributed in the hope that it will be useful, but WITHOUT
+// ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+// FOR A PARTICULAR PURPOSE. See the GNU Affero General Public License for more
+// details.
+
+// You should have received a copy of the GNU Affero General Public License
+// along with this program. If not, see <https://www.gnu.org/licenses/>
+
+//! Helpers to create a report for faillible processes.
+use serde::Serialize;
+
+/// Each report record will be categorized with a type implementing this
+/// `ReportCategory` trait.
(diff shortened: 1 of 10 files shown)


src/report.rs
```diff
@@ -0,0 +1,97 @@
...
+/// and errors.
+#[derive(Debug, Default, Serialize)]
+pub struct Report<R: ReportType> {
+    errors: Vec<ReportRow<R>>,
+    warnings: Vec<ReportRow<R>>,
+}
+
+impl<R: ReportType> Report<R> {
+    /// Creates an empty new report.
+    pub fn new() -> Self {
```
woshilapin, line 68: Why not implementing `Default` manually instead?

src/report.rs
```diff
@@ -0,0 +1,97 @@
...
+
+// You should have received a copy of the GNU Affero General Public License
+// along with this program. If not, see <https://www.gnu.org/licenses/>
+
+//! Helpers to create a report for faillible processes.
+use serde::Serialize;
+
+/// Each report record will be categorized with a type implementing this
+/// `ReportType` trait.
+pub trait ReportType: Serialize + PartialEq {}
```
woshilapin, line 20: Name the trait with a verb instead of a name.

At the very least `ReportCategory` is better (`Type` is connoted in Rust).

src/report.rs
woshilapin: Done in dac6b6a2f69ba1bed26f8556507f38598188b6bf.

src/report.rs
woshilapin: Done in 7e0410ef902f40d3d1c835bf163b1b0d8609ca57.

PR #645, opened 2020-06-10
[tech] remove netexidf reader

patochectp:


src/lib.rs
```diff
@@ -33,8 +33,8 @@ pub mod model;
 #[cfg(feature = "proj")]
 pub mod netex_france;
 #[cfg(feature = "proj")]
-pub mod netex_idf;
-mod netex_utils;
+#[doc(hidden)]
```
woshilapin, line 36: If we expose, I think we must document. I wouldn't add this clause.

Issue #646, opened 2020-06-10
gtfs2netexfr installation issue


jvanoost:
Hi,

I'm a great newb with rust and cargo
I followed the install procedure by install PROJ and then clang and libssl-dev => that's OK

I tried to install gtfs2netexfr and get this when I lauched the command
```
~/home/gtfs2netexfr$ cargo install gtfs2netexfr
    Updating crates.io index
error: could not find `gtfs2netexfr` in registry `https://github.com/rust-lang/crates.io-index`

```

how can I solve this ?
Thanks

datanel:
@jvanoost is it good for you now?

jvanoost:
not yet, but my fault

I made a mistake while wanted to add proj to the path.
I'm on it to correct
But all you give me is ok, so I'm pretty sure it will be ok ;)

datanel:
sorry I'm not sure to understand well.

Your question is how to test the command with your GTFS Lille or how to test the netex produced ?

jvanoost:
Sorry it wasn't clear
the question was about testing the netex produced.
I tried to zip and upload it on : http://appli.chouette.mobi/ but it got an error

datanel:
I don't know what tool you can test with but you can test that the XML produced has a valid schema.
see https://github.com/CanalTP/transit_model/blob/master/tests/write_netex_france.rs#L79

with xmllint
`xmllint --noout --nonet --huge --schema NeTEx_publication.xsd your_produced_netex`

NetEx_publication.xsd coming from https://github.com/NeTEx-CEN/NeTEx/blob/348033a470ae7cfc50c57c911b0e786b52dd85bc/xsd/NeTEx_publication.xsd

jvanoost:
Ok thanks a lot ! I close the issue !

PR #648, opened 2020-06-11
[doc] fix the installation of binairies

datanel:
and fixed some warnings/errors form my markdown linter (https://github.com/DavidAnson/markdownlint/blob/v0.20.3/doc/Rules.md#rules)

fix https://github.com/CanalTP/transit_model/issues/646

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 authors = ["Kisio Digital <team.coretools@kisio.com>", "Guillaume Pinot <texitoi@texitoi.eu>"]
 name = "transit_model"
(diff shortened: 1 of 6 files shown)


Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 authors = ["Kisio Digital <team.coretools@kisio.com>", "Guillaume Pinot <texitoi@texitoi.eu>"]
 name = "transit_model"
-version = "0.18.1"
```
woshilapin, line 4: Not necessarily useful for documentation but why not.

PR #650, opened 2020-06-12
[tech] remove binary filter-ntfs

patochectp:


src/ntfs/mod.rs
```diff
@@ -16,7 +16,6 @@
 //! format management.
 
 #![allow(missing_docs)]
-pub mod filter;
```
woshilapin, line 19: If you remove this line, please remove the line above with it.

PR #656, opened 2020-06-22
[feature] gtfs2ntfs and ntfs2ntfs: generates transfers

datanel:


ntfs2ntfs/src/main.rs
```diff
@@ -41,6 +41,19 @@ struct Opt {
         default_value = &transit_model::CURRENT_DATETIME
     )]
     current_datetime: DateTime<FixedOffset>,
+
+    // The max distance in meters to compute the tranfer
+    #[structopt(long, short = "d", default_value = "500")]
```
ArnaudOggy, line 46: This letter being taken for 2 other processes on the TT side we may change these latter processes (to harmonize, and keep "d", "s", "t") :wink:

datanel: processes that need transfers? I'm not sure

ArnaudOggy: Effectively. I just looked at those who had ntfs out :+1:

gtfs2ntfs/src/main.rs
```diff
@@ -58,6 +58,19 @@ struct Opt {
         default_value = &transit_model::CURRENT_DATETIME
     )]
     current_datetime: DateTime<FixedOffset>,
+
+    // The max distance in meters to compute the tranfer
+    #[structopt(long, short = "d", default_value = "500")]
```
ArnaudOggy, line 63: It will be easier with constants in lib.rs (like & transit_model::CURRENT_DATETIME) just above. 
We can use them in TT too.

datanel: done

PR #658, opened 2020-06-22
[doc] move internal doc files and links

pbougue:
* move `src/documentation/` -> `documentation/` (debatable)
* rename `common.md` -> `common_ntfs_specs.md` (and change links)
* add links into binaries readmes to relevant specs

The final format of binaries readme will change and be enriched later with `proj` links (starting with `gtfs2netexfr`).

- [ ] Squash to be done on merge.


documentation/gtfs_to_ntfs_specs.md
```diff
@@ -9,14 +9,14 @@ specification, the conversion should stop immediately with an error, unless
 otherwise specified.
 
 At the end of the conversion, a sanitizing operation is started on the final
-model. See [common.md](common.md) for more information.
+model. See [common NTFS specifications] for more information.
```
woshilapin, line 12: This is not really "common NTFS specifications" since _specifications_ are on [another Github repository](https://github.com/CanalTP/ntfs-specification/). This is more automatic processing and rules we apply every time we're producing an NTFS: sanitizing dangling ID, reconstructing missing route names, recalculating stop area geolocation, etc. Maybe "common NTFS operations", or "common NTFS rules" would be more appropriated.

gtfs2netexfr/README.md
```diff
@@ -40,3 +40,13 @@ gtfs2netexfr --input /path/to/gtfs/folder/ --output /path/to/netexfr/ --particip
   export; it is exported in each NeTEx file
 
 Get more information about the available options with `--help`.
+
+## Specifications
+
+As NTFS is the center format for conversion, [common NTFS specifications] is useful.
```
woshilapin, line 46:
```suggestion
As NTFS is the base format for conversion, [common NTFS specifications] is useful.
```
or
```suggestion
As NTFS is the central format for conversion, [common NTFS specifications] is useful.
```
or maybe even better
```suggestion
As NTFS is the pivot format for conversion, [common NTFS specifications] is useful.
```

documentation/gtfs_to_ntfs_specs.md
papailio: What about "common operations during conversion" (along with "common_conversion_operations.md") ?
I think that we should more highlight the fact that it is about general rules applied _when converting_ an input dataset into NTFS than NTFS itself.

woshilapin: By the way, is it always for conversion from a format into NTFS? For example, ˋ ˋrestrict-validity-period` is not such a conversion but is probably applying the rules too.

papailio: @pbougue spotted this too and you're right, it's not always a conversion. So, highlighting it might not be a good idea after all :)
I would go for "common NTFS rules".

gtfs2netexfr/README.md
pbougue: :heavy_check_mark: Done: "`pivot` format"

PR #661, opened 2020-06-25
[doc] no inline html in markdown

datanel:
- replaced `<br>` by `\` https://github.github.com/gfm/#example-658

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -57,9 +57,9 @@ Integration tests verify that the conversion produces files in accordance with
 the NeTEx specification.
 
 For that, tests are using the tool `xmllint` which can be installed on Debian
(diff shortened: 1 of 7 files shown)


.editorconfig
```diff
@@ -0,0 +1,13 @@
...
+[*]
+indent_style = space
+indent_size = 4
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+
+[*.md]
+trim_trailing_whitespace = false
```
pbougue, line 13: What we would like is to keep only the 2-trailing-whitespaces. Some editors do that but I don't remember which.
https://github.com/editorconfig/editorconfig/issues/391

datanel: editorconfig removed

PR #662, opened 2020-06-26
[tech] Sanitize before executing business rules

woshilapin:


src/model.rs
```diff
@@ -226,11 +226,6 @@ impl Collections {
         let mut level_id_used = HashSet::<String>::new();
         let mut calendars_used = HashSet::<String>::new();
 
-        // Keep fallback modes even if not referenced by the model
```
pbougue, line 229: Curious why you put that here first.
And also, why do you remove this now?

woshilapin: Yes but now that we run sanitize before the enhancement rules so they are now added after the sanitize. Therefore, we don't need specific code in the `sanitize()` to keep them.

PR #664, opened 2020-06-26
[tech] Add default direction_type (forward) to route

ArnaudOggy:
If a route doesn't have a `direction_type` (or empty),  the `direction_type` "forward" is assigned by default.

In order to lighten the workload, we will use a new generic method `enhance_route_directions`. 
Later it will disappear by deleting the Option on the `direction_type` field on `Route` structure
(which will require an update of the ntfs)

Ref. ND-945


src/model.rs
```diff
@@ -940,6 +940,15 @@ impl Collections {
         self.routes = CollectionWithId::new(routes).unwrap();
     }
 
+    /// If a route direction is empty, it's set by default with the "forward" value
+    pub fn enhance_route_directions(&mut self) {
+        let mut routes = self.routes.take();
+        for mut route in &mut routes.iter_mut().filter(|r| r.direction_type.is_none()) {
+            route.direction_type = Some(String::from("forward"));
+        }
+        self.routes = CollectionWithId::new(routes).unwrap();
```
woshilapin, line 949: I tried this new pattern to avoid to use `.take()`. It also allows you to avoid the `.unwrap()` below and I believe it saves you from doing 2 entire iteration on `routes` (the first one in the `for-loop` and the second one in `CollectionWithId::new()`).

```rust
let mut direction_types: BTreeMap<Idx<Route>, Option<String>> = BTreeMap::new();
for (route_idx, _) in self                                                      
    .routes                                                                     
    .iter()                                                                     
    .filter(|(_, r)| r.direction_type.is_none())                                
{                                                                               
    direction_types.insert(route_idx, Some(String::from("forward")));           
}                                                                               
for (route_idx, direction_type) in direction_types {
    self.routes.index_mut(route_idx).direction_type = direction_type;           
}                                                                               
```

In this code, there is also 2 loops but the second one doesn't iterate on all the routes, only the needed one. And you avoid an `.unwrap()`.

ArnaudOggy: Done in https://github.com/CanalTP/transit_model/pull/664/commits/a441d06f2d0d5fdce997b01e190173c81b6fb133
Thanks :+1:

PR #665, opened 2020-06-29
[doc] Improve contributing

pbougue:
* Consistency on how `clippy` and `test` are checked
* Word on our environments and tools

This work is debatable.
A future PR should attempt to use cargo (and maybe makefile) to minimize/centralize commands to type.

Ref. ND-930

PR #666, opened 2020-06-29
[doc] Improve documentation for binaries

woshilapin:
Ref. ND-929

### DIFF
--- a/gtfs2netexfr/src/main.rs
+++ b/gtfs2netexfr/src/main.rs
@@ -30,24 +30,19 @@ struct Opt {
     #[structopt(short, long, parse(from_os_str), default_value = ".")]
     input: PathBuf,
 
-    /// Output directory
+    /// Output directory.
     #[structopt(short, long, parse(from_os_str))]
     output: PathBuf,
 
-    /// Config file
-    #[structopt(
-        short,
-        long,
-        parse(from_os_str),
-        long_help = r"JSON file containing additional configuration.
-
-See https://github.com/CanalTP/transit_model/blob/master/documentation/common_ntfs_rules.md#configuration-of-each-converter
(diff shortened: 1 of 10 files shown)


gtfs2netexfr/src/main.rs
```diff
@@ -30,40 +30,41 @@ struct Opt {
     #[structopt(short, long, parse(from_os_str), default_value = ".")]
     input: PathBuf,
 
-    /// Output directory
+    /// Output directory.
```
datanel, line 33: Strange behaviour. All the dots are not displayed. Only if there are multilines.

woshilapin: Ah, indeed, I didn't notice and that is strange. Although, I would let them for at least consistency in the source file.

woshilapin:
> The point of this PR as I understood it was to have all sub-readme very similars.
> 
> Any diff with gtfs2netexfr/README.md should be just about specificities of the bin:
> 
>     * If using proj, the topo should be the same (no mention of the version, just in main readme)
> 
>     * adding/removing specific required params
> 
>     * same topo on the NeTEx manual-check (link will change with #665 : rebase!)
> 
>     * Maybe some other diffs I didn't spot for `restrict` or `ntfs2ntfs` ?
> 
> 
> So current work is good for me, but I think there is a little more changes to do on sub-readmes.

So, here is the last modifications I made:

- I remove the section about configuration in `gtfs2ntfs/README.md` (it is already covered by the `--help` option
- I removed the PROJ version from `ntfs2netexfr` (already mentioned in `CONTRIBUTING.md`)
- I added a link to the section `check the output manually` in `ntfs2netexfr` (and I rebased for that)

PR #667, opened 2020-06-30
[tech] Upgrade simple deps

pbougue:
Upgrade non-geo dependencies (those are tricky will come in later PR).

Also remove useless dep (tested `cargo udeps`).

TODO:

- [x] check that  projects depending on TM are OK
- [x] bump version > No need if merged before release.
- [x] squash (hence `do_not_merge`)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,13 +39,13 @@ xmllint = ["proj"]
 
 [dependencies]
 chrono = "0.4"
-csv = "1.0"
-derivative = "1"
+csv = "1"
+derivative = "2"
 failure = "0.1"
 geo = "0.12"
 geo-types = "0.4"
 iso4217 = "0.3"
-lazy_static = "1.2"
+lazy_static = "1"
 log = "0.4"
 md5 = "0.7"
 minidom = "0.12"
@@ -54,16 +54,15 @@ minidom_writer = "1"
 num-traits = "0.2"
 pretty_assertions = "0.6"
 proj = { version = "0.14", optional = true }
-quick-xml = "0.17"
+quick-xml = "0.18"
 relational_types = "1"
-rust_decimal = "1.0"
+rust_decimal = "1"
 serde = { version = "1", features = ["derive"] }
 serde_json = "1"
 skip_error = { version = "1", features = ["log"] }
 tempfile = "3"
-time-parse = "0.1"
 typed_index_collection = "1"
-walkdir = "2.1"
+walkdir = "2"
 wkt = "0.5"
 zip = "0.5"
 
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -21,6 +21,9 @@ exclude = [
 	"tests/",
 ]
 
+[package.metadata.cargo-udeps.ignore]
+normal = ["minidom_writer", "md5"]
+
```
datanel, line 26: we really want to use this?

woshilapin: I do agree with @datanel, we do not want/cannot update these dependencies yet but it might be possible at some point in the future and putting this section in `Cargo.toml` creates 2 problems for me:
- we acknowledge the use of yet another tool `cargo-udeps` (even if it's a useful tool, no need to make the project dependent on it)
- these dependencies might never be updated anymore (except if we remember to check these conditions each time)

Cargo.toml
```diff
@@ -39,13 +42,13 @@ xmllint = ["proj"]
 
 [dependencies]
 chrono = "0.4"
-csv = "1.0"
-derivative = "1"
+csv = "1.1"
+derivative = "2.1"
```
woshilapin, line 46: How about we put it like this?

```suggestion
csv = "1"
derivative = "2"
```

Cargo.toml
```diff
@@ -54,16 +57,15 @@ minidom_writer = "1"
 num-traits = "0.2"
 pretty_assertions = "0.6"
 proj = { version = "0.14", optional = true }
-quick-xml = "0.17"
+quick-xml = "0.18"
```
woshilapin, line 60: It feels to me that except for `quick-xml`, all the other should already be to the last version with a simple `cargo update` (see [Specifying Dependencies](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#caret-requirements)).

Cargo.toml
pbougue: There are 2 points here I think.
1. Little misunderstanding: this "ignore" is used to ignore warnings about "unused dependency", not about "not up-to-date".
2. Anyway, I also think we do not want to enforce the use of udeps (my idea was to check it every now and then, manually). So yes, we might as well remove those lines and just wait for tests to crash telling us what's a false-positive of udeps.

TODO if you agree: remove this cargo section.

Cargo.toml
pbougue: OK, so the way I see it : we just change the toml so that the behavior is explicit, like you suggested before (use only major versions above 1.x.x, minor for 0.x.x)

PR #671, opened 2020-07-09
[tech] [doc] Centralize fmt, clippy & test in makefile

pbougue:
Centralize in makefile and expose in doc and CI

Also:
* rm system package for install_proj (+ comments and cosmetic refactor)
* minor doc update after update on TT (https://github.com/CanalTP/tartare-tools/pull/103#discussion_r451521210)

TODO:
- [x] bump version > No need, "major" bump in another PR

Ref. ND-947

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,3 +16,24 @@ install_proj: ## Install PROJ and clang (requirements to use proj crate)
 	sudo make install
 	popd
 	rm -rf proj-$(PROJ_VERSION) proj-$(PROJ_VERSION).tar.gz
+
+fmt: format ## Check formatting of the code (alias for 'format')
+format: ## Check formatting of the code
+	cargo fmt --all -- --check
+
+clippy: lint ## Check quality of the code (alias for 'lint')
+lint: ## Check quality of the code
+	cargo clippy --workspace --all-features --all-targets -- --warn clippy::cargo --allow clippy::multiple_crate_versions --deny warnings
+
+test: ## Launch all tests
+	# Run all the tests of `transit_model` in the entire repository,
+	# activating all features (including `xmllint`), then without features
+	# to make sure that both work
+	cargo test --workspace --all-features
+	cargo test --workspace
+
+help: ## Print this help message
+	@grep -E '^[a-zA-Z_-]+:.*## .*$$' $(CURDIR)/$(firstword $(MAKEFILE_LIST)) | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}'
+
+.PHONY: install_proj fmt format clippy lint test help
+.DEFAULT_GOAL := help
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,18 @@
 PROJ_VERSION = 6.3.0
-
-install_proj:
+install_proj: ## Install PROJ and clang (requirements to use proj crate)
 	sudo apt update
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -1,8 +1,39 @@
...
+
+clippy: lint ## Check quality of the code (alias for 'lint')
+lint: ## Check quality of the code
+	cargo clippy --workspace --all-features --all-targets -- --warn clippy::cargo --allow clippy::multiple_crate_versions --deny warnings
+
+test: ## Launch all tests
+	# Run all the tests of `transit_model` in the entire repository,
+	# activating all features (including `xmllint`), then without features
+	# to make sure that both work
+	cargo test --workspace --all-features
```
ArnaudOggy, line 32: Wouldn’t a check be useful before launching tests (or clippy) to verify that proj is well installed (and in 6.3) ? :thinking:

pbougue: It's a bit outside the initial scope of the PR, but why not.

I wouldn't do it on clippy as someone may want to work without the proj feature (so avoid netex).

As for the tests, it won't work without proj, so it might be a good hint to just check that.
But is it enough to check that cmd `proj` returns a version number?

I wouldn't pin a version, as there are probably multiple ones that are compatible (using 6.3.1 locally), and it's nicer to be able to test them and still use `make test`.

Makefile
```diff
@@ -1,8 +1,39 @@
...
+	# remove PROJ system version from packages
+	sudo apt remove libproj-dev
+
+	wget https://github.com/OSGeo/proj.4/releases/download/$(PROJ_VERSION)/proj-$(PROJ_VERSION).tar.gz
+	tar -xzvf proj-$(PROJ_VERSION).tar.gz
+	pushd proj-$(PROJ_VERSION)
+	./configure --prefix=/usr && make
+	sudo make install
+	popd
 	rm -rf proj-$(PROJ_VERSION) proj-$(PROJ_VERSION).tar.gz
```
woshilapin, line 18: minor: do we want to `apt remove` once Proj has been compiled and installed?

pbougue: It was a comment previously (before the makefile even appeared), as we do not want to side-effect in removing some packages of the user.
So it would be only for doc if we keep it that way, and I considered commenting/separating the packages needed for proj install was covering for that (above).

Makefile
```diff
@@ -1,8 +1,39 @@
...
+clippy: lint ## Check quality of the code (alias for 'lint')
+lint: ## Check quality of the code
+	cargo clippy --workspace --all-features --all-targets -- --warn clippy::cargo --allow clippy::multiple_crate_versions --deny warnings
+
+test: ## Launch all tests
+	# Run all the tests of `transit_model` in the entire repository,
+	# activating all features (including `xmllint`), then without features
+	# to make sure that both work
+	cargo test --workspace --all-features
+	cargo test --workspace
```
woshilapin, line 33:
```suggestion
	cargo test --workspace --all-features --all-targets
	cargo test --workspace --all-targets
```

Makefile
pbougue: As discussed: not adding a check as the ratio value/effort is not clear.

Makefile
pbougue: As discussed, this removes doctests.
See https://github.com/rust-lang/cargo/issues/6669 and https://doc.rust-lang.org/cargo/commands/cargo-test.html

So adding this and the corresponding lines to just trigger doctests.

PR #672, opened 2020-07-10
[feature] Auto-generate route.destination_id if missing

woshilapin:
Auto-generate `route.destination_id` always (it was auto-generated only when `route.name` was empty).

ref: ND-916


woshilapin:
> I guess testing on a report (and full fixture) is a bit heavy for this, although I'm not the best to judge this yet :-)

I actually tried to take a look to test the Report. The problem being that `struct Report` has all the fields private and no getters (only setters). So a test would mean either comparing files or parsing the JSON file which might be a little bit too much. Another solution I'm thinking about as I write would be to have getters only in test mode...

PR #673, opened 2020-07-15
[tech] use comment id instead comment idx

patochectp:


src/model.rs
```diff
@@ -177,23 +177,8 @@ impl Collections {
...
-        {
-            for elt in container.iter_mut() {
-                let links = elt.comment_links_mut();
-                *links = links
-                    .iter()
-                    .map(|l| comment_old_idx_to_new_idx[l])
-                    .collect::<BTreeSet<_>>();
-            }
+            comments_used.extend(comment_links.iter().map(|cl| cl.to_string()));
         }
```
woshilapin, line 182: I'm wondering if this whole function is still necessary, maybe it can be inlined?

src/objects.rs
```diff
@@ -116,6 +117,12 @@ macro_rules! impl_comment_links {
             fn comment_links_mut(&mut self) -> &mut CommentLinksT {
                 &mut self.comment_links
             }
+            fn add_prefix_on_comment_id(&self, prefix: &str) -> CommentLinksT {
```
woshilapin, line 120: Instead of this method, I think you should be able to implement the trait `AddPrefix` on `CommentLinksT`.

tests/read_ntfs.rs
```diff
@@ -168,7 +168,9 @@ fn ntfs() {
     }
     assert_eq!(4, pt_objects.comments.len());
     let rera_comment_indexes = &pt_objects.lines.get("RERA").unwrap().comment_links;
```
woshilapin, line 170: These are not indexes anymore. Variable name can probably be adjusted.

```suggestion
    let rera_comment_ids = &pt_objects.lines.get("RERA").unwrap().comment_links;
```

src/gtfs/write.rs
```diff
@@ -73,11 +73,11 @@ fn get_first_comment_name<T: objects::CommentLinks>(
     obj: &T,
     comments: &CollectionWithId<objects::Comment>,
 ) -> Option<String> {
-    comments
-        .iter_from(obj.comment_links())
-        .map(|c| &c.name)
+    obj.comment_links()
+        .iter()
+        .filter_map(|comment_id| comments.get(comment_id))
+        .map(|cmt| cmt.name.clone())
```
pbougue, line 79: It looks like cloning here will clone every name, while previously only the selected one was cloned.

tests/fixtures/gtfs2ntfs/minimal/output/trips.txt
```diff
@@ -1,5 +1,5 @@
 trip_id,route_id,physical_mode_id,dataset_id,service_id,trip_headsign,trip_short_name,block_id,company_id,trip_property_id,geometry_id,journey_pattern_id
-trip:3,route:2,Metro,default_dataset,service:1,pouet,,,1,,,
 trip:4,route:2,Metro,default_dataset,service:1,pouet,,,1,,,
 trip:5,route:3,Metro,default_dataset,service:2,pouet,,,2,,,
+trip:3,route:2,Metro,default_dataset,service:1,pouet,,,1,,,
```
pbougue, line 4: Isn't the sorting supposed to be stable?

patochectp: There is no sorting

PR #674, opened 2020-07-15
[feature] Improve route_name generation in GTFS parsing

woshilapin:
When creating a Route, we want to apply generic `route_name` generation unless all trips in that route are in the same direction, in this case, we keep the route's name.

Bonus: The specification proposes to use `route_short_name` if `route_long_name` is empty but our previous implementation was only using `route_long_name`. This PR fixes it.

ref: ND-681

### DIFF
--- a/documentation/gtfs_to_ntfs_specs.md
+++ b/documentation/gtfs_to_ntfs_specs.md
@@ -158,15 +158,20 @@ _Warning :_ If the GTFS route has no trips, the Navitia Route should NOT be crea
 | NTFS file    | NTFS field     | Constraint | GTFS file  | GTFS field      | Note                                                                                                                                                        |
 | ------------ | -------------- | ---------- | ---------- | --------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------- |
 | routes.txt   | route_id       | ID         | routes.txt | route_id        | append a `_R` suffix for the Route grouping trips with `direction_id` = 1 (no suffix for `0` or undefined `direction_id`)                                   |
-| routes.txt   | route_name     | Required   | routes.txt | route_long_name | if `route_long_name` is empty, use `route_short_name`                                                                                                       |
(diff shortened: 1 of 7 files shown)


tests/fixtures/gtfs2ntfs/minimal/output/routes.txt
```diff
@@ -1,3 +1,3 @@
 route_id,route_name,direction_type,line_id,geometry_id,destination_id
-route:2,plop - plop,forward,route:2,,stoparea:1
-route:3,plop - plop,forward,route:3,,stoparea:1
+route:2,ma route 1,forward,route:2,,stoparea:1
```
woshilapin, line 2: This modification is due to using `route_short_name` when `route_long_name` is empty (the previous implementation was leaving it empty, and therefore, the auto-generation rule was triggered).

documentation/gtfs_to_ntfs_specs.md
```diff
@@ -158,15 +158,20 @@ _Warning :_ If the GTFS route has no trips, the Navitia Route should NOT be crea
...
 | routes.txt   | destination_id | Optional   |            |                 | This field contains a stop_area.id of the most frequent destination of the contained trips (ie. the parent_station of the most frequent last stop of trips) |
-| comments.txt | comment_value  | Optional   | routes.txt | route_desc      | See (2) for additional properties                                                                                                                           |
+| comments.txt | comment_value  | Optional   | routes.txt | route_desc      | See (3) for additional properties                                                                                                                           |
 
-(1) the field `direction_type` contains `backward` when grouping GTFS Trips with `direction_id` = 1, `forward` otherwise
+(1) if only one route is created (only one direction in included trips), use
+`route_long_name` or, if empty, use `route_short_name`. In case on multiple
```
papailio, line 168: `route_long_name` or, if empty, use `route_short_name`. In case of multiple

woshilapin: Fixed. Thanks.

PR #675, opened 2020-07-21
[feature] AddPrefix can prefix referential or scheduled objects

woshilapin:
Evolution of the `AddPrefix` trait in order to be able to prefix with more information. This evolution is needed for merging together objects of the same contributors that might have the same identifier but are different objects from `transit_model` point-of-view (for example, a `VehicleJourney` that would share the same identifier for the winter service and the summer service).

ref: ND-964

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 authors = ["Kisio Digital <team.coretools@kisio.com>", "Guillaume Pinot <texitoi@texitoi.eu>"]
 name = "transit_model"
-version = "0.23.0"
+version = "0.24.0"
 license = "AGPL-3.0-only"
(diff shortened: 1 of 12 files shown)


woshilapin:
I'm gonna try to add a test that combine a `config.json` and a prefix in the converter `gtfs2ntfs`.

ArnaudOggy:
> I'm gonna try to add a test that combine a `config.json` and a prefix in the converter `gtfs2ntfs`.

In https://github.com/CanalTP/transit_model/pull/677 :wink:

woshilapin:
Small fix concerning the case when dataset identifiers are smaller than 6 characters. See 7c29b7eef63f97abc67d4ec48dff813930bd6542.

PR #680, opened 2020-07-28
[tech] use crate chrono tz for timezone

patochectp:
ref: ND-940


src/objects.rs
```diff
@@ -1809,6 +1812,26 @@ impl AddPrefix for GridRelCalendarLine {
...
+    }
+}
+
+impl FromStr for TzExt {
+    type Err = String;
+    fn from_str(s: &str) -> std::result::Result<Self, Self::Err> {
+        match Tz::from_str(s) {
+            Ok(timezone) => Ok(TzExt(timezone)),
+            Err(err) => Err(err),
+        }
```
datanel, line 1831:
```suggestion
        Tz::from_str(s).map(|tz| TzExt(tz))
```

src/objects.rs
```diff
@@ -344,8 +345,9 @@ pub struct Network {
     pub url: Option<String>,
     #[serde(skip)]
     pub codes: KeysValues,
+    // #[derivative(Default(value = "\"Europe/Paris\".parse().ok()"))]
```
datanel, line 348: you can remove this comment.

src/objects.rs
```diff
@@ -1809,6 +1812,26 @@ impl AddPrefix for GridRelCalendarLine {
     }
 }
 
+#[derive(Debug, Serialize, Deserialize, Clone, Copy, PartialEq)]
+pub struct TzExt(pub Tz);
```
pbougue, line 1816: It's worth commenting why you had to do that, and maybe rename:
```suggestion
pub struct ParsableTz(pub Tz);
```

patochectp: we may have other needs so I prefer a generic name

PR #685, opened 2020-08-05
[feature] Full comment deduplication for referential objects, based on name only for similarity

pbougue:
Schedule objects are not handled, as no merge is done for them (duplicates would only be the result of initial data).

Ref. ND-979

TODO:

- [x] Check tests are OK for libs depending on this

### DIFF
--- a/src/model.rs
+++ b/src/model.rs
@@ -762,7 +762,7 @@ impl Collections {
 
     /// Some comments are identical and can be deduplicated
     pub fn comment_deduplication(&mut self) {
-        let duplicate2ref = &self.get_comment_map_duplicate_to_referent();
+        let duplicate2ref = self.get_comment_map_duplicate_to_referent();
         if duplicate2ref.is_empty() {
             return;
         }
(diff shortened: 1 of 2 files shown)


src/model.rs
```diff
@@ -760,6 +760,71 @@ impl Collections {
...
+    /// From comment collection only, return a map of the similar comments.
+    ///
+    /// Result: doubloons (comments to be removed) are mapped to their similar
+    /// referent (unique to be kept)
+    fn get_comment_map_doubloon_to_referent(&self) -> BTreeMap<String, String> {
+        let mut doubloon2ref: BTreeMap<String, String> = BTreeMap::new();
+        // Map of the referent comments id (uniqueness given the similarity_key)
+        let mut map_ref: BTreeMap<&str, &str> = BTreeMap::new();
+
+        for (_, comment) in &self.comments {
```
ArnaudOggy, line 817:
```suggestion
        let mut map_ref: HashMap<&str, &str> = HashMap::new();

        for comment in self.comments.values() {
```
Hashmap less heavy (and no order needed for map_ref)

datanel: or `let mut map_ref = HashMap::<&str, &str>::new();` you save 1 `HashMap`

src/model.rs
```diff
@@ -760,6 +760,70 @@ impl Collections {
         self.vehicle_journeys = CollectionWithId::new(vehicle_journeys).unwrap();
     }
 
+    /// Some comments are identical and can be deduplicated
+    pub fn comment_deduplication(&mut self) {
+        let duplicate2ref = &self.get_comment_map_duplicate_to_referent();
```
datanel, line 765:
```suggestion
        let duplicate2ref = self.get_comment_map_duplicate_to_referent();
```

src/model.rs
```diff
@@ -760,6 +760,70 @@ impl Collections {
...
+                let mut intersection = vec![];
+                for comment_id in pt_object.comment_links() {
+                    if let Some((duplicate_id_ref, _)) = duplicate2ref.get_key_value(comment_id) {
+                        intersection.push(duplicate_id_ref.as_str());
+                    }
+                }
+                if !intersection.is_empty() {
+                    map_pt_object_duplicates.insert(idx, intersection);
+                }
+            }
```
datanel, line 793: if you want "rustic"  code

```suggestion
           let map_pt_object_duplicates: BTreeMap<Idx<T>, Vec<&str>> = collection
                .iter()
                .filter_map(|(idx, pt_object)| {
                    let intersection: Vec<&str> =
                        pt_object.comment_links().iter().filter_map(|comment_id| {
                            duplicate2ref
                                .get_key_value(comment_id)
                                .map(|(duplicate_id_ref, _)| duplicate_id_ref.as_str())
                        }).collect();

                    if !intersection.is_empty() {
                        Some((idx, intersection))
                    } else {
                        None
                    }
                })
                .collect();
```
a bit longer but in one step

PR #686, opened 2020-08-06
[feature] Transfers: add the possibility to determine if a transfer could be created

datanel:
By default transfers that do not already exist and where the distance is less than `max_distance` will be created.

You can now provide a closure that determines if a transfer must be created.
For example, you could create transfers between stop points that belong to 2 different contributors only.

See the test https://github.com/CanalTP/transit_model/pull/686/files#diff-2349738cf674a54d8af122d1ec0bf821R57


src/transfers.rs
```diff
@@ -25,6 +25,10 @@ use typed_index_collection::{Collection, CollectionWithId, Idx};
 
 type TransferMap = HashMap<(Idx<StopPoint>, Idx<StopPoint>), Transfer>;
 
+/// The closure that will determine whether a connection should be created between 2 stops.
+/// See See [generates_transfers](./fn.generates_transfers.html).
+pub type NeedTransfer<'a> = Box<dyn 'a + FnMut(&Model, Idx<StopPoint>, Idx<StopPoint>) -> bool>;
```
pbougue, line 30: Why allowing mutation of the context?
```suggestion
pub type NeedTransfer<'a> = Box<dyn 'a + Fn(&Model, Idx<StopPoint>, Idx<StopPoint>) -> bool>;
```

src/transfers.rs
```diff
@@ -49,6 +53,7 @@ fn generate_transfers_from_sp(
     max_distance: f64,
     walking_speed: f64,
     waiting_time: u32,
+    mut need_transfer: Option<NeedTransfer>,
```
pbougue, line 56: Why would this mutate along the transfers build?
```suggestion
    need_transfer: Option<NeedTransfer>,
```

src/transfers.rs
datanel: It was necassary until I change my code in tartare-tools

datanel: done

src/transfers.rs
datanel: removed

PR #688, opened 2020-08-07
[feature] Gtfs2Ntfs: added parameter --schedule-prefix

datanel:
ref: ND-993

This is a breaking change. schedule prefix is no longer automatically created by the 6 characters of the dataset id.
a new parameter `--schedule-prefix` has been added and may be provided in `gtfs2ntfs`

\+ some markdown linting


gtfs2ntfs/src/main.rs
```diff
@@ -45,6 +45,10 @@ struct Opt {
     #[structopt(short, long)]
     prefix: Option<String>,
 
+    /// Schedule sub prefix after the prefix on all scheduled objects (`123` turned into `prefix::schedule_subprefix::123`).
```
pbougue, line 48:
```suggestion
    /// Schedule subprefix added after the prefix on all scheduled objects (`123` turned into `prefix::schedule_subprefix::123`).
```

Issue #329, opened 2019-08-27
Update Readme


datanel:
We could have a better Readme
- remove references to binaries once https://github.com/CanalTP/transit_model/pull/316 is merged
- complete the list of features (maybe add some examples in `./examples`)

antoine-de:
I do agree that the documentation could be improved :stuck_out_tongue_winking_eye: 

It would for example be very nice to see the ergonomic of handling transit data. It's a killer feature, but It's impossible to know it's existence!

something like?

```rust
let transit_data = transit_model::gtfs::read_from_path("example", None, None, false).expect("impossible to read GTFS");

// get a stop by it's ID
let stop_area_idx = transit_data
    .stop_areas
    .get_idx("stoparea:1")
    .expect("impossible to find stop area");

println!("routes passing by stop stoparea:1");

// efficiently get all the routes that pass through the stop_area
for route in transit_data
    // The rust typesystem is able to guess that we are looking for the routes linked to the stop
    // because the Idx are strongly typed, thus route_idx is Idx<Route>
    .get_corresponding_from_idx(stop_area_idx)
    .into_iter()
    .map(|route_idx| &transit_data.routes[route_idx])
{
    println!(" - {}", route.name);
}
```

And maybe add a reference to the quite good [technical documentation](https://docs.rs/transit_model) ?

datanel:
yes that's the idea

datanel:
the readme has been updated but I think we need to provide some examples as you suggest @antoine-de

PR #692, opened 2020-08-27
[doc] improve examples

pbougue:
fix https://github.com/CanalTP/transit_model/issues/329

Also some clippy-related work.

### DIFF
--- a/examples/ntfs_crawler.rs
+++ b/examples/ntfs_crawler.rs
@@ -1,4 +1,4 @@
-// Copyright (C) 2017 Kisio Digital and/or its affiliates.
+// Copyright (C) 2020 Kisio Digital and/or its affiliates.
 //
 // This program is free software: you can redistribute it and/or modify it
 // under the terms of the GNU Affero General Public License as published by the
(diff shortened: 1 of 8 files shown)


datanel (approved):

examples/ntfs_crawler.rs
```diff
@@ -30,16 +30,30 @@ where
...
+    // stop_area by stop_area, print PT objects related to it
+    for (idx, stop_area) in &transit_objects.stop_areas {
+        // retrieve idx from id
+        assert_eq!(
+            transit_objects.stop_areas.get_idx(&stop_area.id).unwrap(),
+            idx
+        );
+
+        // lines passing by stop
+        let lines = get(idx, &transit_objects.lines, &transit_objects);
```
antoine-de, line 45: completly minor, but don't you think that for the purpose of an example it would be more straightforward to directly use `get_corresponding`? (at least for the first ones)

antoine-de (commented):
nice!

But don't you think the link is a bit easy to miss and that some inline code in the doc would have more impact (with [doc-comment](https://crates.io/crates/doc-comment), [skeptic](https://github.com/budziq/rust-skeptic) or whatewhere is the trend now :stuck_out_tongue_winking_eye: )

datanel:
> But don't you think the link is a bit easy to miss and that some inline code in the doc would have more impact (with [doc-comment](https://crates.io/crates/doc-comment), [skeptic](https://github.com/budziq/rust-skeptic) or whatewhere is the trend now  )

@antoine-de  get_corresponding is already well documented in https://github.com/CanalTP/relational_types/blob/master/src/lib.rs

examples/ntfs_crawler.rs
pbougue: True (especially if it goes in main readme), I tried to mix existing and suggestions :dancers:

pbougue (commented):
@antoine-de I'm OK with your suggestion.
That said:
* I don't think that lib examples should be too prominent compared to bin (maybe move examples below if they are in main readme)
* I don't plan on spend much more time on this major doc refactor topic in the medium term
* I'm OK to merge something that goes that direction (You know, open-source and all :stuck_out_tongue_winking_eye: )

antoine-de:
> 
> 
> > But don't you think the link is a bit easy to miss and that some inline code in the doc would have more impact (with [doc-comment](https://crates.io/crates/doc-comment), [skeptic](https://github.com/budziq/rust-skeptic) or whatewhere is the trend now  )
> 
> @antoine-de get_corresponding is already well documented in https://github.com/CanalTP/relational_types/blob/master/src/lib.rs

sure, but it's more difficult to discover as is.


> 
> 
> @antoine-de I'm OK with your suggestion.
> That said:
> 
>     * I don't think that lib examples should be too prominent compared to bin (maybe move examples below if they are in main readme)
> 
>     * I don't plan on spend much more time on this major doc refactor topic in the medium term
> 
>     * I'm OK to merge something that goes that direction (You know, open-source and all 😜 )

honestly I don't plan on working on transit_model in the mid term, you can do whatever you want with my suggestion 😜

PR #694, opened 2020-08-31
[tech] Use action-rs/audit-check

pbougue:
* It visually breaks on vulnerabilities, in forks or not
* It does **not** block merging a PR (`continue-on-error: true` as I think we might end up with some vulnerabilities that we could not immediately deal with, and that are sometimes out of our reach : deps of deps)
* It displays a nice report **only** on main-repo branches, not on forks (but console output is readable anyway)
* It does **not** visually output warnings (one have to search for the report or console output). And no `--deny-warning` available (not sure we would want it anyway as it might be too strict).

Current warnings (no vulnerabilities):
* failure is deprecated/unmaintained (already identified to be replaced)
* term (brought by slog-term) is looking for maintainer since end of 2018

--Comment heavily updated--

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,6 +39,17 @@ jobs:
     - name: Linting
       run: make lint
 
+  audit:
+    name: Audits
+    runs-on: ubuntu-latest
+    continue-on-error: true
+    steps:
+    - uses: actions/checkout@v1
+    - name: Security audit
+      uses: actions-rs/audit-check@v1
+      with:
+        token: ${{ secrets.GITHUB_TOKEN }}
+
   tests:
     name: Tests
     runs-on: ${{ matrix.os }}


woshilapin:
Did you test it? Because I'm not sure I understand fully what will the Github Actions do exactly. Will it report as a comment in the PR? Because if not, then it will only be part of the logs of the build and, nobody is looking at them as long as the build is passing... which defeat the purpose.

woshilapin:
For information, it adds a section in the CI UI.
https://github.com/CanalTP/transit_model/pull/694/checks?check_run_id=1056552631.

Warnings might be a bit invisible (as you have to browse to this section to see if there is any warning), but hopefully, errors will fail the build.

Cargo.toml
```diff
@@ -52,6 +52,7 @@ minidom = "0.12"
 minidom_ext = "1"
 minidom_writer = "1"
 num-traits = "0.2"
+once_cell = "=1.0.0-pre.2"
```
woshilapin, line 55: Probably gonna need to remove that before merging. Apart from that, I'm approving this message.

woshilapin (approved):

pbougue:
Here is the status of this PR.
* It visually breaks on vulnerabilities, in forks or not (see :x: in audit above when I added the vulnerable crate)
* It does **not** prevent from merging a PR (`continue-on-error: true` as I think we might end up with some vulnerabilities that we could not immediately deal with, and that are sometimes out of our reach : deps of deps) : Can be discussed.
* It displays a nice report **only** on main-repo branches, not on forks (but console output is readable anyway)
* It does **not** visually output warnings (one have to search for the report or console output). And no `--deny-warning` available (not sure we would want it anyway as it might be too strict).

TODO:
* Decide if we keep it: IMO :+1: as it will output useful things when CVE appears, and even if we don't deal with it right away, the :x: will be a good enough motivator.
* Remove that vulnerable crate before merge if we agree on it.
* Decide if it should be blocking PR, if we use it.

Cargo.toml
datanel: same

datanel (approved):

PR #698, opened 2020-09-03
[feature] Make transit_model::Result generic on Error

woshilapin:
This small PR takes inspiration from [this little tip](https://octodon.social/@rust/104670964806314333).

**Note -** This is not a breaking change even if we change the type `transit_model::Result` which is part of our API.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 authors = ["Kisio Digital <team.coretools@kisio.com>", "Guillaume Pinot <texitoi@texitoi.eu>"]
 name = "transit_model"
-version = "0.31.0"
+version = "0.31.1"
 license = "AGPL-3.0-only"
 description = "Transit data management"
 repository = "https://github.com/CanalTP/transit_model"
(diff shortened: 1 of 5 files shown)


patochectp (approved):

datanel:
Is it in Draft for discussion or you are working on it? (label WIP)

woshilapin:
It's a draft for discussion, that's why I didn't put the `wip` label.

datanel:
but we never put wip when it's draft. anyway
It seems good to me

woshilapin:
Another side effect is this. Before, the documentation would appear as

image

Now, it will appear as

image

And when the error type is `transit_model::Error`, the documentation does not change, it is still showing as (does not show the error type when it is the default one).

image

`StdResult` felt like a new type the reader of the documentation would need to check to understand (arguably one more click to understand the signature of the function). Now, it shows like the standard `Result`... although, it's a little bit misleading because it looks like it is the standard `Result` but in fact, it is `transit_model::Result`. Personally, I'm fine with this.

pbougue (approved):

PR #699, opened 2020-09-09
[feature] Automatically remove empty comments from the model

woshilapin:
The Model can be cleaned up of comments without any text.

ref: ND-1003.

### DIFF
--- a/src/model.rs
+++ b/src/model.rs
@@ -811,6 +811,46 @@ impl Collections {
         }
     }
 
+    /// Remove comments with empty message from the model
+    pub fn clean_comments(&mut self) {
+        fn remove_comment<T: Id<T> + CommentLinks>(
+            collection: &mut CollectionWithId<T>,
+            comment_id: &str,
+        ) {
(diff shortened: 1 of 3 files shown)


datanel (dismissed):

datanel (approved):

ArnaudOggy (approved):

woshilapin:
I realized I forgot to bump the version of `transit_model`.

src/model.rs
```diff
@@ -1172,6 +1212,7 @@ impl Model {
     /// ```
     pub fn new(mut c: Collections) -> Result<Self> {
         c.comment_deduplication();
+        c.clean_comments();
```
pbougue, line 1215: Current "comment_deduplication" would handle multiple things, but for performances it's probably way better to do it inside `comment_deduplication` (and rename?).
If so, to b future-proof, please keep in mind that the key for comment comparison might change (and contain more than just `comment.name` in the future) :wink:

PR #701, opened 2020-09-09
[feature] Improve test utils for report on files count

woshilapin:
Historically, in our `test_utils` framework, we test the number of files produced vs the number of files in the fixture and compare them. This sometimes leads to something like this kind of error.

image

This is not very helpful. So I changed it to a comparison of the 2 lists which would show something like this (more helpful I believe). Note that I used a `BTreeSet` instead of a `Vec` so the list is ordered and the diff much more helpful.

image

As a bonus, hidden files are also filtered. This was inconvenient when you're currently editing a file (like `transfers.txt`) because some editors like `vim` create a swap file (like `.transfers.txt.swp`).

### DIFF
--- a/src/test_utils.rs
+++ b/src/test_utils.rs
@@ -47,7 +47,7 @@ pub fn get_lines_content<P: AsRef<Path>>(path: P) -> BTreeSet<String> {
     set
 }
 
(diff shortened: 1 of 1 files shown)


datanel (approved):
much better indeed

PR #702, opened 2020-09-10
[fix] Do not create transfers when geolocation is (0, 0)

woshilapin:
When a `StopPoint` geolocation is `(0, 0)`, generating a transfer doesn't make sense and will create weird and incorrect behaviors. This PR prevent the creation of such transfers.

ref: ND-1016.

### DIFF
--- a/src/transfers.rs
+++ b/src/transfers.rs
@@ -16,10 +16,10 @@
 
 use crate::{
     model::Model,
-    objects::{StopPoint, Transfer},
+    objects::{Coord, StopPoint, Transfer},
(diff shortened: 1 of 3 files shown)


patochectp (approved):

datanel (approved):

datanel (commented):
If it's a fix, you don't increment the version?

woshilapin:
> If it's a fix, you don't increment the version?

It's already version 0.31.2 in `Cargo.toml` which has not yet been released.

PR #703, opened 2020-09-11
[feature] Default forbid pickup/dropoff on last/first stop times of vehicle journeys

woshilapin:
For trip planning to be handled correctly by Navitia, we need to forbid pick-up on the last stop time of each vehicle journey and drop-off on the first stop time of each vehicle journey. However, there is one specific case when we still want to leave the possibility of pick-up on the last and drop-off on the first: prolongation trips when the last stop of the first trip and the first stop of second trip are different.

- [x] Squash/rebase the PR

ref: ND-991.


src/model.rs
```diff
@@ -724,6 +724,124 @@ impl Collections {
...
+            .collect();
+        // Sort the vehicule journeys by block_id, and if equal, order them by
+        // departure time.
+        vj_idxs.sort_by(|vj1_idx, vj2_idx| {
+            let vj1 = &self.vehicle_journeys[*vj1_idx];
+            let vj2 = &self.vehicle_journeys[*vj2_idx];
+            match vj1.block_id.cmp(&vj2.block_id) {
+                Ordering::Equal => vj1.stop_times[0]
+                    .departure_time
+                    .cmp(&vj2.stop_times[0].departure_time),
```
patochectp, line 796: I think sorting is not good. It must be done on the minimal positive duration between the last schedule of the first vj ant the first schedule of the second vj.

woshilapin: If they're part of the same `block_id` (condition `Ordering::Equal` to end up on this line of the code), then ordering by starting time should be enough I believe. Remember that in any case, I'm checking for overlapping times later in this function.

If I understand correctly what you propose, I believe it will actually cause problems. Imagine a VJ_A that starts at 10:00 et arrives at 14:00 then a VJ_B that starts at 11:00 et arrives at 13:00. Following your rule, the sorting will be different depending of if VJ_A is `vj1` or is VJ_A is `vj2` (or said differently, depending on the order they appears in `trips.txt`).

In first case, `vj1.stop_times[vj1.stop_times.len() - 1] = 14:00` (last schedule of first vj) is bigger than `vj2.stop_times[0] = 11:00` (first schedule of second vj), therefore final order will be `VJ_B -> VJ_A`. 

In second case, `vj1.stop_times[vj1.stop_times.len() - 1] = 13:00` (last schedule of first vj) is bigger than `vj2.stop_times[0] = 10:00` (first schedule of second vj), therefore final order will be `VJ_A -> VJ_B`.

woshilapin: After a discussion with @patochectp, the following case would cause a problem.

```    
          SP0           SP1           SP2           SP3            
           |             |             |             |             
VJ1   09:00-10:00   10:00-11:00        |             |             
           |             |             |             |             
VJ2        |             |        12:00-13:00   14:00-15:00        
           |             |             |             |             
VJ3   09:30-10:30   10:30-11:30        |             |             
           |             |             |             |             
VJ4        |             |        12:30-13:30   14:30-15:30        

```
In this case, VJ1 and VJ2 should be considered part of a prolongation trip pair; VJ3 and VJ4 should be considered part of another prolongation trip pair. But in terms of modelization, they might have the same `block_id`.

The rest of the implemented algorithm relies on the fact that successive vehicle journeys in a prolongation trip pair, should be successive after the sort. Therefore, a correct sort is VJ1 < VJ2 < VJ3 < VJ4. But the sort function I implemented so far will use `stop_times[0].departure_time` and the resulting sort will be VJ1 < VJ3 < VJ2 < VJ4 which is incorrect.

woshilapin:
Some updates about this PR (removing the `wip` label as well as the `Draft` status).

- e18b4a2fd0f7de7c417f5a1f4b6da211f855f219 is about improving documentation about the different cases, as well as making explicit the scenarios the current implementation does NOT handle.
- 32bd4146918ab0be793922a620940e8410c8abab fixes all the fixtures so the tests pass
- f9bf9c406487989cfa89e5d344f02f7740d3839d is also about fixing the tests but it is a bit specific. It's part of Netex FR parser and is trying to generate Journey Patterns from multiple Vehicle Journeys. Bad luck, we had 2 different Vehicle Journeys in the test but the current PR was making these 2 Vehicle Journeys the same -> the tests were expecting 2 Journey Patterns but was getting one. I changed the second Vehicle Journey so it doesn't have the same Journey Pattern than the first.

woshilapin:
A small improvement, I moved the `.sort_by()` later in the implementation, after removing all vehicle journeys that do not contain a `block_id` (call to `.retain()`). See 4a1563ff3b254a7233e8c0db017b6974f988c8c8.

src/model.rs
```diff
@@ -724,6 +724,138 @@ impl Collections {
...
+                ordering => ordering,
+            }
+        });
+        for (prev_vj_idx, next_vj_idx) in vj_idxs.iter().zip(vj_idxs.iter().skip(1)) {
+            let prev_vj = &self.vehicle_journeys[*prev_vj_idx];
+            let next_vj = &self.vehicle_journeys[*next_vj_idx];
+            if prev_vj.block_id != next_vj.block_id {
+                // we can discard two vehicle journeys which are not part of the
```
pbougue, line 832:
```suggestion
                // we can discard last pickup/first dropoff if two vehicle journeys are not part of the
```

src/model.rs
```diff
@@ -724,6 +724,138 @@ impl Collections {
...
+                // same prolongation trip.
+                continue;
+            }
+            let stop_times_len = prev_vj.stop_times.len();
+            let last_stop = &prev_vj.stop_times[stop_times_len - 1];
```
pbougue, line 837: you can probably use (deref to slice, and you know it's not empty)
```suggestion
            let last_stop = &prev_vj.stop_times.last().unwrap();
```

src/model.rs
```diff
@@ -724,6 +724,138 @@ impl Collections {
...
+                continue;
+            }
+            if last_stop.departure_time > first_stop.arrival_time {
+                // We can discard when timing overlaps between arrival of first
+                // vehicle journey and departure of next vehicle journey of a
+                // prolongation trip (see Example 2 above).
+                continue;
+            }
+            // We're now in Example 3 (see above), let's allow pick-up and
+            // drop-off on prolongation trip.
```
pbougue, line 851: Maybe add a comment on the fact it might be a `2` (ODT) originally here but it is lost.

src/model.rs
```diff
@@ -1438,6 +1571,249 @@ mod tests {
         }
     }
 
+    mod enhance_pickup_dropoff {
```
pbougue, line 1574: Not sure (as I don't get the need for a specific module here):
```suggestion
    mod enhance_pickup_dropoff_tests {
```

src/model.rs
```diff
@@ -724,6 +724,138 @@ impl Collections {
...
+    ///                         out         in   out         in
+    ///                         |- Stay-In -|
+    ///
+    /// Example 2:
+    /// ##########
+    ///       out          in   out         in   out         in   out         in
+    ///        X    SP1    |    ▲    SP2    X    X    SP3    |    ▲   SP4     X
+    ///        X           ▼    |           X    X           |    |           X
+    ///  VJ:1   08:00-09:00      10:00-12:00     X           ▼    |           X
+    ///  VJ:2                                     11:00-13:00      13:00-14:00
```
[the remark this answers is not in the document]
woshilapin: Done in d243875dc8ded24e41cafcf985e425de349a83e3.

src/model.rs
woshilapin: First of all, the use of `.unwrap()` here is not riskier than using slices `[]` since slice's operator panics if out-of-bound. So what you propose is equivalent and I do indeed prefer it.

I've actually made the modification everywhere to replace with `.first()`/`.last()` (`.first_mut()`/`.last_mut()` when it applies). See 2f90dfbf71468da64af9fd87cc05f538ac17cd2e.

src/model.rs
woshilapin: Done in 9d01051ae1b943af6b669f445179534b8e0d3ba6.

src/model.rs
woshilapin: I changed the wording but didn't apply what you proposed because we're not really discarding last pick-up at this point of the code (it has been done above). Here, we're choosing to not reenable it.

src/model.rs
```diff
@@ -724,6 +724,138 @@ impl Collections {
...
 
+    /// Forbid pickup on last stop point of vehicle journeys and forbid dropoff
+    /// on first stop point of vehicle journeys.
+    ///
+    /// However, there is an exception to this rule for prolongation trips. It
+    /// is possible to get in the last stop point of a vehicle journey and get
+    /// out on the first stop point of a vehicle journey, if and only if the 2
+    /// stop points are different and times do not overlap.
+    ///
+    /// WARNING: The current implementation does not handle prolongation trips
```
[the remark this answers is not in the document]
woshilapin: Done in 9d01051ae1b943af6b669f445179534b8e0d3ba6.

src/model.rs
```diff
@@ -724,6 +724,138 @@ impl Collections {
         }
     }
 
+    /// Forbid pickup on last stop point of vehicle journeys and forbid dropoff
+    /// on first stop point of vehicle journeys.
+    ///
+    /// However, there is an exception to this rule for prolongation trips. It
+    /// is possible to get in the last stop point of a vehicle journey and get
```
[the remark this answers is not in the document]
woshilapin: Done in 9d01051ae1b943af6b669f445179534b8e0d3ba6.

src/model.rs
```diff
@@ -724,6 +724,138 @@ impl Collections {
         }
     }
 
+    /// Forbid pickup on last stop point of vehicle journeys and forbid dropoff
+    /// on first stop point of vehicle journeys.
+    ///
+    /// However, there is an exception to this rule for prolongation trips. It
```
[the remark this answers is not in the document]
woshilapin: Done in 9d01051ae1b943af6b669f445179534b8e0d3ba6.

src/model.rs
woshilapin: The only goal of these modules is to regroup tests by category (all the tests in this module relate to the algorithm about pick-up and drop-off). We could suffix with `_tests` but:
- it's not the convention we've used so far in other similar modules
- this module is already wrapped in a module `tests` so it's namespaced

src/model.rs
```diff
@@ -724,6 +724,156 @@ impl Collections {
...
+    /// current implementation since the validity patterns are different; this
+    /// is undefined behavior (most likely, some stay-in will be forbidden by
+    /// the default rule).
+    pub fn enhance_pickup_dropoff(&mut self) {
+        let mut vj_idxs: Vec<Idx<VehicleJourney>> = self
+            .vehicle_journeys
+            .iter()
+            // Filtering out vehicle journeys without any stop time allowing us
+            // below to access first and last stop_times safely.
+            .filter(|(_, vj)| !vj.stop_times.is_empty())
```
pbougue, line 802: Would it be better to filter on length <= 1 ?

src/model.rs
```diff
@@ -724,6 +724,156 @@ impl Collections {
...
+    /// between vehicle journeys. It is possible to get in the last stop point
+    /// of a vehicle journey or get out on the first stop point of a vehicle
+    /// journey, if and only if the 2 stop points are different and times do not
+    /// overlap.
+    ///
+    /// WARNING: The current implementation does not handle stay-in for vehicle
+    /// journeys with different validity patterns.
+    ///
+    /// Here is examples explaining the different situations of stay-in
+    /// situations (for pick-up and drop-off, XX means forbidden, ―▶ means
```
pbougue, line 740:
```suggestion
    /// Here are examples explaining the different stay-in situations
    /// (for pick-up and drop-off, XX means forbidden, ―▶ means
```

src/model.rs
```diff
@@ -835,26 +836,29 @@ impl Collections {
...
+                // we can discard when two vehicle journeys are not in a stay-in
+                // situation.
                 continue;
             }
             let last_stop = &prev_vj.stop_times.last().unwrap();
             let first_stop = &next_vj.stop_times.first().unwrap();
             if last_stop.stop_point_idx == first_stop.stop_point_idx {
-                // We can discard prolongation trip when the stop point is
-                // identical (see Example 1 above).
+                // We can when the stop points are identicals (see Example 1
```
pbougue, line 846:
```suggestion
                // We can discard when the stop points are identicals (see Example 1
```

pbougue (approved):

src/model.rs
woshilapin: Technically, with a vehicle journey containing only one `stop_time`, I believe we should forbid pick-up and drop-off on this only stop and pretty much make the vehicle journey unusable.

Now, if we filter with `.len() <= 1` as you propose, then we won't forbid pick-up and drop-off on this weird vehicle journey with the more risk it end up being showed up to a traveler? In terms of traveler information, I'd think we would prefer to forbid at maximum this kind of weird vehicle journey?

src/model.rs
woshilapin: Done in 0e5e40408a6ab5201738c7c4c344dde909a9ac04.

src/model.rs
woshilapin: Done in 0e5e40408a6ab5201738c7c4c344dde909a9ac04.

pbougue (approved):

src/model.rs
woshilapin: We decided to not support stay-in situations for vehicle journeys with different validity patterns. The above case is most likely to make sense with different validity patterns. If the 4 vehicle journeys above have the same validity pattern, then VJ1/VJ2 will most likely have a different `block_id` than VJ3/VJ4.

To answer to the original [comment](https://github.com/CanalTP/transit_model/pull/703#discussion_r486838087) about the sort, using `departure_time` as a sort-key is not perfect but should be enough if we ignore validity patterns. The logical implication here might not be obvious (it's a shortcut :stuck_out_tongue:), but source code has been documented for more details about that reasoning.

I'm resolving this conversation.

patochectp (approved):

PR #705, opened 2020-09-18
[feature] warning if more than one timezone is reported on agency file

patochectp:
ref: ND-1005.

### DIFF
--- a/src/gtfs/read.rs
+++ b/src/gtfs/read.rs
@@ -576,6 +576,22 @@ where
 {
     let filename = "agency.txt";
     let gtfs_agencies = read_objects::<_, Agency>(file_handler, filename)?;
+
+    if let Some(referent_agency) = gtfs_agencies.first() {
+        for agency in gtfs_agencies.iter().skip(1) {
+            if referent_agency.timezone != agency.timezone {
+                warn!(
+                    "different agency timezone: {} ({}) - {} ({})",
+                    referent_agency.timezone,
+                    referent_agency.id.clone().unwrap_or_default(),
+                    agency.timezone,
+                    agency.id.clone().unwrap_or_default(),
+                );
+                break;
+            }
+        }
+    }
+
     let networks = gtfs_agencies
         .iter()
         .cloned()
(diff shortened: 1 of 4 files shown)


src/gtfs/read.rs
```diff
@@ -576,6 +576,17 @@ where
...
+
+    for (current_agency, next_agency) in gtfs_agencies.iter().zip(gtfs_agencies.iter().skip(1)) {
+        if current_agency.timezone != next_agency.timezone {
+            warn!(
+                "different agency timezone: {} - {}",
+                current_agency.timezone, next_agency.timezone
+            );
+            break;
+        }
+    }
```
woshilapin, line 588: How about something like this?

```suggestion
    let timezones: HashSet<_> = gtfs_agencies.iter().map(|agency| agency.timezone).collect();
    if timezones.len() > 1 {
        if current_agency.timezone != next_agency.timezone {
            warn!(
                "different agency timezone: {} - {}",
                current_agency.timezone, next_agency.timezone
            );
            break;
        }
    }
```

src/gtfs/read.rs
```diff
@@ -1450,6 +1461,28 @@ mod tests {
...
+            create_file_with_content(path, "agency.txt", agency_content);
+            super::read_agency(&mut handler).unwrap();
+            testing_logger::validate(|captured_logs| {
+                assert_eq!(captured_logs.len(), 2);
+                assert_eq!(
+                    captured_logs[1].body,
+                    "different agency timezone: Europe/London - Europe/Paris"
+                );
+                assert_eq!(captured_logs[1].level, LogLevel::Warn);
+            });
```
woshilapin, line 1482: :+1:

src/gtfs/read.rs
```diff
@@ -576,6 +576,20 @@ where
 {
     let filename = "agency.txt";
     let gtfs_agencies = read_objects::<_, Agency>(file_handler, filename)?;
+
+    for (current_agency, next_agency) in gtfs_agencies.iter().zip(gtfs_agencies.iter().skip(1)) {
```
pbougue, line 580: Why not just compare all others to the first one (that would be the only referent, and it would save one iterator) ?
It looks weird to me to use a different "referent" each time we compare (in the current version), but it works.

Do we want to point out the less-represented timezone ? Not sure it's super useful, and it would require more code.

pbougue (commented):
Should we add test with only one agency ?

src/gtfs/read.rs
ArnaudOggy: @pbougue 
It breaks at the first difference, and without memory allocation :thinking:

ArnaudOggy (approved):

src/gtfs/read.rs
pbougue: To clarify my comment after discussion, I suggest comparing only `gtfs_agencies.first()` with all the following.
Something along that (not tested at all, maybe store only referent timezone, not agency):
```rust
let ref_agency = gtfs_agencies.first();

for agency in gtfs_agencies.iter().skip(1) {
    if ref_agency.timezone != agency.timezone {
        [...]
    }
}
```

Anyway it's just a suggestion not a veto as current code will warn if and only if problems arise.

patochectp: the number of iterations will be the same.
If we have a, b, ,c and d

my proposal will give : 
 a compare with b
 b compare with c
 c compare with d

your proposal will give : 
 a compare with b
 a compare with c
 a compare with d

woshilapin: I like @pbougue's solution because:
- it's one iterator instead of 2 iterators
- it feels easier to understand what we're trying to do (the `.zip()` feel a little less natural)

But in any case, I'm fine with whatever we end up merging.

patochectp: I capitulate because it is more readable

pbougue (approved):

woshilapin (approved):

PR #713, opened 2020-10-27
[feature] add object codes on companies

patochectp:
ref: ND-1080

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 authors = ["Kisio Digital <team.coretools@kisio.com>", "Guillaume Pinot <texitoi@texitoi.eu>"]
 name = "transit_model"
-version = "0.31.5"
+version = "0.32.0"
 license = "AGPL-3.0-only"
 description = "Transit data management"
 repository = "https://github.com/CanalTP/transit_model"
--- a/src/gtfs/read.rs
+++ b/src/gtfs/read.rs
@@ -79,6 +79,7 @@ impl From<Agency> for objects::Company {
             url: Some(agency.url),
             mail: agency.email,
             phone: agency.phone,
+            codes: BTreeSet::new(),
         }
     }
 }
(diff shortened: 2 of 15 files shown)


Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 authors = ["Kisio Digital <team.coretools@kisio.com>", "Guillaume Pinot <texitoi@texitoi.eu>"]
 name = "transit_model"
-version = "0.31.5"
+version = "0.31.6"
```
ArnaudOggy, line 4: `0.32.0` maybe ?

woshilapin:
This PR can't be merged before the specification for NTFS is modified accordingly (see https://github.com/CanalTP/ntfs-specification/pull/22). This PR will lead to the release of NTFS 0.11.4, we should also update the supported version of NTFS in `transit_model` (see [here](https://github.com/CanalTP/transit_model/blob/master/src/lib.rs#L61) and [here](https://github.com/CanalTP/transit_model/blob/master/src/ntfs/mod.rs#L397) and probably some fixtures).

Cargo.toml
woshilapin, line 4: Should we also upgrade all the other crates of the project?

woshilapin:
https://github.com/CanalTP/ntfs-specification/pull/22 has been merged and version 0.11.4 of the NTFS specification has been released.

woshilapin (approved):

src/gtfs/read.rs
```diff
@@ -79,6 +79,7 @@ impl From<Agency> for objects::Company {
             url: Some(agency.url),
             mail: agency.email,
             phone: agency.phone,
+            codes: BTreeSet::new(),
```
ArnaudOggy, line 82: `..Default::default()` maybe ?
And same thing in tests in in `/src/ntfs/mods.rs`

patochectp: In From trait, I prefer it to be explicit.
in the tests why not @woshilapin ?

ArnaudOggy: It's just a detail. PR is ready to go :wink:

woshilapin: For one field, I'd prefer it to be explicit.

ArnaudOggy: Ok.So go !

ArnaudOggy (approved):

PR #714, opened 2020-10-29
[feature] Remove ZipHandler

woshilapin:
In the direction of eventually removing `FileHandler`, this PR only removes `ZipHandler` for now.

One interesting question remaining is about `transit_model::gtfs::read_from_zip()`. I kept this API replacing it with a simple unzip into a temporary folder and using `transit_model::gtfs::read_from_path()` instead. But in reality, I'm breaking a little bit more than that because we were supporting different kind of ZIP:
- with all NTFS files at the root of the ZIP file
- with all NTFS files inside a sub-folder of the ZIP file
	- even with hidden files

Do we want to keep `transit_model::gtfs::read_from_zip()` or completely remove it? My vote would be to remove it. For now, we're using it in `gtfs2ntfs` and `gtfs2netexfr`. We could remove `zip` dependency if we do that.

ref: ND-956.


src/gtfs/mod.rs
```diff
@@ -353,8 +359,20 @@ pub fn read_from_path<P: AsRef<Path>>(p: P, configuration: Configuration) -> Res
 /// identifiers, allowing to namespace the dataset. By default, no
 /// prefix will be added to the identifiers.
 pub fn read_from_zip<P: AsRef<Path>>(path: P, configuration: Configuration) -> Result<Model> {
-    let mut file_handler = read_utils::ZipHandler::new(path)?;
-    read(&mut file_handler, configuration)
+    let output_dir = tempfile::TempDir::new()?;
```
woshilapin, line 362: What about the weird ZIP with files inside a sub-folder? @Tristramg @antoine-de Do you need this functionality?

antoine-de: yes it's a very handy feature as it's a very common case for GTFS in the wild.

I don't really see the point of removing the reading from zip functionality :thinking: If it's to remove the dependency why don't you just feature gate it?

And why do you need to remove the `ZipHandler` ? it's very nice to be able to read gtfs without having to write it on files no?

woshilapin: `ZipHandler` (and `FileHandler` in general) is very limited. For example, we couldn't use it for a `zip` format with a folder structure. And I did try to make `FileHandler` evolves to something more generic, able to navigate through folders... without success (return references to path of the ZIP **and** keeping a mutable reference on `ZipArchive`...). So we're in the process of totally removing `FileHandler` (looks like a half-baked functionality and difficult to maintain) from `transit_model` at the moment and the current PR is only a step towards this direction.

Secondly, zipping and unzipping is really not the core business of `transit_model`. Sure it could be flagged as a _feature_ but really, we want to let this responsibility to the binary that uses this library instead.

All that said, I was asking the question because the current PR will still support ZIP in `gtfs2netexfr`... but by reducing the kind of ZIP (notably the one with weird sub-folders). So I'll try to add these missing cases to my implementation of `read_from_zip()` for now. Eventually, this code will likely move into `gtfs2netexfr/src/main.rs`.

antoine-de: a gtfs needs to be ziped in [the spec](https://gtfs.org/reference/static#file-requirements). I know you don't use them this way, but if you remove this functionality you limit the number of external people that can use transit-model.

Anyway I think it's a pity but that's your call.

woshilapin: You've got a good point about citing the specifications. Thank you.

In any case, we're probably moving away from `FileHandler`... but we're now thinking that maybe the API `transit_model::gtfs::read()` should accept a ZIP, nothing else (a folder's path is not really standard). I'm putting back this PR in Draft for now. Still wondering though if it is the responsability of `transit_model` to handle non-valid ZIP GTFS... I'm balanced between applying the [robustness principle](https://www.wikiwand.com/en/Robustness_principle) inside `transit_model` or outside of it (in the binaries that are using `transit_model`).

woshilapin:
Small summary of all the information I have been gathering so far.
- `FileHandler` is complex to maintain
- Having a generic parameter `&mut FileHandler` on most of our functions is affecting compilation times and makes code more complex. But can we accept different formats (folder's `Path` or `ZipArchive`) without being generic? And without losing performance at runtime (if we need to convert from one to another...)?
- GTFS format is a ZIP (see [specification](https://gtfs.org/reference/static/#file-requirements)) so `transit_model` must accept ZIP as an input
- ZIP might have different formats (all files at the root, all files in a subdirectory, etc.), the library (not the binary) must be able to accept all these formats (see [robustness principle](https://www.wikiwand.com/en/Robustness_principle))
- `read_from_zip()` and `read_from_path()` are useful (we could have a generic `read()` that try to be smart and choose but in case it's not smart enough, let's keep the two other exposed)

woshilapin:
I'm closing this PR for all the reasons explained above

PR #715, opened 2020-10-29
Manage calendar no file handler

woshilapin:
This PR depends on #714 which needs to be merged first.

In this PR, `transit_model::calendars::manage_calendars()` is refactored to take a `AsRef<Path>` as input parameter instead of a `FileHandler`.

In order to make it work, I introduced a `FileHandler::as_path_buf()` method for now. I hope this will help us during the work of removing `FileHandler` bit-by-bit, especially in functions that receive a `FileHandler` but need to call sub-functions with a `AsRef<Path>` (which is the case now in `transit_model::gtfs::read()` that calls `transit_model::calendars::manage_calendars()`). In the end, `FileHandler` will disappear anyway so this is only a temporary helper for the transition.

ref: ND-956.

### DIFF
--- a/src/calendars.rs
+++ b/src/calendars.rs
@@ -19,7 +19,6 @@
 
 use crate::model::Collections;
 use crate::objects::{self, Date, ExceptionType};
-use crate::read_utils::FileHandler;
 use crate::utils::*;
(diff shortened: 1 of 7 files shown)


woshilapin:
I'm closing this PR (see #714 for details).

PR #724, opened 2020-12-22
[feature] read gtfs route as line

patochectp:
ref: ND-1090

### DIFF
--- a/examples/gtfs_reader.rs
+++ b/examples/gtfs_reader.rs
@@ -13,22 +13,10 @@
 // along with this program. If not, see <https://www.gnu.org/licenses/>
 
 use serde_json::json;
-use std::collections::BTreeMap;
-use transit_model::{
-    gtfs,
-    objects::{Contributor, Dataset},
-    Result,
-};
+use transit_model::{gtfs, Result};
 
 fn run() -> Result<()> {
-    let configuration: gtfs::Configuration = gtfs::Configuration {
-        contributor: Contributor::default(),
-        dataset: Dataset::default(),
-        feed_infos: BTreeMap::new(),
-        prefix_conf: None,
-        on_demand_transport: false,
-        on_demand_transport_comment: None,
-    };
+    let configuration = gtfs::Configuration::default();
     // read GTFS from current directory
     let objects = gtfs::read_from_path(".", configuration)?;
     // output internal model as JSON
--- a/gtfs2ntfs/src/main.rs
+++ b/gtfs2ntfs/src/main.rs
@@ -58,6 +58,12 @@ struct Opt {
     #[structopt(long = "odt-comment")]
     odt_comment: Option<String>,
 
+    /// If true, each GTFS `Route` will generate a different `Line`.
+    /// Else we group the routes by `agency_id` and `route_short_name`
+    /// (or `route_long_name` if the short name is empty) and create a `Line` for each group.
+    #[structopt(long = "read-as-line")]
+    read_as_line: bool,
+
     /// Current datetime.
     #[structopt(
         short = "x",
@@ -118,6 +124,7 @@ fn run(opt: Opt) -> Result<()> {
         prefix_conf: Some(prefix_conf),
         on_demand_transport: opt.odt,
         on_demand_transport_comment: opt.odt_comment,
+        read_as_line: opt.read_as_line,
     };
 
     let model = if opt.input.is_file() {
--- a/src/gtfs/mod.rs
+++ b/src/gtfs/mod.rs
@@ -253,6 +253,7 @@ struct Shape {
 }
 
 ///parameters consolidation
+#[derive(Default)]
 pub struct Configuration {
     /// The Contributor providing the Dataset
     pub contributor: Contributor,
@@ -266,6 +267,10 @@ pub struct Configuration {
     pub on_demand_transport: bool,
     /// on demand transport comment template
     pub on_demand_transport_comment: Option<String>,
+    /// If true, each GTFS `Route` will generate a different `Line`.
+    /// Else we group the routes by `agency_id` and `route_short_name`
+    /// (or `route_long_name` if the short name is empty) and create a `Line` for each group.
+    pub read_as_line: bool,
 }
 
 fn read<H>(file_handler: &mut H, configuration: Configuration) -> Result<Model>
@@ -282,6 +287,7 @@ where
         prefix_conf,
         on_demand_transport,
         on_demand_transport_comment,
+        read_as_line,
     } = configuration;
 
     manage_calendars(file_handler, &mut collections)?;
@@ -303,7 +309,7 @@ where
 
     read::manage_shapes(&mut collections, file_handler)?;
 
-    read::read_routes(file_handler, &mut collections)?;
+    read::read_routes(file_handler, &mut collections, read_as_line)?;
     collections.equipments = CollectionWithId::new(equipments.into_equipments())?;
     read::manage_stop_times(
         &mut collections,
--- a/src/gtfs/read.rs
+++ b/src/gtfs/read.rs
@@ -267,14 +267,18 @@ impl<'de> ::serde::Deserialize<'de> for RouteType {
 impl_id!(Route);
 
 impl Route {
-    fn get_line_key(&self) -> (Option<String>, String) {
+    fn generate_line_key(
+        &self,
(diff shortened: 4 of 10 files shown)


src/gtfs/read.rs
```diff
@@ -267,14 +268,18 @@ impl<'de> ::serde::Deserialize<'de> for RouteType {
...
             self.short_name.clone()
         } else {
             self.long_name.clone()
         };
-
-        (self.agency_id.clone(), name)
+        let key = if read_as_line {
+            Uuid::new_v4()
+        } else {
+            Uuid::NAMESPACE_OID
```
antoine-de, line 280: what is this used for? it's only to always have the same value no? If so, maybe it's better to use an `Option<Uuid>` and set it to `None` no?

src/gtfs/mod.rs
```diff
@@ -266,6 +266,8 @@ pub struct Configuration {
     pub on_demand_transport: bool,
     /// on demand transport comment template
     pub on_demand_transport_comment: Option<String>,
+    /// each route described is a line
+    pub read_as_line: bool,
```
antoine-de, line 270: I think it might be worth it to explain how the routes are created otherwise

examples/gtfs_reader.rs
```diff
@@ -28,6 +28,7 @@ fn run() -> Result<()> {
         prefix_conf: None,
         on_demand_transport: false,
         on_demand_transport_comment: None,
+        read_as_line: false,
```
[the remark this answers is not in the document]
patochectp: I don't like default impl but why not.

gtfs2ntfs/src/main.rs
```diff
@@ -58,6 +58,19 @@ struct Opt {
     #[structopt(long = "odt-comment")]
     odt_comment: Option<String>,
 
+    /// If we don’t have the flag activated,
```
antoine-de, line 61: honestly this description is great, but here is an attempt at another way to describe this (more to be sure to understand correctly, feel free to keep it that way if you prefer)

```
If true, each GTFS `Route` will generate a different `Line`.
Else we group the routes by `agency_id` and `route_short_name` (or `route_long_name` if the short name is empty) and create a `Line` for each group.
```

gtfs2ntfs/src/main.rs
```diff
@@ -58,6 +58,19 @@ struct Opt {
...
+    ///         they have the same agency_id: the 2 routes are merged in the same line
+    ///         they have different agency_id: we will have 2 lines in output
+    ///     2 GTFS routes with an empty route_short_name and the same route_long_name
+    ///         they have the same agency_id: the 2 routes are merged in the same line
+    ///         they have different agency_id: we will have 2 lines in output
+    /// If the flag is activated:
+    ///     2 GTFS routes with the same route_short_name (and even agency_id) will generate 2 lines
+    ///     2 GTFS routes with the same route_long_name (and even agency_id) will generate 2 lines
+    #[structopt(long = "read-as-line")]
+    read_as_line: bool,
```
antoine-de, line 72: wouldn't it be better to explicitly set the default value to false in structopt too?

antoine-de (approved):
great! :heart: 

I added 2 very minor comments feel free to skip them

gtfs2ntfs/src/main.rs
patochectp: default value is meaningless for bool

antoine-de: oh right I forgot it was a flag sorry

antoine-de (approved):

src/gtfs/read.rs
```diff
@@ -267,14 +268,18 @@ impl<'de> ::serde::Deserialize<'de> for RouteType {
...
+    fn get_line_key(&self, read_as_line: bool) -> (Option<String>, String, Option<Uuid>) {
         let name = if self.short_name != "" {
             self.short_name.clone()
         } else {
             self.long_name.clone()
         };
-
-        (self.agency_id.clone(), name)
+        let key = if read_as_line {
+            Some(Uuid::new_v4())
```
datanel, line 278: Is it necessary to use uuid here? Can't you use the index of a route in `gtfs_routes`?

src/gtfs/read.rs
```diff
@@ -3297,4 +3307,103 @@ mod tests {
             );
         });
     }
+
+    #[test]
+    fn read_gtfs_routes_as_line() {
```
woshilapin, line 3312: I do not see any generated-`UUID` in this test. Do we test this?

woshilapin (changes requested):
A few missing things:
- Please bump the version (minor) of `transit_model`
- The specification needs to be updated
https://github.com/CanalTP/transit_model/blob/master/documentation/gtfs_to_ntfs_specs.md

src/gtfs/read.rs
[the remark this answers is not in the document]
patochectp: That’s what we want. We want to deactivate the consolidation of routes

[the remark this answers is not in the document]
woshilapin: Yet with the name of the function, I'd expect a function that always return the same thing. Maybe it's only a problem of the function's name. How about something like `fn generate_line_key()` instead?

src/gtfs/read.rs
```diff
@@ -3297,4 +3307,103 @@ mod tests {
...
+            collections.datasets = CollectionWithId::new(vec![dataset]).unwrap();
+            super::read_routes(&mut handler, &mut collections, read_as_line).unwrap();
+            collections
+        }
+
+        test_in_tmp_dir(|path| {
+            // read route as route
+            let collections = get_collection(path, false);
+            // read route as line
+            let collections_as_line = get_collection(path, true);
```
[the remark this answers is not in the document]
patochectp: It is to highlight the difference between read gtfs as line instead of read gtfs as route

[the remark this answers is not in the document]
woshilapin: You can group them in a `mod` and have tests name reflect this difference. At least, I find it very weird to do twice every line. Maybe this can be a parameterized function? Something like [this _crate_](https://crates.io/crates/parameterized) is doing (not necessarily using this _crate_ or creating macros but something like this)...

src/gtfs/read.rs
patochectp: This requires passing an additional parameter

datanel: if you use Idx instead of Uuid

```rust 
type MapLineRoutes<'a> = BTreeMap<(Option<String>, String, Option<Idx<Route>>), Vec<&'a Route>>;

fn map_line_routes<'a>(
    gtfs_routes: &'a CollectionWithId<Route>,
    gtfs_trips: &[Trip],
    read_as_line: bool,
) -> MapLineRoutes<'a> {
    let mut map = BTreeMap::new();
    for (idx, r) in gtfs_routes.iter().filter(|(_, r)| {
        if !gtfs_trips.iter().any(|t| t.route_id == r.id) {
            warn!("Couldn't find trips for route_id {}", r.id);
            return false;
        }
        true
    }) {
        map.entry(r.get_line_key(read_as_line, idx))
            .or_insert_with(Vec::new)
            .push(r);
    }
    map
}
```

I should work, what do you think?

patochectp: doone in f33d5556747a5738d0d81909c89afd2163969329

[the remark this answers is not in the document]
patochectp: done in f33d5556747a5738d0d81909c89afd2163969329

src/gtfs/mod.rs
```diff
@@ -266,6 +266,24 @@ pub struct Configuration {
     pub on_demand_transport: bool,
     /// on demand transport comment template
     pub on_demand_transport_comment: Option<String>,
+    /// If true, each GTFS `Route` will generate a different `Line`.
+    /// Else we group the routes by `agency_id` and `route_short_name`
+    /// (or `route_long_name` if the short name is empty) and create a `Line` for each group.
+    pub read_as_line: bool,
+}
+
+impl Default for Configuration {
```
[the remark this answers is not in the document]
patochectp: done in f33d5556747a5738d0d81909c89afd2163969329

datanel (approved):

woshilapin (dismissed):

datanel (approved):

PR #725, opened 2020-12-23
[feature] remove transfers at the same stop point in gtfs export

patochectp:
ref: ND-1091

### DIFF
--- a/src/gtfs/write.rs
+++ b/src/gtfs/write.rs
@@ -39,8 +39,10 @@ pub fn write_transfers(path: &path::Path, transfers: &Collection<NtfsTransfer>) 
     let mut wtr =
         csv::Writer::from_path(&path).with_context(|_| format!("Error reading {:?}", path))?;
     for t in transfers.values() {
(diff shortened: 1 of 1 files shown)


src/gtfs/write.rs
```diff
@@ -39,8 +39,10 @@ pub fn write_transfers(path: &path::Path, transfers: &Collection<NtfsTransfer>)
     let mut wtr =
         csv::Writer::from_path(&path).with_context(|_| format!("Error reading {:?}", path))?;
     for t in transfers.values() {
-        wtr.serialize(Transfer::from(t))
-            .with_context(|_| format!("Error reading {:?}", path))?;
+        if t.from_stop_id != t.to_stop_id {
```
antoine-de, line 42: I don’t really remember this stuff, but don’t you want to remove those transfer at the model loading?

patochectp: We want to keep them in ntfs, navitia uses them

antoine-de (approved):

datanel (approved):

PR #726, opened 2021-01-04
[tech] Bump to 0.33.0

ArnaudOggy:


Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 authors = ["Kisio Digital <team.coretools@kisio.com>", "Guillaume Pinot <texitoi@texitoi.eu>"]
 name = "transit_model"
-version = "0.32.2"
+version = "0.32.3"
```
woshilapin, line 4: I would do a `0.33.0` since the only PR you'll be including is changing the behavior of the GTFS export.

woshilapin (approved):

Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 authors = ["Kisio Digital <team.coretools@kisio.com>", "Guillaume Pinot <texitoi@texitoi.eu>"]
 name = "transit_model"
-version = "0.32.2"
+version = "0.33.0"
```
woshilapin, line 4: But you need to update all the _subcrates_ now :stuck_out_tongue:

woshilapin (changes requested):

Cargo.toml
ArnaudOggy: Oops :hand_over_mouth: 
Done 822464f9190dcccb79177d049c82f507bcf4455c

woshilapin (approved):

PR #728, opened 2021-01-04
[fix] removal of trips containing stop zone

patochectp:
ref: ND-865

### DIFF
--- a/src/gtfs/mod.rs
+++ b/src/gtfs/mod.rs
@@ -416,15 +416,20 @@ fn remove_stop_zones(model: Model) -> Result<Collections> {
     collections
         .stop_points
         .retain(|sp| sp.stop_type != StopType::Zone);
-    let stop_point_ids: Vec<&Idx<StopPoint>> =
-        collections.stop_points.get_id_to_idx().values().collect();
-    let mut vjs = collections.vehicle_journeys.take();
-    for vj in &mut vjs {
+
+    let stop_point_ids: Vec<Idx<StopPoint>> = collections
+        .stop_points
+        .get_id_to_idx()
+        .values()
+        .copied()
+        .collect();
+
+    collections.vehicle_journeys.retain(|vj| {
         vj.stop_times
-            .retain(|st| stop_point_ids.contains(&&st.stop_point_idx));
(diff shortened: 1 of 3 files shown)


src/gtfs/mod.rs
```diff
@@ -412,13 +412,18 @@ fn remove_stop_zones(model: Model) -> Result<Collections> {
...
+        .vehicle_journeys
+        .into_iter()
+        .filter(|vj| {
+            vj.stop_times
+                .iter()
+                .all(|st| stop_point_ids.contains(&&st.stop_point_idx))
+        })
+        .collect();
+
+    collections.vehicle_journeys = CollectionWithId::new(vjs_to_keep)?;
```
woshilapin, line 426: You can directly use `.retain()` function here keeping the same closure (and removing a `&` if you made the modification in the comment above :wink:).

```suggestion
    collections.vehicle_journeys.retain(|vj| {
        vj.stop_times
            .iter()
            .all(|st| stop_point_ids.contains(&st.stop_point_idx))
    });
```

patochectp: done

Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 authors = ["Kisio Digital <team.coretools@kisio.com>", "Guillaume Pinot <texitoi@texitoi.eu>"]
 name = "transit_model"
-version = "0.32.3"
+version = "0.32.4"
```
woshilapin, line 4: No need for a `0.32.4` since `0.32.3` has not yet been released. However, since this PR is changing the behavior of the GTFS export, I'm wondering if it should not be a `0.33.0` instead.

ArnaudOggy: @patochectp @woshilapin It's already a `0.33.0` in Cargo.

woshilapin: Yes, indeed, because we did merge #726 before waiting for the current PR. Then `0.33.0` it is I guess!

ArnaudOggy (approved):

ArnaudOggy (approved):

woshilapin (approved):

PR #733, opened 2021-01-26
gtfs2ntfs: read route_desc into comments

Tristramg:
Internal issue: https://jira.kisio.org/browse/ND-1075

### DIFF
--- a/src/gtfs/read.rs
+++ b/src/gtfs/read.rs
@@ -611,7 +611,7 @@ where
     Ok((networks, companies))
 }
 
-fn get_stop_comment(stop: &Stop) -> Option<objects::Comment> {
+fn generate_stop_comment(stop: &Stop) -> Option<objects::Comment> {
     stop.desc.as_ref().map(|desc| objects::Comment {
         id: "stop:".to_string() + &stop.id,
         comment_type: objects::CommentType::Information,
@@ -621,7 +621,7 @@ fn get_stop_comment(stop: &Stop) -> Option<objects::Comment> {
     })
 }
 
-fn get_route_comment(route: &Route) -> Option<objects::Comment> {
+fn generate_route_comment(route: &Route) -> Option<objects::Comment> {
     route.desc.as_ref().map(|desc| objects::Comment {
         id: "route:".to_string() + &route.id,
         comment_type: objects::CommentType::Information,
@@ -755,7 +755,7 @@ where
     let mut stop_locations = vec![];
     for stop in gtfs_stops {
         let mut comment_links = CommentLinksT::default();
-        if let Some(comment) = get_stop_comment(&stop) {
+        if let Some(comment) = generate_stop_comment(&stop) {
             comment_links.insert(comment.id.to_string());
             comments
                 .push(comment)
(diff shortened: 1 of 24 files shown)


src/gtfs/read.rs
```diff
@@ -631,6 +631,26 @@ fn manage_comment_from_stop(
     comment_links
 }
 
+fn manage_comment_from_route(
+    comments: &mut CollectionWithId<objects::Comment>,
+    route: &Route,
+) -> CommentLinksT {
```
woshilapin, line 637: I know we have a lot of examples of non-pure functions in the code but do you think it'd be possible to have a pure-function here that may only return a `Option<Comment>`? Something like this.

```rust
fn get_route_comment(route: &Route) -> Option<Comment> {
   // omitted
}
```

src/gtfs/read.rs
```diff
@@ -631,6 +631,26 @@ fn manage_comment_from_stop(
...
+    if let Some(desc) = &route.desc {
+        let comment_id = "route:".to_string() + &route.id;
+        let comment = objects::Comment {
+            id: comment_id.clone(),
+            comment_type: objects::CommentType::Information,
+            label: None,
+            name: desc.to_string(),
+            url: None,
+        };
+        comments.push(comment).unwrap();
```
woshilapin, line 648: Please try to avoid `.unwrap()` or at least comment about why this will never panic.

woshilapin (commented):
Wasn't it possible to only add a new comment on existing features instead of adding an entire new set of fixtures?

src/gtfs/read.rs
Tristramg: I used an expect, it doesn’t seem common practice, but it does both the work of documenting and unwrapping. And if the unpossible happens, we’ll have some hint.

src/gtfs/read.rs
Tristramg: Way nicer!
I’m adding a commit to handle the same way the stop comments (I just copied-paste…)

src/gtfs/read.rs
woshilapin: Yes, `.expect()` or document is good (actually, I should push for the use of `.expect()` now so thanks :smile:)

Tristramg:
Concerning the fixtures, many tests rely on the output of minimal. This would require small changes a bit everywhere. Do you prefer that approach?

woshilapin:
> Concerning the fixtures, many tests rely on the output of minimal. This would require small changes a bit everywhere. Do you prefer that approach?

I see, if the alternative is to modify `minimal`, then I guess you made it right. Keep what you've done then.

woshilapin (dismissed):

Tristramg:
I added a commit which applies the same behaviour for stop comments.
At this point I’m wondering if we shouldn’t add a trait WithDescription to the GTFS model, but I guess I should leave the rabbit hole for now :(

woshilapin (approved):
I'm just wondering if `generate_route_comment` wouldn't be more accurate as the function name but apart from that, this PR is good to go.

Tristramg:
Thanks, I renamed it

woshilapin (approved):

PR #735, opened 2021-01-28
gtfs: don’t fail if there are bad references in stop_times.txt

Tristramg:
ref: ND-1101


src/gtfs/read.rs
```diff
@@ -400,30 +400,26 @@ where
...
-        if let Some(headsign) = headsign {
-            headsigns.insert(
-                (stop_time.trip_id.clone(), stop_time.stop_sequence),
-                headsign,
-            );
-        }
+        if let Some(vj_idx) = collections.vehicle_journeys.get_idx(&stop_time.trip_id) {
+            // consume the stop headsign
+            let headsign = std::mem::replace(&mut stop_time.stop_headsign, None);
+            if let Some(headsign) = headsign {
```
woshilapin, line 406: I know it was the previous code but as you're modifying this piece of code, I believe `Option::take()` could be used instead of `std::mem::replace()`. And possibly, since it's going to be shorter to write, you might inline it in the `if` below.

```suggestion
            if let Some(headsign) = stop_time.stop_headsign.take() {
```

woshilapin (approved):

antoine-de (approved):

woshilapin (approved):

PR #738, opened 2021-01-29
[tech] move the enhance co2 function out of the model

antoine-de:
This is the first step to split `model.rs` into several files to improve maintanability.

To ease review I think you can review this and if it's ok we can merge and do separate PR

### DIFF
--- a/src/enhancers/enhance_co2.rs
+++ b/src/enhancers/enhance_co2.rs
@@ -1,0 +1,136 @@
+use crate::{
+    model::{self, Collections},
+    objects::PhysicalMode,
+};
+use typed_index_collection::CollectionWithId;
+
+lazy_static::lazy_static! {
+    static ref CO2_EMISSIONS: std::collections::HashMap<&'static str, f32> = {
+        let mut modes_map = std::collections::HashMap::new();
+        modes_map.insert(model::AIR_PHYSICAL_MODE, 144.6f32);
+        modes_map.insert(model::BIKE_PHYSICAL_MODE, 0f32);
+        modes_map.insert(model::BIKE_SHARING_SERVICE_PHYSICAL_MODE, 0f32);
+        // Unknown value
+        // modes_map.insert(model::"Boat", 0.0f32);
+        modes_map.insert(model::BUS_PHYSICAL_MODE, 132f32);
+        modes_map.insert(model::BUS_RAPID_TRANSIT_PHYSICAL_MODE, 84f32);
+        modes_map.insert(model::CAR_PHYSICAL_MODE, 184f32);
+        modes_map.insert(model::COACH_PHYSICAL_MODE, 171f32);
+        modes_map.insert(model::FERRY_PHYSICAL_MODE, 279f32);
+        modes_map.insert(model::FUNICULAR_PHYSICAL_MODE, 3f32);
+        modes_map.insert(model::LOCAL_TRAIN_PHYSICAL_MODE, 30.7f32);
+        modes_map.insert(model::LONG_DISTANCE_TRAIN_PHYSICAL_MODE, 3.4f32);
+        modes_map.insert(model::METRO_PHYSICAL_MODE, 3f32);
+        modes_map.insert(model::RAPID_TRANSIT_PHYSICAL_MODE, 6.2f32);
+        // Unknown value
+        // modes_map.insert(model::RailShuttle, 0.0f32);
+        // Unknown value
+        // modes_map.insert(model::Shuttle, 0.0f32);
+        // Unknown value
+        // modes_map.insert(model::SuspendedCableCar, 0.0f32);
+        modes_map.insert(model::TAXI_PHYSICAL_MODE, 184f32);
+        modes_map.insert(model::TRAIN_PHYSICAL_MODE, 11.9f32);
+        modes_map.insert(model::TRAMWAY_PHYSICAL_MODE, 4f32);
+        modes_map
+    };
+}
+
+/// Physical mode should contains CO2 emissions. If the values are not present
+/// in the NTFS, some default values will be used.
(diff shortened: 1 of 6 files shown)


src/enhancers/mod.rs
```diff
@@ -0,0 +1,5 @@
+//! This module contains various functions that enhance / cleanup `Collections`
+
+mod enhance_co2;
+
+pub(crate) use enhance_co2::enhance_with_co2;
```
woshilapin, line 5: Since it's part of the module `enhancers`, maybe the function can be just named `with_co2()` now? And you would use it as `enhancer::with_co2()`? What do you think?

src/enhancers/enhance_co2.rs
```diff
@@ -0,0 +1,136 @@
...
+        modes_map.insert(model::TAXI_PHYSICAL_MODE, 184f32);
+        modes_map.insert(model::TRAIN_PHYSICAL_MODE, 11.9f32);
+        modes_map.insert(model::TRAMWAY_PHYSICAL_MODE, 4f32);
+        modes_map
+    };
+}
+
+/// Physical mode should contains CO2 emissions. If the values are not present
+/// in the NTFS, some default values will be used.
+pub fn enhance_with_co2(collections: &mut Collections) {
```
woshilapin, line 40: I'm never sure which is better between

```rust
pub fn enhance_with_co2(collections: &mut Collections) { ... }
```

and 

```rust
pub fn enhance_with_co2(collections: Collections) -> Collections { ... }
```

The benefit of the second version is that the function is pure.

woshilapin (approved):

src/enhancers/enhance_co2.rs
antoine-de: sure but I wanted to have minimal changes and I'm not really convinced one is really better than the other :thinking:

src/enhancers/enhance_co2.rs
```diff
@@ -0,0 +1,136 @@
...
+        // modes_map.insert(model::SuspendedCableCar, 0.0f32);
+        modes_map.insert(model::TAXI_PHYSICAL_MODE, 184f32);
+        modes_map.insert(model::TRAIN_PHYSICAL_MODE, 11.9f32);
+        modes_map.insert(model::TRAMWAY_PHYSICAL_MODE, 4f32);
+        modes_map
+    };
+}
+
+/// Physical mode should contains CO2 emissions. If the values are not present
+/// in the NTFS, some default values will be used.
```
pbougue, line 39:
```suggestion
/// Fills the missing values for CO2 emissions of Physical mode. If the values are not present
/// in the NTFS, some default values will be used.
```

pbougue (approved):
It looks OK, although I think @woshilapin and @datanel should have a look at the new storage.
Detail (if I'm not mistaken): The order in which enhancements/cleanups are applied matters for performance and functionnality, so we have to pay attention to have this readable and accessible I guess. Maybe just pay attention to enhancers' rust-comments is doing most of the job.

src/enhancers/mod.rs
antoine-de: I'm not convinced by the `enhancer` name so I don't know :grin:
I plan to also move the `sanitize` method into it so it would be `enhancer::sanitize`? (well, having written it I don't really think it's a problem, what do you think?)

woshilapin: Maybe we can find a better name for the module? `filters`? `refine`? `process`? Since this module is part of the `model` module, I believe, it's clear that it is applied on `Model`.

I kind of like `process` or even `refine`...

src/enhancers/enhance_co2.rs
woshilapin: And you were probably right to do minimal changes at this point. I guess I took the opportunity to review it since the code was changed. Anyone else has an opinion on the topic?

src/enhancers/mod.rs
pbougue: Not fan of 'process' (too blurry) or 'filter' (there is more than that into this), but 'refine' or 'enhance' is OK to me.

antoine-de: so `refine::enhance_with_co2` and `refine::sanitize`? not mind blown but fine for me

woshilapin: Well that would be `refine::with_co2` but really, `enhancer`, `enhance`, `refiner` or `refine` are all good for me, my first comment was more on removing `enhance_` from the name of the function since we now have the module's name for that...

pbougue: Then `fill_co2`?

antoine-de: `enhancers::fill_co2` seems good for me

src/model.rs
```diff
@@ -1448,7 +1389,7 @@ impl Model {
         )?;
 
         c.update_stop_area_coords();
-        c.enhance_with_co2();
+        fill_co2(&mut c);
```
woshilapin, line 1392: I thought you would keep this here.
```suggestion
        enhancers::fill_co2(&mut c);
```

woshilapin (approved):

ArnaudOggy (approved):

src/model.rs
antoine-de: you're right, I changed it (but did not took the commit to also change the import)

woshilapin (dismissed):

ArnaudOggy (dismissed):

pbougue (approved):

woshilapin:
You need to bump the `Cargo.toml` to 0.33.2 by the way since 0.33.1 has been released.

pbougue (approved):

woshilapin (approved):

ArnaudOggy (approved):

datanel (approved):

Issue #737, opened 2021-01-29
Add more flexibility to read inputs


antoine-de:
previously there was a method `gtfs::read_from_url` that was quite handy to read a gtfs from the network (and is used by https://github.com/etalab/transpo-rt/) but the dependency to `reqwest` was slowing down the compilation time of transit_model without being directly used by it.

It makes however the upgrade to the newest version of transit_model impossible for transpo-rt (because `FileHandler` and `gtfs::read` are not public).

I see several way to make this possible

## A: put back the reqwest dependency
:+1: easiest solution
:+1: can be used by kisio too
:-1: build time

## B: put back the reqwest dependency behind a feature
:+1: easy solution
:+1: can be used by kisio too
:-1: :-1: build time for ci (if ci is run for both feature, else :+1: )
:+1:  build time for other crates using transit_model without feature (like tartare-tools) 

## C: expose a way to do the `reqwest` call externally
I think we could expose a `gtfs::from_ziped_read` (not fan of the name if you have suggestion)
```rust
mod gtfs {

pub fn from_ziped_read<P: AsRef<Path>>(
    reader: R,
    source_name: &str,  // it's a bother but `source_name` is useful here to have nicer error msg
)-> Result<Model> 
where
    R: std::io::Seek + std::io::Read {
    let mut file_handle = read_utils::ZipHandler::new(reader, &source_name)?;
    read(&mut file_handle, config_path, prefix)
}
}
```
and in another crate (like [transpo-rt](https://github.com/etalab/transpo-rt/))
```rust
fn read_url(url: &str) -> Result<std::io::Cursor<Vec<u8>>> {
    let mut res = reqwest::get(url)?;
    let mut body = Vec::new();
    res.read_to_end(&mut body)?;
    Ok(std::io::Cursor::new(body))
}

pub fn main() -> Result<()> {
    let url = "http://pouet.zip";
    let r = read_url(&url)?;
    let model = gtfs::from_ziped_read(r, &url)?;
}
```
(it could be added to transit_model doc to make it easier)
:-1: more code but not too complex
:+1: can be used for other stuff (like a POC for WASM done by @datanel )
:+1: build time

what do you think about this? I'd personally vote for B + C :stuck_out_tongue_winking_eye:

woshilapin:
I would say yes for option C, even if the `source_name` is pretty weird but that is mostly exposing a weirdness of `ZipHandler::new()` (but now, this weirdness will be publicly exposed :cry:).

[...]

Issue #739, opened 2021-01-29
[tech] Easier hello world


antoine-de:
The current API to load a model (from a GTFS or NTFS) is honestly quite straightforward and easy. I think we can still improve it a bit and I think a very neat library example can ease adoption.

For the moment to load a `Model` from a GTFS (and I think it can be applied to the other format) you need to do:
```rust
let model = gtfs::read_from_path(input_dir, gtfs::Configuration::default())?;
```

I think here the `gtfs::Configuration::default()` could be hidden.
Moreover in most uses of `gtfs` (like all transit_model's executables) the model is initialized like:
```rust
    let model = if opt.input.is_file() {
        transit_model::gtfs::read_from_zip(opt.input, configuration)?
    } else if opt.input.is_dir() {
        transit_model::gtfs::read_from_path(opt.input, configuration)?
    } else {
        bail!("Invalid input data: must be an existing directory or a ZIP archive");
    };
```
since it's a common use, it could be done for the users.

To ease the canonical way to create a model I think we can:
 * add a magic function that can read a zip or path (and urls in the futur ? :heart_eyes_cat: )
 * add a builder pattern to hide the configuration

### First the usage:

The easiest
```rust
let model = gtfs::read("./pouet")?;
// or
let model = gtfs::read("./pouet.zip")?;
// or later
let model = gtfs::read("http://pouet.zip")?; :heart_eyes_cat: 
```
or directly
```rust
let model = gtfs::from_zip("./pouet.zip")?;
```
and with a custom config
```rust
let model = gtfs::Reader::default()
  .prefix("IDFM")
  .on_demand_transport(true)
  .read("./pouet")?;
```

### And the gtfs module

```rust
mod gtfs {
  pub fn from_zip(p: impl AsRef<Path>) -> Result<Model> {
     gtfs::Reader::default().from_zip(p)
  }
  pub fn from_path(path: impl AsRef<Path>) -> Result<Model> {…}
  pub fn read(path: impl AsRef<Path>) -> Result<Model> {…} // read from a zip or a path
  pub fn from_reader(path: impl AsRef<Path>) -> Result<Model> {…} // see https://github.com/CanalTP/transit_model/issues/737
  
  pub struct Reader {
    configuration: transit_model::gtfs::Configuration
  }
  impl Reader {  
    // The 'builder' functions that 'eats' the `Reader`
[...]

PR #740, opened 2021-01-29
add a gtfs::from_read method

antoine-de:
closes https://github.com/CanalTP/transit_model/issues/737

Add a `gtfs::from_read` method to be able to read gtfs from various
sources (like from the network).

this makes it possible to read a gtfs like
```rust
let url = "http://some_url/gtfs.zip";
let resp = reqwest::blocking::get(url)?; // or async call
let data = std::io::Cursor::new(resp.bytes()?.to_vec());
let model = transit_model::gtfs::from_read(data, &url, configuration)?;
```

### DIFF
--- a/src/gtfs/mod.rs
+++ b/src/gtfs/mod.rs
@@ -333,13 +333,7 @@ where
 /// Imports a `Model` from the [GTFS](https://gtfs.org/reference/static)
 /// files in the `path` directory.
 ///
-/// The `config_path` argument allows you to give a path to a file
-/// containing a json representing the contributor and dataset used
-/// for this GTFS. If not given, default values will be created.
-///
-/// The `prefix` argument is a string that will be prepended to every
-/// identifiers, allowing to namespace the dataset. By default, no
-/// prefix will be added to the identifiers.
+/// The `Configuration` is used to control various parameters during the import.
 pub fn read_from_path<P: AsRef<Path>>(p: P, configuration: Configuration) -> Result<Model> {
     let mut file_handle = read_utils::PathFileHandler::new(p.as_ref().to_path_buf());
     read(&mut file_handle, configuration)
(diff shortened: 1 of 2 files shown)


src/read_utils.rs
```diff
@@ -142,24 +142,26 @@ impl<'a, P: AsRef<Path>> FileHandler for &'a mut PathFileHandler<P> {
 /// Unlike ZipArchive, it gives access to a file by its name not regarding its path in the ZipArchive
 /// It thus cannot be correct if there are 2 files with the same name in the archive,
 /// but for transport data if will make it possible to handle a zip with a sub directory
-pub(crate) struct ZipHandler {
-    archive: zip::ZipArchive<File>,
+pub(crate) struct ZipHandler<R: std::io::Seek + std::io::Read> {
```
woshilapin, line 145: I'm never sure which strategy to adopt but, do you need the bound here? I believe the bounds are only needed on the `impl` of `fn new()` and I know some `std` implementations put the bounds only were they're strictly needed. Any opinion on such methods?

src/read_utils.rs
```diff
@@ -172,7 +174,10 @@ impl ZipHandler {
     }
 }
 
-impl<'a> FileHandler for &'a mut ZipHandler {
+impl<'a, R> FileHandler for &'a mut ZipHandler<R>
+where
+    R: std::io::Seek + std::io::Read,
```
woshilapin, line 179: You're using multiple times `std::io::Seek` and `std::io::Read` so maybe they can be imported in the prelude of the file with `use` clauses (or maybe just `use std::io;` at least)?

woshilapin (commented):
I'm actually wondering if `read_from_zip()` should just go away and be replaced by your new method, and maybe rename your new method `from_zip()`? Actually, `read_from_zip()` doesn't do much more and having an external API that takes a `Read` and another that takes a `AsRef<Path>` feels almost like a duplicate.

About the name of the function, I know we discussed using `from_read()` at first, but now that I think about it, it's pretty obvious that it takes a `Read` from the signature of the function. However, it's not obvious that the `Read` should be a ZIP. That's why I'm proposing `from_zip()` instead. What do you think about it?

src/read_utils.rs
antoine-de: The bound is needed in `ZipArchive` no?

woshilapin: Oops, you're right, forget about it :smile:

src/read_utils.rs
antoine-de: I'm not a fan of import I find it important to know where the types are from but since those are quite known Trait I guess we can import them.

antoine-de:
from_zip is more straightforward since you only need to give the path once. In the spirit of simplifying the hello world I'd prefer to keep it

src/gtfs/mod.rs
```diff
@@ -348,15 +342,33 @@ pub fn read_from_path<P: AsRef<Path>>(p: P, configuration: Configuration) -> Res
...
+/// ```
+// let url = "http://some_url/gtfs.zip";
+// let resp = reqwest::blocking::get(url)?; // or async call
+// let data = std::io::Cursor::new(resp.bytes()?.to_vec());
+// let model = transit_model::gtfs::from_read(data, &url, configuration)?;
+/// ```
+///
+/// The `source_name` is needed to have nicer error messages.
+/// The `Configuration` is used to control various parameters during the import.
+pub fn from_read<R>(reader: R, source_name: &str, configuration: Configuration) -> Result<Model>
```
datanel, line 367:
```suggestion
pub fn from_reader<R>(reader: R, source_name: &str, configuration: Configuration) -> Result<Model>
```

antoine-de: `parse_zid_read` no? (cf https://github.com/CanalTP/transit_model/issues/739#issuecomment-771466749)

antoine-de:
can we merge ?

src/gtfs/mod.rs
datanel: we will rename later

antoine-de: nop you're right we can discuss it in #739 
you want me to roll back to the previous name?

datanel: yes please

antoine-de: done

datanel (dismissed):

woshilapin (commented):
Might need a rebase since I see commits about CO2 that already have been merged. And also, I thought you were planning to address [this comment](https://github.com/CanalTP/transit_model/pull/740/files/9012308299fd929d1bfde54eb530221de5a9d63c..4c602b6ebcc32793f5cf4a74ce2009e6f4bcffc9#r567624882) :smile:

antoine-de:
oups, done

woshilapin (approved):

datanel (approved):

PR #736, opened 2021-01-29
[doc] gtfs2ntfs: precision on line comments

papailio:
cf. ND-1075

PR #741, opened 2021-02-03
handle route comments on line when using read_as_line

Tristramg:
cf. ND-1075 and replaces #736

### DIFF
--- a/documentation/gtfs_to_ntfs_specs.md
+++ b/documentation/gtfs_to_ntfs_specs.md
@@ -188,7 +188,7 @@ _Warning :_ If the GTFS route has no trips, the Navitia Route should NOT be crea
 | routes.txt   | direction_type | Optional   |            |                 | (2)                                                                                                                                                         |
 | routes.txt   | line_id        | Required   |            |                 | corresponding `line.id` (see Line construction above)                                                                                                       |
 | routes.txt   | destination_id | Optional   |            |                 | This field contains a stop_area.id of the most frequent destination of the contained trips (ie. the parent_station of the most frequent last stop of trips) |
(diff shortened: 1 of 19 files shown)


woshilapin (commented):
I'm a bit annoyed that `generate_route_comment()` might return a `Comment` either with an identifier prefixed with `route:` or `line:`. Then only after in `read_routes()`, the code is branching on `read_as_line` and decide to use this comment in a line or in a route. So in terms of code architecture, it's possible to add a comment `route:` in a `line` which would be wrong. I know both actions are conditioned to `read_as_line` so it won't happen at the moment but since the logic is in 2 different places, future maintenance of the code might break this invariant. I'm wondering how the code could be refactored to better reflect this.

I'm not sure if I'm very clear about what I'm trying to explain, so do not hesitate to ping me in private to better explain.

Tristramg:
I think I understand what annoys you, but I’m not sure how to adress it.

First, we can rename it `comment_from_gtfs_route` to avoid some abiguity.
Secondly, would you prefer to duplicate the call?

woshilapin:
> I think I understand what annoys you, but I’m not sure how to adress it.
> 
> First, we can rename it `comment_from_gtfs_route` to avoid some abiguity.
> Secondly, would you prefer to duplicate the call?

I propose [this PR](https://github.com/CanalTP/transit_model/pull/742) that might address the problem. Tell me what you think about it.

woshilapin (approved):
Good for me. Do not hesitate to squash/rebase if needed.

Tristramg:
I just squashed you changes. Thank you!

woshilapin (approved):

PR #742, opened 2021-02-04
[tech] Refactorization of generate comment for GTFS read_as_line 

woshilapin:
Here is an idea to address this [comment](https://github.com/CanalTP/transit_model/pull/741#issuecomment-773359625).

### DIFF
--- a/src/gtfs/read.rs
+++ b/src/gtfs/read.rs
@@ -612,15 +612,28 @@ fn generate_stop_comment(stop: &Stop) -> Option<objects::Comment> {
     })
 }
 
-fn generate_route_comment(route: &Route, read_as_line: bool) -> Option<objects::Comment> {
-    let prefix = if read_as_line { "line:" } else { "route:" };
-    route.desc.as_ref().map(|desc| objects::Comment {
-        id: prefix.to_string() + &route.id,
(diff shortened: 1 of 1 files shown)


src/gtfs/read.rs
```diff
@@ -1183,23 +1172,44 @@ where
     collections.routes = CollectionWithId::new(routes)?;
 
     gtfs_routes_collection.iter().for_each(|(_id, gtfs_route)| {
-        if let Some(comment) = generate_route_comment(&gtfs_route, read_as_line) {
-            if read_as_line {
-                if let Some(mut line) = collections.lines.get_mut(&gtfs_route.id) {
-                    line.comment_links.insert(comment.id.to_string());
-                    collections
-                        .comments
+        fn insert_comment<'c, T: typed_index_collection::Id<T> + objects::CommentLinks>(
```
datanel, line 1175: this function inside the for_each disrupts my reading. Can you put it outside please?

woshilapin: I intentionally put it inside since it is of no use anywhere else. Anyway, it was already the case with the previous `generate_route_comment()` function, so I'll put it back there to replace it.

Done in 2254c72fb0024eff1cf1d95a04896866c94ac900.

datanel (approved):

antoine-de (approved):

Tristramg (approved):

PR #745, opened 2021-02-16
[ci] use 'rust-ci:latest-proj'

woshilapin:
Use now the custom image that already contains `rustfmt`, `clippy`,`cargo-audit` and `libproj`. This should improve the build time and make it more reliable with `libproj` since we're not depending anymore on the custom `.deb`.

- [x] release `kisiodigital/rust-ci` (see https://github.com/CanalTP/ci-images/pull/17)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,40 +8,25 @@ jobs:
   rustfmt:
     name: Formatting check
     runs-on: ubuntu-latest
+    container: kisiodigital/rust-ci:latest
     steps:
     - uses: actions/checkout@master
-    - name: Install Rust stable
-      uses: actions-rs/toolchain@v1
-      with:
-          toolchain: stable
-          profile: minimal
-          components: rustfmt
     - name: Check formatting
       run: make format
 
   clippy:
     name: Analyzing code with Clippy
     runs-on: ubuntu-latest
+    container: kisiodigital/rust-ci:latest-proj
     steps:
-    - name: Install proj
-      run: |
-        wget --quiet --output-document - "https://kisiodigital.jfrog.io/kisiodigital/api/gpg/key/public" | sudo apt-key add -
-        echo "deb [arch=amd64] https://kisiodigital.jfrog.io/kisiodigital/debian-local stretch main" | sudo tee /etc/apt/sources.list.d/kisio-digital.list
-        sudo apt update
-        sudo apt install --yes pkg-config libssl-dev clang libtiff-dev libcurl4-nss-dev proj=${PROJ_VERSION}
     - uses: actions/checkout@master
-    - name: Install Rust stable
-      uses: actions-rs/toolchain@v1
-      with:
-          toolchain: stable
-          profile: minimal
-          components: clippy
     - name: Linting
       run: make lint
 
   audit:
     name: Audits
     runs-on: ubuntu-latest
+    container: kisiodigital/rust-ci:latest-proj
     continue-on-error: true
     steps:
     - uses: actions/checkout@v1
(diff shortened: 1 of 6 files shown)


antoine-de (approved):

pbougue (approved):

.github/workflows/ci.yml
```diff
@@ -52,55 +37,23 @@ jobs:
...
-            rust: beta
+    runs-on: ubuntu-latest
+    container: kisiodigital/rust-ci:latest-proj
     steps:
-    - name: Install proj
-      run: |
-        wget --quiet --output-document - "https://kisiodigital.jfrog.io/kisiodigital/api/gpg/key/public" | sudo apt-key add -
-        echo "deb [arch=amd64] https://kisiodigital.jfrog.io/kisiodigital/debian-local stretch main" | sudo tee /etc/apt/sources.list.d/kisio-digital.list
-        sudo apt update
-        sudo apt install --yes libxml2-utils pkg-config libssl-dev clang libtiff-dev libcurl4-nss-dev proj=${PROJ_VERSION}
```
woshilapin, line 72: Oops, `libxml2-utils` is needed for tests (and `xmllint` verification). See 349c181066715a14f1e4714eb5bd3f1ae5d65e9f.

woshilapin:
- I removed all `EPSG:*` instance of the code to not depend anymore on `proj-data` (see 9530a1a4f408c91f67d44c39bf930f725027e659)
- I updated the Dockerfile to not depend anymore on `.deb` for `proj` (see 311d8328e4248d9c87b5297f48690df90d7c3fda)
- `cargo-tarpaulin` seems to be in need of `pkg-config` and `libssl-dev` (see b7694b6ba3bbd47f5ed5ecf118a8ac04ab88824b)

Dockerfile
```diff
@@ -17,18 +9,27 @@ RUN cargo build --workspace --release \
 
 FROM debian:stretch-slim
```
pbougue, line 10: Not sure about this (I know it's after my suggestion): using stretch (debian 9) here, buster (debian 10) in builder.

Dockerfile
```diff
@@ -17,18 +9,27 @@ RUN cargo build --workspace --release \
...
+## Inspired by Docker image `kisiodigital/rust-ci:latest-proj`
+## See https://github.com/CanalTP/ci-images/blob/master/rust/proj/Dockerfile
+# For building `libproj' and the Rust's crate `proj-sys`, the following Debian packages are needed:
+ARG BUILD_DEPENDENCIES="libcurl4-gnutls-dev libsqlite3-dev libtiff5-dev cmake pkg-config sqlite3 wget"
+# For running `proj`, the following Debian packages are needed:
+# - 'clang' provides 'llvm-config', 'libclang.so' and 'stddef.h' needed for compiling 'proj-sys'
+# - 'libtiff5' provides 'libtiff.so', needed for linking when 'proj-sys' is used
+# - 'libcurl3-nss' provides 'libcurl-nss.so', needed for linking when 'proj-sys' is used
+# - 'proj' provides 'proj.h' and 'libproj.so', needed for compiling 'proj-sys' (installed manually below)
+ARG RUNTIME_DEPENDENCIES="clang libtiff5 libcurl3-nss"
```
pbougue, line 21: Those changed, is it after work on rust-ci that you detected we could use more precise packages?
Also it looks like we got them on both debian 9 and 10, but should we test that a bit?

pbougue (commented):
Can you also update Makefile and readme to tell we're using proj 7.2.1 and not 7.1.0 anymore?

Dockerfile
woshilapin: Oh right, now that we depend on `kisiodigital/rust-ci`, we depend on `buster`, not `stretch` anymore. I would say it'd be safer to use a `buster` here. Do you say any risk @datanel @ArnaudOggy?

Dockerfile
woshilapin: I've done this work a while ago. What kind of failure do we expect that would need additional testing?

And indeed, they're available on `buster`:
- https://packages.debian.org/stable/clang
- https://packages.debian.org/stable/libcurl3-nss
- https://packages.debian.org/stable/libtiff5

pbougue: I think we should just test that all concerned binaries are running fine : are they all running on dev?

woshilapin: I've made some improvement on how the `Dockerfile` build the project (see 367d943380d1fcae67d847efebbb1259f4dcef83). I was able to test that the binaries are running fine inside the container.

Dockerfile
woshilapin: Now using `stretch` everywhere to avoid making more modification that the primary focus of this PR (see 367d943380d1fcae67d847efebbb1259f4dcef83).

woshilapin:
> Can you also update Makefile and readme to tell we're using proj 7.2.1 and not 7.1.0 anymore?

Should be good now, I've fixed it in b27f79f43de5da6a9c8b3f5e3efc55ea66adbd3e.

pbougue (approved):

Dockerfile
```diff
@@ -1,34 +1,56 @@
...
-	&& rm -rf /var/lib/apt/lists/*
+ARG RUNTIME_DEPENDENCIES
+# install 'proj'
+# We copy all the files to the root ('/tmp/proj-build' contains a 'usr/' folder)
+COPY --from=proj-builder /tmp/proj-build /
+RUN apt update \
+    && apt install --yes ${RUNTIME_DEPENDENCIES} \
+    && apt autoremove --yes \
+    && rm -rf /var/lib/apt/lists/*
 COPY --from=builder /usr/src/bin/* /usr/local/bin/
```
pbougue, line 56:
```suggestion
COPY --from=rust-builder /usr/src/bin/* /usr/local/bin/
```

pbougue (changes requested):

Dockerfile
```diff
@@ -1,34 +1,56 @@
...
-RUN cargo build --workspace --release \
-	&& mkdir /usr/src/bin && for file in ls ./target/release/*; do if test -f $file -a -x $file; then cp $file /usr/src/bin; fi; done \
-	&& cd .. && rm -rf app
+# install rustup
+RUN apt install --yes curl
+RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
+ENV PATH "/root/.cargo/bin:$PATH"
+# build the project
+RUN cargo build --workspace --release
+RUN mkdir /usr/src/bin && for file in ls ./target/release/*; do if test -f $file -a -x $file; then cp $file /usr/src/bin; fi; done
```
pbougue, line 44:
```suggestion
RUN mkdir /usr/src/bin && for file in ls ./target/release/*; do if test -f $file -a -x $file; then cp $file /usr/src/bin; fi; done
```

Dockerfile
```diff
@@ -1,34 +1,56 @@
-ARG PROJ_VERSION="7.1.0"
+## Inspired by Docker image `kisiodigital/rust-ci:latest-proj` for `proj` installation
+## See https://github.com/CanalTP/ci-images/blob/master/rust/proj/Dockerfile
```
pbougue, line 2: Those could be moved below, on proj-builder

Dockerfile
```diff
@@ -1,34 +1,56 @@
...
+# install rustup
+RUN apt install --yes curl
+RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
+ENV PATH "/root/.cargo/bin:$PATH"
+# build the project
+RUN cargo build --workspace --release
+RUN mkdir /usr/src/bin && for file in ls ./target/release/*; do if test -f $file -a -x $file; then cp $file /usr/src/bin; fi; done 
 
 FROM debian:stretch-slim
 ARG PROJ_VERSION
```
pbougue, line 47:
```suggestion
```

Dockerfile
woshilapin: Not the `RUNTIME_DEPENDENCIES` which is also used in other stages. And for `PROJ_VERSION`, I think it's pretty nice to have it configurable for `docker build`. I'll move `BUILD_DEPENDENCIES` and transform into a `ENV` (also, `curl` is not needed anymore in `BUILD_DEPENDENCIES`).

Dockerfile
woshilapin: Fixed in 53ce6c1a24f668febcc59c5fc76660b05277acbe. Thanks

Dockerfile
woshilapin: Done in 1c641b1fcc6be28f7446a7a064050b95988cc4b7.

Dockerfile
woshilapin: Done in 15d15f081878b80fb4160c289389516bd2b6ec36.

pbougue (approved):

PR #747, opened 2021-02-22
[feature] Improve line names

ArnaudOggy:
If a line name is empty, it's set with the name of its first "forward" route (in alphabetical order)

Ref. ND-1229

Note: apparently no fixtures to update for its little brother tartare-tools :wink:


src/model.rs
```diff
@@ -1153,6 +1153,31 @@ impl Collections {
         }
     }
 
+    /// If a line name is empty, it's set with the name of its first "forward" route (in alphabetical order)
+    pub fn enhance_line_names(&mut self, lines_to_routes: &impl Relation<From = Line, To = Route>) {
+        let mut line_names: HashMap<Idx<Line>, String> = HashMap::new();
```
antoine-de, line 1158: very minor and quite picky but if I understand this correctly, this could be a `Vec<(Idx<Line>, String)>` no?

src/model.rs
```diff
@@ -2566,6 +2592,179 @@ mod tests {
...
+                OneToMany::new(&collections.lines, &collections.routes, "lines_to_routes").unwrap();
+
+            collections.enhance_line_names(&lines_to_routes);
+            let line1 = collections.lines.get("line_id1").unwrap();
+            assert_eq!("my route id1", line1.name);
+        }
+
+        #[test]
+        fn non_empty_line_with_non_empty_route() {
+            let mut collections = Collections::default();
```
antoine-de, line 2631: If the line were handled by the `ModelBuilder` you can use it here and I think it would be nice to.
This could be written as 
```rust
let model = transit_model_builder::ModelBuilder::default()
        .line("line_id1", |l| {l.name = "my line id".to_owned();})
        .route("route_id1", |r| {
            r.line_id = "line_id1".to_owned();
            r.direction_type = Some("forward".to_owned());
            r.name = "my route id1".to_owned();
        })
        .build();
```

woshilapin (commented):
I'm wondering why `enhance_route_name` was not used for that. A function like the following one could have been extracted
```
fn generate_name(trips: impl Iterator<Item = VehicleJourney>) -> String;
```
and then used for both `enhance_route_name` and `enhance_line_name`. This would make it homogeneous in the way we generate these names.

Note however that what I'm proposing has a substantial performance impact compared to your solution since calculating frequencies is more work and is obviously different in term of requirement.

src/model.rs
```diff
@@ -1153,6 +1153,31 @@ impl Collections {
         }
     }
 
+    /// If a line name is empty, it's set with the name of its first "forward" route (in alphabetical order)
+    pub fn enhance_line_names(&mut self, lines_to_routes: &impl Relation<From = Line, To = Route>) {
+        let mut line_names: HashMap<Idx<Line>, String> = HashMap::new();
+        for (line_idx, _) in self.lines.iter().filter(|(_, l)| l.name.is_empty()) {
```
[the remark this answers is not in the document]
antoine-de: nop, because he need the index in the loop I think

src/model.rs
antoine-de: I can add it to your PR if you want writing rust would be a nice Jenkins/kube break :grin:

src/model.rs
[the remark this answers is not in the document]
woshilapin: Oh right, forget about it.

src/model.rs
ArnaudOggy: Lighter indeed ! Done 06f0279a0bccb921596f676c8c1bbc376f74f0b3

ArnaudOggy:
> I'm wondering why `enhance_route_name` was not used for that. A function like the following one could have been extracted
> 
> ```
> fn generate_name(trips: impl Iterator<Item = VehicleJourney>) -> String;
> ```
> 
> and then used for both `enhance_route_name` and `enhance_line_name`. This would make it homogeneous in the way we generate these names.
> 
> Note however that what I'm proposing has a substantial performance impact compared to your solution since calculating frequencies is more work and is obviously different in term of requirement.

as seen together we will keep it simple for the moment, but indicate this possibility in code comment, and in the associated request ticket

src/model.rs
ArnaudOggy: we keep it in mind for later :smile:

antoine-de (approved):
:+1:

src/enhancers/adjust_lines_names.rs
```diff
@@ -0,0 +1,216 @@
...
+            .map(|route| &route.name)
+            .collect::<BTreeSet<_>>()
+            .into_iter()
+            .next();
+        if let Some(route_name) = route_name {
+            line_names.push((line_idx, route_name.clone()));
+        }
+    }
+    for (line_idx, line_name) in line_names {
+        // note: choice is made to keep 'forward_line_name' and 'backward_line_name' as its are
```
woshilapin, line 38:
```suggestion
        // note: choice is made to keep 'forward_line_name' and 'backward_line_name' as it is
```

woshilapin (approved):

PR #752, opened 2021-03-02
[ci] report 'audit' errors

woshilapin:
Try to use the original Github Actions to produce `cargo audit` comments in the PR. Superseed https://github.com/CanalTP/transit_model/pull/751 (if it works).


woshilapin:
Not working and no time to investigate more (might be due to the bug mentioned [here](https://github.com/actions-rs/clippy-check/issues/2)).

PR #755, opened 2021-03-12
[fix] allow millisecond in schedule

patochectp:
ref: ND-1241

### DIFF
--- a/src/objects.rs
+++ b/src/objects.rs
@@ -786,7 +786,8 @@ impl FromStr for Time {
         };
         let hours: u32 = hours.parse()?;
         let minutes: u32 = minutes.parse()?;
-        let seconds: u32 = seconds.parse()?;
+        let last_second_number = seconds.find('.').unwrap_or_else(|| seconds.len());
+        let seconds: u32 = seconds[..last_second_number].parse()?;
         if minutes > 59 || seconds > 59 {
             return Err(TimeError::WrongValue);
         }
@@ -1926,9 +1927,12 @@ mod tests {
         assert_eq!(Time::new(13, 37, 0), de("13:37:00").unwrap());
         assert_eq!(Time::new(0, 0, 0), de("0:0:0").unwrap());
         assert_eq!(Time::new(25, 42, 42), de("25:42:42").unwrap());
+        assert_eq!(Time::new(13, 37, 0), de("13:37:00.000").unwrap());
+        assert_eq!(Time::new(13, 37, 0), de("13:37:00.999").unwrap()); // currently floor on ms (not strictly required)
 
         assert!(de("").is_err());
         assert!(de("13:37").is_err());
+        assert!(de("13:37:").is_err());
         assert!(de("AA:00:00").is_err());
         assert!(de("00:AA:00").is_err());
         assert!(de("00:00:AA").is_err());


src/objects.rs
```diff
@@ -786,7 +786,10 @@ impl FromStr for Time {
         };
         let hours: u32 = hours.parse()?;
         let minutes: u32 = minutes.parse()?;
-        let seconds: u32 = seconds.parse()?;
+        let seconds: u32 = match seconds.split('.').next() {
+            Some(seconds) => seconds.parse()?,
+            _ => return Err(TimeError::WrongValue),
+        };
```
woshilapin, line 792: This is kind of misleading since the `_` will never be reached as far as I understand (`split()` always return at least one element). An alternative solution would be to parse a sub-slice of `seconds` from start until the character dot `.` or until the end.

```suggestion
        let seconds: u32 = seconds[..seconds.find('.').unwrap_or(seconds.len())].parse()?;
```

or maybe more readable in two lines

```suggestion
        let last_second_number = seconds.find('.').unwrap_or(seconds.len());
        let seconds: u32 = seconds[..last_second_number].parse()?;
```

patochectp:
I know but I didn’t want to use `unwrap()`

src/objects.rs
woshilapin: Note that this is a `.unwrap_or()` (which will not `panic!()` in case of `None`), not an `.unwrap()` (which will `panic!()` in case of `None`).

patochectp: I know but I didn’t want to use unwrap()

woshilapin: Note that this is a `.unwrap_or()` (which will not `panic!()` in case of `None`), not an `.unwrap()` (which will `panic!()` in case of `None`).

woshilapin: Another option if you prefer the `match` would be to do the following.

```suggestion
        let seconds: u32 = match seconds.split('.').next() {
            Some(seconds) => seconds.parse()?,
            _ => unreachable!(".split() always return at least one element"),
        };
```

patochectp: I know but in my use I didn't want any default value

patochectp: done in 2720d2b7feba5aa9639dc27a2dc37e6e46ac0ee7

woshilapin (approved):

src/objects.rs
```diff
@@ -1926,9 +1927,11 @@ mod tests {
         assert_eq!(Time::new(13, 37, 0), de("13:37:00").unwrap());
         assert_eq!(Time::new(0, 0, 0), de("0:0:0").unwrap());
         assert_eq!(Time::new(25, 42, 42), de("25:42:42").unwrap());
+        assert_eq!(Time::new(13, 37, 0), de("13:37:00.000").unwrap());
```
pbougue, line 1930: Could you add a test on `"13:37:00.999"` just to be explicit (maybe tell it's not a requirement, but it's assumed that `currently we floor values`):
```suggestion
        assert_eq!(Time::new(13, 37, 0), de("13:37:00.000").unwrap());
        assert_eq!(Time::new(13, 37, 0), de("13:37:00.999").unwrap());  // currently floor on ms (not strictly required)
```

pbougue (approved):

ArnaudOggy (dismissed):

ArnaudOggy (approved):

antoine-de (approved):

PR #759, opened 2021-03-26
[tech] Update to clippy 1.51

patochectp:


woshilapin (approved):

src/objects.rs
```diff
@@ -1589,7 +1589,7 @@ impl AddPrefix for PriceV1 {
 }
 
 #[derive(Serialize, Deserialize, Debug, PartialEq, Eq, PartialOrd, Ord)]
-pub struct ODFareV1 {
+pub struct OdFareV1 {
```
pbougue, line 1592: I'm not so sure about that change.

src/model.rs
```diff
@@ -1484,16 +1484,16 @@ mod tests {
         // - stop_point_idx (usize -> index of one of the four test stop points)
         // - arrival_time (Time)
         // - departure_time (Time)
-        type VJConfig = (String, usize, Time, Time);
+        type VjConfig = (String, usize, Time, Time);
```
pbougue, line 1487: Same, not sure, as `vj` (or `Vj`) is not a word.

pbougue: And then I read this: https://users.rust-lang.org/t/naming-convention-suggests-name-that-is-read-incorrectly/55002/27
So... Fine by me :shrug: , I'm OK to not argue against Rust convention on that one.

pbougue (approved):

pbougue (commented):
Should you update the toml version on the way?

src/model.rs
woshilapin: Yes, `clippy` actually links to [this](https://rust-lang.github.io/api-guidelines/naming.html#casing-conforms-to-rfc-430-c-case) and they're pretty specific about the acronyms...

PR #761, opened 2021-03-31
[feature] add 'Model::add_calendar' under 'mutable-model' feature

woshilapin:
Introduce a new function to mutate the model.

```rust
Model::add_calendar(
  &mut self,
  calendar: Calendar,
  // Other parameters ?
)
```

### DIFF
--- a/src/model.rs
+++ b/src/model.rs
@@ -1242,6 +1242,7 @@ impl Collections {
 pub struct Model {
     collections: Collections,
 
+    // WARNING: Please check all methods that takes &mut self before adding a new relation (see feature 'mutable-model')
     // original relations
     networks_to_lines: OneToMany<Network, Line>,
     commercial_modes_to_lines: OneToMany<CommercialMode, Line>,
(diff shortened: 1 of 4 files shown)


src/model.rs
```diff
@@ -1446,6 +1446,78 @@ impl Model {
...
     }
 }
+#[cfg(feature = "mutable-model")]
+impl Model {
+    /// Add a Calendar inside the model and link it automatically to the corresponding Vehicle Journeys
+    pub fn add_calendar(
+        &mut self,
+        calendar: Calendar,
+        vehicle_journeys: IdxSet<VehicleJourney>,
+    ) -> Result<()> {
```
woshilapin, line 1456: Maybe this can be split into 2 functions. One that add objects into the `Model` (and might create orphan objects) and another one specifically to modify the relations.

woshilapin:
This PR depends on https://github.com/CanalTP/relational_types/pull/11. However, there is still a bunch of discussions about that other PR and [no consensus](https://github.com/CanalTP/relational_types/pull/11#issuecomment-812170876). Therefore, I'm not going to rely on this new `OneToMany::add_link()` function, but I'm going to recreate entirely the relation every time with `OneToMany::new()`. This might be less performant, but it will be always correct and performance improvement can still be addressed later [if needed](https://www.wikiwand.com/en/Program_optimization#/When_to_optimize).

src/model.rs
woshilapin: The function has been split into 2 parts.

```rust
fn add_calendar(&mut self, calendar: Calendar) -> Result<Idx<Calendar>>;
fn connect_calendar_to_vehicle_journeys(
  &mut self,
  calendar_idx: Idx<Calendar>,
  vehicle_journey_idxs: IntoIterator<Item = Idx<VehicleJourney>>,
) -> Result<()>;

**Review by ArnaudOggy — DISMISSED** [review]

File: src/model.rs [hunk 2]
```diff
@@ -1446,6 +1447,86 @@ impl Model {
...
+            .add_calendar(Calendar {
+                id: "service2".to_string(),
+                ..Default::default()
+            })
+            .unwrap();
+        model
+            .connect_calendar_to_vehicle_journeys(service2_idx, vec![vj2_idx])
+            .unwrap();
+
+        // hat 'service2' is accessible from 'vj2'
```
pbougue, line 1516: typo? (same below)
```suggestion
        // Verify that 'service2' is accessible from 'vj2'
```

pbougue (approved):

src/model.rs
woshilapin: Thanks, fixed in a3b1a9aa50f9fd196ab061e163cfdb4ec04094b8.

pbougue (approved):

PR #763, opened 2021-04-09
[features] object_properties are a BTreeMap instead of BTreeSet

woshilapin:
Related to https://github.com/CanalTP/ntfs-specification/pull/24.


src/objects.rs
```diff
@@ -87,17 +87,29 @@ macro_rules! impl_codes {
...
 }
 
+pub type PropertiesMap = std::collections::BTreeMap<String, String>;
+/// Helper to create a map of properties. Take a list of tuples `(key, value)`.
+#[macro_export]
+macro_rules! properties_map {
+    ($(($k:expr, $v:expr)),*) => {{
+        let mut map = std::collections::BTreeMap::default();
+            $(
+                map.insert($k, $v);
```
pbougue, line 97: So this is blindly updating if there is a conflict (last is kept).
I don't see any problem in this PR about that init.
Just commenting to ask if anyone anticipates behavior change on that point.

pbougue (commented):
This looks OK.

Although we might have silent behavior changes, but I'm not sure how to detect them all: check `.insert()` returns `None` inside `insert_object_property`?
I'm not sure that it's an issue either: did not find doc specifying how the conflict should be resolved.

Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 authors = ["Kisio Digital <team.coretools@kisio.com>", "Guillaume Pinot <texitoi@texitoi.eu>"]
 name = "transit_model"
-version = "0.35.0"
+version = "0.36.0"
```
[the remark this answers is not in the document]
woshilapin: `0.35.0` has been released now so this comment is outdated. I'm resolving it.

woshilapin:
Some details about most of the comments. If I'm summarizing, the main concern is what do we want between these 2 behaviors:
- when a duplicate key is to be inserted, we discard it, **first one should win**
- when a duplicate key is to be inserted, we just insert it, **last one should win**

After some more discussions about that topic, we decided that it does not matter that much which one is chosen, as long as it is consistent. Therefore, since **last one should win** is the default strategy for `BTreeMap`, we'll keep it as it is, this will avoid some surprising syntax like

```rust
properties_map.entry(some_key).or_insert(some_value);
```

instead of the usually more expected

```rust
properties_map.insert(some_key, some_value);
```

src/objects.rs
woshilapin: I believe [this comment](https://github.com/CanalTP/transit_model/pull/763#issuecomment-829297250) answer your comment.

patochectp (dismissed):

pbougue (approved):

patochectp (approved):

PR #768, opened 2021-04-22
[feature] check stop time growing

patochectp:
ref: ND-1355


src/objects.rs
```diff
@@ -691,6 +691,44 @@ impl GetObjectType for VehicleJourney {
     }
 }
 
+#[derive(Debug)]
+pub enum StopTimeError {
+    DuplicateStopSequence { vj_id: String },
+    IncoherentStopTimes { vj_id: String },
+}
```
woshilapin, line 698: Excellent idea to introduce a specific type to characterize the error. However, a few remarks.

# Give more information in the error variants

Since you define different variants for the different types, you can use them to keep more information about the error itself, so it would help whoever is trying to make sense of it.

For example, this could be done.

```suggestion
pub enum StopTimeError {
    DuplicateStopSequence {
    	vj_id: String,
    	duplicated_sequence: u32,
    },
    IncoherentStopTimes {
    	vj_id: String,
    	first_incorrect_sequence: u32,
    	first_incorrect_time: Time,
    },
}
```

# Implements `std::error::Error`

Since you're defining an error type, let's implement the standard trait for errors in Rust. Since you're would implement `std::error::Error` and you're implementing `std::fmt::Display` and `std::fmt::Debug`, it would be an excellent opportunity to bring the crate `thiserror` for this.

```suggestion
#[derive(thiserror::Error)]
pub enum StopTimeError {
    #[error("duplicate stop_sequence '{duplicated_sequence}' for the trip '{vj_id}'")]
    DuplicateStopSequence {
    	vj_id: String,
    	duplicated_sequence: u32,
    },
    #[error("incoherent stop times '{first_incorrect_sequence}' at time '{first_incorrect_time}' for the trip '{vj_id}'")]
    IncoherentStopTimes {
    	vj_id: String,
    	first_incorrect_sequence: u32,
    	first_incorrect_time: Time,
    },
}
```

src/objects.rs
```diff
@@ -691,6 +691,44 @@ impl GetObjectType for VehicleJourney {
...
+
+impl std::fmt::Display for StopTimeError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            StopTimeError::DuplicateStopSequence { vj_id } => {
+                write!(f, "duplicate stop_sequence for the trip '{}'", vj_id)
+            }
+            StopTimeError::IncoherentStopTimes { vj_id } => {
+                write!(f, "incoherent stop times for the trip '{}'", vj_id)
```
woshilapin, line 707: The message doesn't make it clear that it is an ordering problem of the times. Here are a few suggestions:
- "stop times are not ordered correctly for the trip '{}'"
- "stop times are not strictly growing for the trip '{}'"

src/objects.rs
```diff
@@ -691,6 +691,44 @@ impl GetObjectType for VehicleJourney {
...
+            }
+        }
+    }
+}
+
+impl VehicleJourney {
+    pub fn sort_and_check_stop_times(&mut self) -> Result<(), StopTimeError> {
+        self.stop_times.sort_unstable_by_key(|st| st.sequence);
+        for (curr_st, next_st) in self.stop_times.iter().zip(self.stop_times.iter().skip(1)) {
```
woshilapin, line 716: I think you can use `.windows()` function for the same effect.

```suggestion
        for [curr_st, next_st] in self.stop_times.windows(2) {
```

src/objects.rs
```diff
@@ -691,6 +691,44 @@ impl GetObjectType for VehicleJourney {
...
+            if curr_st.sequence >= next_st.sequence {
```
woshilapin, line 717: Shouldn't this be an equality comparison? Since you ordered it above, it should not be possible that `curr_st.sequence > next_st.sequence`.

src/objects.rs
```diff
@@ -691,6 +691,44 @@ impl GetObjectType for VehicleJourney {
...
+                return Err(StopTimeError::DuplicateStopSequence {
+                    vj_id: self.id.clone(),
+                });
+            }
+            if curr_st.departure_time > next_st.arrival_time {
```
woshilapin, line 722: Shouldn't this be a "greater or equal" comparison? Because if 2 subsequent stops are delivered at the exact same time, it seems weird that it would be a valid use case... unless some data producers are not able to provide precision to the second :smile:

patochectp: some data producers are provide precision to the millisecond https://github.com/CanalTP/transit_model/pull/755

woshilapin: OK, I'm resolving this comment and also [this one](https://github.com/CanalTP/transit_model/pull/768#discussion_r618981702).

src/model.rs
```diff
@@ -2758,4 +2757,43 @@ mod tests {
...
+        fn check_time_growing() {
+            let collections = collections();
+            let mut vj = create_vehicle_journey_with(
+                "trip:1",
+                vec!["stop_point:1", "stop_point:2"],
+                &collections,
+            );
+            for (sequence, mut stop_time) in vj.stop_times.iter_mut().enumerate() {
+                stop_time.sequence = sequence as u32;
+                let shift = Time::new(0, sequence as u32, 0);
```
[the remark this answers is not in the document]
woshilapin: Resolving because of https://github.com/CanalTP/transit_model/pull/768#discussion_r619021776.

src/objects.rs
patochectp: I didn’t know it but "refutable pattern in `for` loop binding: `&[]`, `&[_]` and `&[_, _, _, ..]` not covered"

woshilapin: Then you can do it like this in this case.

```suggestion
        for window in self.stop_times.windows(2) {
            let curr_st = &window[0];
            let next_st = &window[1];
```

src/objects.rs
patochectp: These are the messages indicated in specifications

src/objects.rs
patochectp: done in d03225efc5decf59f273b03e368313e49e03f244

src/objects.rs
patochectp: done in d03225efc5decf59f273b03e368313e49e03f244

src/objects.rs
patochectp: done in d03225efc5decf59f273b03e368313e49e03f244

src/objects.rs
```diff
@@ -691,6 +692,45 @@ impl GetObjectType for VehicleJourney {
...
+        for window in self.stop_times.windows(2) {
+            let curr_st = &window[0];
+            let next_st = &window[1];
+            if curr_st.sequence == next_st.sequence {
+                return Err(StopTimeError::DuplicateStopSequence {
+                    duplicated_sequence: curr_st.sequence,
+                    vj_id: self.id.clone(),
+                });
+            }
+            if curr_st.departure_time > next_st.arrival_time {
```
ArnaudOggy, line 722: `if curr_st.arrival_time > curr_st.departure_time || curr_st.departure_time > next_st.arrival_time`
Maybe better like this, because it can be wrong within a single stop point (arrival> departure)

ArnaudOggy, line 722:
```suggestion
            if curr_st.arrival_time > curr_st.departure_time || curr_st.departure_time > next_st.arrival_time {
```
Maybe better like this, because it can be wrong within a single stop point (arrival> departure)

ArnaudOggy (approved):

src/ntfs/read.rs
```diff
@@ -334,7 +334,8 @@ pub fn manage_stop_times(collections: &mut Collections, path: &path::Path) -> Re
     collections.stop_time_ids = stop_time_ids;
     let mut vehicle_journeys = collections.vehicle_journeys.take();
     for vj in &mut vehicle_journeys {
-        vj.stop_times.sort_unstable_by_key(|st| st.sequence);
+        vj.sort_and_check_stop_times()
+            .map_err(|e| format_err!("{}", e))?;
```
woshilapin, line 338: Note that now that your new error type implements `std::error::Error`, you can do this

```suggestion
        vj.sort_and_check_stop_times()?;
```

src/objects.rs
```diff
@@ -691,6 +692,48 @@ impl GetObjectType for VehicleJourney {
...
+    #[error("incoherent stop times '{first_incorrect_sequence}' at time '{first_incorrect_time}' for the trip '{vj_id}'")]
+    IncoherentStopTimes {
+        vj_id: String,
+        first_incorrect_sequence: u32,
+        first_incorrect_time: Time,
+    },
+}
+
+impl VehicleJourney {
+    #[allow(clippy::suspicious_operation_groupings)]
```
woshilapin, line 711: You can put this on line 723 to target more precisely the impacted piece of code.

woshilapin (dismissed):

woshilapin (approved):

ArnaudOggy (approved):

Issue #748, opened 2021-02-25
Allow reading GTFS csv files with a flexible option for columns number


fchabouis:
Hi,

I am Francis Chabouis, working as a developper for transport.data.gouv.fr.
We are using gtfs2netexfr to generate NeTEx files from the GTFS available on our plateform.

For the moment I didn't see an option to read GTFS files with a flexible option, meaning a way to convert GTFS files containing CSV with a non-constant number of columns.

It is unfortunately possible that provided CSV files have missing columns on some rows, and we would like gtfs2netexfr to be able to handle such cases.

We have an [issue](https://github.com/etalab/transport-site/issues/1527) on our project, with a an example of such a GTFS.

Thank you and let me know if you need further explanations on this request.

woshilapin:
We'll take a look. Seems like an easy fix. Let us come back to you.

woshilapin:
Fixed by #770.

PR #770, opened 2021-04-27
[fix] gtfs flexible csv

patochectp:
fix https://github.com/CanalTP/transit_model/issues/748
ref: ND-1373


src/read_utils.rs
```diff
@@ -214,18 +239,29 @@ where
...
-    match reader {
-        None => {
+    match (reader, required_file) {
+        (None, false) => {
             info!("Skipping {}", basename);
             Ok(vec![])
         }
-        Some(reader) => {
+        (None, true) => {
+            bail!("file {:?} not found", path)
```
woshilapin, line 248: This is not entirely skipping all errors since it can still create reading errors. I'm wondering if the name should be more something like `read_object_loose()`, `read_objects_weak()`, or `read_opt_objects()`? Personally, I'd prefer the first option.

src/read_utils.rs
```diff
@@ -189,23 +190,47 @@ where
 }
 
 /// Read a vector of objects from a zip in a file_handler
-pub(crate) fn read_objects<H, O>(file_handler: &mut H, file_name: &str) -> Result<Vec<O>>
```
woshilapin, line 192: Most of `read_objects()` and `read_objects_skip_error()` are identical. Can they be factorized?

woshilapin (approved):
Nice improvement and good work on the fixtures too.

src/read_utils.rs
patochectp: done in 8bbd58ef72aebda37fcb331abc375eb89ebd9df3

woshilapin (approved):

PR #773, opened 2021-04-30
[feature] Add the ability to read ziped ntfs

antoine-de:
change the `transit_model::ntfs::read` method to be able to read a zip archive

like the `gtfs` module, the ntfs module now have:
```rust
let model = transit_model::ntfs::read("path_to_a_ntfs_directory_or_a_zip_archive")?; // <- main method, black magic to detect zip or not
```
```rust
let model = transit_model::ntfs::read_from_path("path_to_a_ntfs_directory")?;
```
```rust
let model = transit_model::ntfs::read_from_zip("path_to_a_zip")?;
```
It also brings a lower level `from_read` method if one day we want more flexible reading options like:
```rust
let url = "http://some_url/ntfs.zip";
let resp = reqwest::blocking::get(url)?; // or async call
let data = std::io::Cursor::new(resp.bytes()?.to_vec());
let model = transit_model::ntfs::from_read(data, &url)?;
```

The global behavior can be check in the test file `tests/read_ntfs.rs`.

Note: from method visibility have been changed from `pub` to `pub(crate)` to not expose the `FileHandler`, so this is a breaking change. However I don't think those methods were used in practice. At least not internally by Kisio.


woshilapin (approved):
Instead of doing `pub(crate)` everywhere and having the risk to expose stuff we do not want... maybe it's time to remove the `pub mod ntfs;` in `src/lib.rs` and only expose the methods that intended to be exposed, what do you think?

antoine-de:
hum I think we want a `pub mod` here, we want to be able to do `transit_model::ntfs::read` no?

patochectp (approved):

PR #775, opened 2021-05-03
[tech] add notification to slack

Moomoo-Seukla:
- add notification for each job to slack in case `failure` and `master` branch
 - add a Webhook secret in GitHub `SLACK_CORE_TOOLS_TEAM_URL`  : [here](https://github.com/CanalTP/transit_model/settings/secrets/actions)
- add a message text `global variable` in workflow

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -5,7 +5,7 @@ on:
 env:
   SLACK_TEXT: '{"attachments":[{
                 "text":":warning: Un job demande votre attention !","color":"#D00000",
-                "fields":[{"title":"GitHub URL","value": "${{ env.GITHUB_SERVER_URL }}/${{ env.GITHUB_REPOSITORY }}/actions/runs/${{ env.GITHUB_RUN_ID }}"}]
+                "fields":[{"title":"GitHub URL","value": "https://github.com/CanalTP/transit_model/actions/workflows/publish.yml"}]
                 }]}'
 
 jobs:
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -2,11 +2,6 @@ name: Publish on crates.io
 on:
   release:
     types: [published]
-env:
-  SLACK_TEXT: '{"attachments":[{
-                "text":":warning: Un job demande votre attention !","color":"#D00000",
-                "fields":[{"title":"GitHub URL","value": "https://github.com/CanalTP/transit_model/actions/workflows/publish.yml"}]
-                }]}'
 
 jobs:
   publish:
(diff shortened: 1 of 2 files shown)


.github/workflows/publish.yml
```diff
@@ -2,6 +2,11 @@ name: Publish on crates.io
 on:
   release:
     types: [published]
+env:
+  SLACK_TEXT: '{"attachments":[{
+                "text":":warning: Un job demande votre attention !","color":"#D00000",
+                "fields":[{"title":"GitHub URL","value": "https://github.com/CanalTP/transit_model/actions/workflows/publish.yml"}]
```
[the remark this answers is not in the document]
Moomoo-Seukla: Done, I use context variables

.github/workflows/publish.yml
```diff
@@ -20,3 +25,7 @@ jobs:
       with:
         command: publish
         args: --all-features
+    - name: slack notification (the job has failed)
+      if: failure() && github.ref == 'ref/head/master'
```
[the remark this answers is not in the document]
Moomoo-Seukla: done

.github/workflows/publish.yml
```diff
@@ -2,6 +2,11 @@ name: Publish on crates.io
 on:
   release:
     types: [published]
+env:
+  SLACK_TEXT: '{"attachments":[{
+                "text":":warning: Un job demande votre attention !","color":"#D00000",
```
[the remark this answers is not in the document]
Moomoo-Seukla: done

.github/workflows/publish.yml
```diff
@@ -20,3 +20,13 @@ jobs:
...
         command: publish
         args: --all-features
+    - name: slack notification (the job has failed)
+      if: failure()
+      run: |
+        curl -s -X POST -H "Content-Type: application/json" -d '${{ env.SLACK_TEXT }}' ${{ secrets.SLACK_CORE_TOOLS_TEAM_URL }}
+      env:
+        SLACK_TEXT: '{"attachments":[{
+        "pretext": "`transit_model Publish` requires your attention!",
+        "text":" :warning: Publish failed!","color":"#D00000",
```
pbougue, line 30: minor, to be consistent:
```suggestion
        "text":" :warning: `Publish` failed!","color":"#D00000",
```

.github/workflows/publish.yml
```diff
@@ -20,3 +20,13 @@ jobs:
...
+        "fields":[{"title":"Action URL","value": "https://github.com:/${{ github.action_path }}/${{ github.repository }}/actions/runs/${{ github.run_id }}"}]}
```
pbougue, line 31: (If you want to test it, maybe just use `echo` instead of `curl` to avoid spamming slack :canned_food:)

It does work, but it doesn't look like a legit url :thinking: 
Maybe more:
```suggestion
        "fields":[{"title":"Action URL","value": "https://github.com${{ github.action_path }}/${{ github.repository }}/actions/runs/${{ github.run_id }}"}]}
```

Also, I don't know if what's suggested [here](https://docs.github.com/en/actions/reference/environment-variables) would work (probably better if possible, as it's straight from official doc):  `$GITHUB_SERVER_URL/$GITHUB_REPOSITORY/actions/runs/$GITHUB_RUN_ID`

Moomoo-Seukla: i change url and test it. Done

.github/workflows/publish.yml
Moomoo-Seukla: This format caused errors in CI workflow ( i don't know why). I prefere those like that if you want :)

pbougue: OK, weird but all is consistent now.

pbougue (approved):

PR #780, opened 2021-05-07
[tech]  simpler gtfs reading api 

### LINKED ISSUE #739: [tech] Easier hello world

antoine-de:
closes #739 and preparation adding a bit of docs.

Hide the `Configuration` needed to read GTFS to have easier hello world and an api closer to the ntfs module.

Also add a `gtfs::read` method that detect if the GTFS is a zip or not (like `ntfs::read`).

now the canonimal way to read a gtfs is:

```rust
let model = transit_model::gtfs::read("path to a zip or directory")?;
```

For better control there is also:

```rust
let model = transit_model::gtfs::read_from_path("path to a directory")?;
```
```rust
let model = transit_model::gtfs::read_from_zip("path to a zip")?;
```
or even a `gtfs::from_read` method that takes a reader (cf doc).

When a configuration is needed, a `Reader` struct needs to be instanciated
```rust
let conf = gtfs::Configuration {
            read_as_line: true,
            ..Default::default()
        };
let model = transit_model::gtfs::Reader::new(conf).from(input_dir)?;
```

Note: I did  not use `parse_from_*` like it was discussed previously because I think this naming is more consistent with the NTFS.


antoine-de:
damn clippy doesn't like `Reader::default().from_path(.)` :cry:

antoine-de:
I feel the `from_` not so bad since they are in a `gtfs::Reader` struct, so I felt they were quite understandable, but clippy (and @datanel ) does not agree with me :smile: 

I feel we have several choices (feel free to add more if I missed some):
### 1 keep it this way 
ignore clippy warning

### 2 use parse in struct as proposed

#### 2a and don't change free functions
this way we are coherent with `ntfs::read` method
```rust
    reader.parse_zip_read()
    reader.parse_zip_path()
    reader.parse_dir_path()
    reader.parse()
// and
transit_model::gtfs::read()
transit_model::gtfs::read_from_zip()
transit_model::gtfs::read_from_path()
transit_model::gtfs::from_read()
```

#### 2b and change free functions
And should we change the `ntfs` free functions for coherence?
```rust
    reader.parse_zip_read()
    reader.parse_zip_path()
    reader.parse_dir_path()
    reader.parse()
// and
transit_model::gtfs::parse()
transit_model::gtfs::parse_zip_path()
transit_model::gtfs::parse_dir_path()
transit_model::gtfs::parse_zip_read()
```

### Conclusion
I'd vote for  1. or 2a. since they are easy to implement and I feel the api is nice, but I'm open to suggestion

src/gtfs/mod.rs
```diff
@@ -355,21 +351,106 @@ pub fn read_from_zip<P: AsRef<Path>>(p: P, configuration: Configuration) -> Resu
 /// This method makes it possible to read from a variety of sources like read a GTFS
 /// from the network.
 ///
-/// ```
-// let url = "http://some_url/gtfs.zip";
-// let resp = reqwest::blocking::get(url)?; // or async call
-// let data = std::io::Cursor::new(resp.bytes()?.to_vec());
-// let model = transit_model::gtfs::from_read(data, &url, configuration)?;
+/// ```ignore
```
woshilapin, line 354: Why do we need to ignore it? Because `reqwest` is not part of the dependency-tree? Why didn't it fail before?

For this specific documentation example, we might be able to find another example that doesn't require to bring a big dependency?

woshilapin (approved):
Nice refactorization. For your question, I'd vote 2a too. One is a builder pattern and the other are free functions so I feel they can be named differently. Note that I see one weird thing: we have `reader.parse_zip_read()` but we have `transit_model::gtfs::from_read()`. Should we have `transit_model::gtfs::from_zip_read()`?

src/gtfs/mod.rs
antoine-de: yes reqwest is not a dependency, it cannot be build.
Before the was only 2 `/` so it was not build at all :sweat_smile: 
Do you have an example in mind? It's true that it's a shame not to build all examples, but I think it can be nice to have a real use case here.

antoine-de:
after some discussion the methods are:
image

and 
image

(to lazy to copy it and I think the doc is easier to read)

The ntfs module have been changed accordingly:
* `ntfs::from_read` -> `ntfs::from_zip_reader` 
* `ntfs::read_from_path` -> `ntfs::from_dir` 
*  `ntfs::read_from_zip` -> ntfs::from_zip`

src/gtfs/mod.rs
```diff
@@ -355,21 +351,106 @@ pub fn read_from_zip<P: AsRef<Path>>(p: P, configuration: Configuration) -> Resu
...
+impl Reader {
+    /// Build a Reader with a custom configuration
+    pub fn new(configuration: Configuration) -> Self {
+        Self { configuration }
+    }
+
+    /// Imports a `Model` from the
+    /// [GTFS](https://gtfs.org/reference/static).
+    /// files in the given directory.
+    /// This method will try to detect if the input is a ziped archive or not.
```
datanel, line 395: ziped or zipped?

datanel (approved):

woshilapin (approved):
Why `parse_from_zip_reader()`? I thought `parse_zip_reader()` was fine and a bit less a mouthful.

antoine-de:
isn't `parse_from_zip_reader` more consistent with the other function (`parse_from_zip`, `parse_from_path`)?

or we can remove all the `from` if you prefer (but I don't think it's very bothersome)

woshilapin:
I'd be OK to remove `from_` everywhere.

antoine-de:
ok, last call I hope :sweat_smile: 

I only changed the Reader functions:
```rust
reader.parse(..)
reader.parse_zip(..)
reader.parse_dir(..)
reader.parse_zip_reader(.., ...)
```

datanel (dismissed):

src/gtfs/mod.rs
```diff
@@ -355,21 +351,106 @@ pub fn read_from_zip<P: AsRef<Path>>(p: P, configuration: Configuration) -> Resu
...
+        Self { configuration }
+    }
+
+    /// Imports a `Model` from the
+    /// [GTFS](https://gtfs.org/reference/static).
+    /// files in the given directory.
+    /// This method will try to detect if the input is a zipped archive or not.
+    /// If the default file type mechanism is not enough, you can use
+    /// [Reader::parse_zip] or [Reader::parse_dir].
+    pub fn parse(self, path: impl AsRef<Path>) -> Result<Model> {
```
pbougue, line 398: If it wasn't already discussed, I'd rather name it `read` (same for other "parse" fn)

src/gtfs/mod.rs
```diff
@@ -355,21 +351,106 @@ pub fn read_from_zip<P: AsRef<Path>>(p: P, configuration: Configuration) -> Resu
...
+    /// ```ignore
+    /// let url = "http://some_url/gtfs.zip";
+    /// let resp = reqwest::blocking::get(url)?; // or async call
+    /// let data = std::io::Cursor::new(resp.bytes()?.to_vec());
+    /// let model = transit_model::gtfs::Reader::default().parse_zip_reader(data, &url)?;
+    /// # Ok::<(), Error>(())
+    /// ```
+    ///
+    /// The `source_name` is needed to have nicer error messages.
+    pub fn parse_zip_reader<R>(self, reader: R, source_name: &str) -> Result<Model>
```
pbougue, line 447: On naming (if not already discussed, too): It looks like a `bulk_reader` (or `file_set_reader`), not especially a zip.

src/gtfs/mod.rs
pbougue: Opened a PR targeting this branch to explore activation of doc test: https://github.com/CanalTP/transit_model/pull/782
Let you take only what's fine out of it.

pbougue (approved):
Fine refactor, did not review all thoroughly but :+1: from me.

src/gtfs/mod.rs
antoine-de: yes, we decided to add `zip` in the name, because the reader should be a `zip` ([see here](https://github.com/CanalTP/transit_model/pull/780/files#diff-249da72fb614dd947b88beb2e4d96fa07412868a711f75506e0b2fd3ed9746ccR451))

src/gtfs/mod.rs
antoine-de: we chose `parse` over `read` because it would have led to `read_zip_reader` and it was a bit strange (and a first it was even `read_zip_read` :stuck_out_tongue_winking_eye: )

src/gtfs/mod.rs
woshilapin: Bringing `reqwest` will considerably increase the build time as it pull an important number of transitive dependencies. This is one of the reason `reqwest` was removed from `transit_model` in the first place. So even if what you propose does work, I'm not sure we want to go that way.

Also, it does bring a dependency to the network inside the tests, I'm not entirely sure if it's risky or not but usually, a test that does not depend on _the outside_ is more reliable.

pbougue: OK, locally I didn't hit the `reqwest` build time (but it looks like CI doesn't have it so easy).
Added to network, and to the files that would move...

Should we add only the transit_model:: for Errors? (and the complete output line for ntfs doc-test)?

pbougue (dismissed):
Fine for me, pick what you want (if anything) from my PR and close it :+1:

datanel (approved):

pbougue (approved):

PR #781, opened 2021-05-10
[feature] use the new Iterator::skip_error

woshilapin:
Some possible use of `skip_error_and_log()` on `Iterator`. Depends on CanalTP/skip_error#9.

### DIFF
--- a/src/read_utils.rs
+++ b/src/read_utils.rs
@@ -20,7 +20,7 @@ use crate::{
 use failure::{bail, format_err, ResultExt};
 use log::{info, Level as LogLevel};
 use serde::Deserialize;
-use skip_error::skip_error_and_log;
+use skip_error::SkipError;
 use std::path;
 use std::path::{Path, PathBuf};
 use std::{collections::BTreeMap, io::Read};
@@ -268,14 +268,11 @@ where
                 .flexible(true)
                 .trim(csv::Trim::All)
                 .from_reader(reader);
-            let mut objects: Vec<O> = vec![];
-            for object in rdr.deserialize() {
-                let object: O = skip_error_and_log!(
-                    object.with_context(|_| format!("Error reading {:?}", path)),
-                    LogLevel::Warn
-                );
-                objects.push(object);
-            }
+            let objects = rdr
+                .deserialize()
+                .map(|object| object.with_context(|_| format!("Error reading {:?}", path)))
+                .skip_error_and_log(LogLevel::Warn)
+                .collect();
             Ok(objects)
         }
     }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 authors = ["Kisio Digital <team.coretools@kisio.com>", "Guillaume Pinot <texitoi@texitoi.eu>"]
 name = "transit_model"
-version = "0.36.0"
+version = "0.36.1"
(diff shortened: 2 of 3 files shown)


src/read_utils.rs
```diff
@@ -268,14 +267,12 @@ where
...
-                    object.with_context(|_| format!("Error reading {:?}", path)),
-                    LogLevel::Warn
-                );
-                objects.push(object);
-            }
+            use skip_error::SkipError;
+            let objects = rdr
+                .deserialize()
+                .map(|object| object.with_context(|_| format!("Error reading {:?}", path)))
+                .skip_error_and_log(LogLevel::Warn)
```
antoine-de, line 274: wouln't warn be a better default value since you always use it with this value ?

antoine-de (approved):

src/read_utils.rs
woshilapin: We do not have a "default value" mechanism for now, so I'm not sure what you mean by that. However, I thought we could also add some convenience methods in `skip_error`'s crate like `skip_error_and_warn()`, `skip_error_and_info()`, etc...

antoine-de: of I though I'd seen a `log::Info` default value in skip_error 🤔

woshilapin:
CanalTP/skip_error#9 has been merged and `skip_error:1.1.0` released. I updated the `Cargo.toml` and removed the `do_not_merge` label.

datanel (approved):

src/read_utils.rs
```diff
@@ -268,14 +267,12 @@ where
...
                 .from_reader(reader);
-            let mut objects: Vec<O> = vec![];
-            for object in rdr.deserialize() {
-                let object: O = skip_error_and_log!(
-                    object.with_context(|_| format!("Error reading {:?}", path)),
-                    LogLevel::Warn
-                );
-                objects.push(object);
-            }
+            use skip_error::SkipError;
```
pbougue, line 270: Minor: Is that more idiomatic than on top? (not consistent with gtfs/read.rs)

pbougue (commented):
You should probably update the cargo.toml to `skip_error = "1.1"`.

ArnaudOggy:
@woshilapin 
+ a little bump in TM `0.36.1` for next release :wink:

src/read_utils.rs
woshilapin: For some of the modifications I've made in this PR, I too wondered and might be OK to keep it as it was.

In this specific case, I'd say we avoid a `mut` variable and the type resolution seems better because we don't need to specify `O` anymore.

woshilapin:
> @woshilapin
> 
>     * a little bump to `0.36.1` in Cargo for next release wink

Indeed. Done in 68e241d39bdaeaa3d0ae7d26c0ad35733a15df4a.

woshilapin:
> You should probably update the cargo.toml to `skip_error = "1.1"`.

Good point, done in ec59328440b13b5c9fa5b690b319bdebe68e63b3.

src/read_utils.rs
woshilapin: However, the `use` statement in the middle might not be that idiomatic! Fixed in 024e98f43b6418ef781640ae4f0c883834c87d5c.

pbougue (approved):

PR #782, opened 2021-05-18
[tech] [doc] test doc examples

pbougue:
After https://github.com/CanalTP/transit_model/pull/780/files#r629252682
Renaming the zip files after their move will be required (and a pain point).
So maybe not worth keeping all this, but tested what was missing to activate all doc-test :shrug:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -75,6 +75,7 @@ required-features = ["proj"]
 
 [dev-dependencies]
 approx = "0.3"
+reqwest = { version = "0.11", features = ["blocking"] }
 rust_decimal_macros = "1"
 testing_logger = "0.1"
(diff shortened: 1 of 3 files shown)


antoine-de:
fine for me, a real doc test is always better, but `reqwest` was removed previously because of the impact on the build time

pbougue:
4 options:

1. let it ignored
  a. as-is
  b. use file read + comment that it could be reqwest
2. We can check https://blog.logrocket.com/the-state-of-rust-http-clients/ for light http clients (`ureq`, `http_req`).
Or add `reqwest` back to check if it's OK for compile times.
  a. add dev-deps
  b. add complete deps (and real url-read)

PR #786, opened 2021-06-10
[feature] deduplicate collection

patochectp:
ref: ND-1375

Tested on TT :heavy_check_mark:

### DIFF
--- a/src/model.rs
+++ b/src/model.rs
@@ -26,9 +26,8 @@ use serde::{Deserialize, Serialize};
 use skip_error::skip_error_and_log;
 use std::{
     cmp::{self, Ordering, Reverse},
-    collections::{hash_map::DefaultHasher, BTreeMap, HashMap, HashSet},
+    collections::{BTreeMap, HashMap, HashSet},
     convert::TryFrom,
-    hash::{Hash, Hasher},
     ops,
 };
 use typed_index_collection::{Collection, CollectionWithId, Id, Idx};
@@ -160,19 +159,14 @@ impl Collections {
             }
         }
 
-        fn dedup_collection<T: Clone + Eq + Hash>(source: &mut Collection<T>) -> Collection<T> {
-            let calculate_hash = |t: &T| -> u64 {
-                let mut s = DefaultHasher::new();
-                t.hash(&mut s);
-                s.finish()
-            };
-            let mut set: BTreeMap<u64, T> = BTreeMap::new();
-            let items = source.take();
-            for item in items {
-                set.insert(calculate_hash(&item), item);
+        fn dedup_collection<T: PartialEq>(source: &mut Collection<T>) {
+            let mut dedup = Vec::default();
+            for object in source.take() {
+                if !dedup.contains(&object) {
+                    dedup.push(object);
+                }
             }
-            let collection: Vec<T> = set.values().cloned().collect();
-            Collection::new(collection)
+            *source = Collection::new(dedup);
         }
 
         self.calendars
@@ -486,18 +480,18 @@ impl Collections {
             .retain(|level| level_id_used.contains(&level.id));
         self.calendars.retain(|c| calendars_used.contains(&c.id));
 
-        self.frequencies = dedup_collection(&mut self.frequencies);
-        self.transfers = dedup_collection(&mut self.transfers);
-        self.admin_stations = dedup_collection(&mut self.admin_stations);
-        self.prices_v1 = dedup_collection(&mut self.prices_v1);
-        self.od_fares_v1 = dedup_collection(&mut self.od_fares_v1);
-        self.fares_v1 = dedup_collection(&mut self.fares_v1);
-        self.ticket_prices = dedup_collection(&mut self.ticket_prices);
-        self.ticket_use_perimeters = dedup_collection(&mut self.ticket_use_perimeters);
-        self.ticket_use_restrictions = dedup_collection(&mut self.ticket_use_restrictions);
-        self.grid_exception_dates = dedup_collection(&mut self.grid_exception_dates);
-        self.grid_periods = dedup_collection(&mut self.grid_periods);
-        self.grid_rel_calendar_line = dedup_collection(&mut self.grid_rel_calendar_line);
+        dedup_collection(&mut self.frequencies);
+        dedup_collection(&mut self.transfers);
+        dedup_collection(&mut self.admin_stations);
+        dedup_collection(&mut self.prices_v1);
+        dedup_collection(&mut self.od_fares_v1);
+        dedup_collection(&mut self.fares_v1);
+        dedup_collection(&mut self.ticket_prices);
+        dedup_collection(&mut self.ticket_use_perimeters);
+        dedup_collection(&mut self.ticket_use_restrictions);
+        dedup_collection(&mut self.grid_exception_dates);
+        dedup_collection(&mut self.grid_periods);
+        dedup_collection(&mut self.grid_rel_calendar_line);
 
         Ok(())
     }
--- a/src/model.rs
+++ b/src/model.rs
@@ -26,8 +26,9 @@ use serde::{Deserialize, Serialize};
 use skip_error::skip_error_and_log;
 use std::{
     cmp::{self, Ordering, Reverse},
-    collections::{BTreeMap, HashMap, HashSet},
+    collections::{hash_map::DefaultHasher, BTreeMap, HashMap, HashSet},
     convert::TryFrom,
+    hash::{Hash, Hasher},
     ops,
 };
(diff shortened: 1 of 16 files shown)


src/model.rs
```diff
@@ -159,6 +160,21 @@ impl Collections {
             }
         }
 
+        fn dedup_collection<T: Clone + Eq + Hash>(source: &mut Collection<T>) -> Collection<T> {
+            let calculate_hash = |t: &T| -> u64 {
```
ArnaudOggy, line 164: `vec.dedup()` to avoid clone , :thinking: 
https://doc.rust-lang.org/std/vec/struct.Vec .html#method.dedup

src/model.rs
```diff
@@ -159,6 +160,21 @@ impl Collections {
             }
         }
 
+        fn dedup_collection<T: Clone + Eq + Hash>(source: &mut Collection<T>) -> Collection<T> {
```
ArnaudOggy, line 163: Maybe with `vec.dedup`, `vec.dedup_by` to avoid cloning

patochectp: but we must also manage Idx<T>

src/model.rs
```diff
@@ -159,6 +160,21 @@ impl Collections {
...
+                let mut s = DefaultHasher::new();
+                t.hash(&mut s);
+                s.finish()
+            };
+            let mut set: BTreeMap<u64, T> = BTreeMap::new();
+            let items = source.take();
+            for item in items {
+                set.insert(calculate_hash(&item), item);
+            }
+            let collection: Vec<T> = set.values().cloned().collect();
```
pbougue, line 174: To avoid clones you can probably use `items.retain()`

src/model.rs
woshilapin: I'd propose this implementation instead for the entire function.

```rust
use std::iter::FromIterator;
let set: HashSet<T> = std::collections::HashSet::from_iter(source.take());
Collection::new(set.into_iter().collect())
```
It should avoid reallocation of all the `T`. One of the downside is that we allocate a `HashSet` but I think we can't really avoid that (unless implementing ourselves the one-to-one comparison to identify the duplicates, but I don't think it worth it). A second downside is the allocation of the `Vec` (with the `.collect()`) but this second one could be avoided if we implement `FromIterator` on `Collection` (needs a modification of `typed_index_collection`).

src/objects.rs
```diff
@@ -1528,6 +1528,20 @@ impl AddPrefix for Transfer {
...
+    }
+}
+
+impl PartialEq for Transfer {
+    fn eq(&self, other: &Transfer) -> bool {
+        self.from_stop_id == other.from_stop_id && self.to_stop_id == other.to_stop_id
+    }
+}
+
+impl Eq for Transfer {}
```
woshilapin, line 1543: Since we already have the `derivative` dependency in our dependencies, instead of manually implementing these, you can probably use `derivative` instead (see [documentation](https://mcarton.github.io/rust-derivative/latest/Hash.html)).

src/test_utils.rs
```diff
@@ -34,16 +34,18 @@ pub fn get_file_content<P: AsRef<Path>>(path: P) -> Vec<String> {
     }
     vec
 }
-pub fn get_lines_content<P: AsRef<Path>>(path: P) -> BTreeSet<String> {
+pub fn get_lines_content<P: AsRef<Path>>(path: P) -> Vec<String> {
```
woshilapin, line 37: Inserting in a `BTreeSet` might be more efficient than pushing to a `Vec` and sorting only at the very end. On top of that, this change of code will now keep the duplicates: is it expected?

src/test_utils.rs
```diff
@@ -34,16 +34,18 @@ pub fn get_file_content<P: AsRef<Path>>(path: P) -> Vec<String> {
...
     let path = path.as_ref();
     let file = File::open(path).unwrap_or_else(|_| panic!("file {:?} not found", path));
     let reader = BufReader::new(file);
-    let mut set = BTreeSet::new();
-    for result_line in reader.lines() {
-        let line =
-            result_line.unwrap_or_else(|_| panic!("Cannot parse as a line in file {:?}", path));
-        set.insert(line);
-    }
+    let mut set: Vec<String> = reader
```
woshilapin, line 41: Can you please rename the variable if you really are keeping a `Vec` since it's not a `set` anymore?

woshilapin:
More generally, do you think it would make sense to move this function `dedup_collection` as a method of the type `Collection`, inside `typed_index_collection`? I'd say it might be useful to expose that functionality directly in the crate, what do you think?

patochectp:
> More generally, do you think it would make sense to move this function `dedup_collection` as a method of the type `Collection`, inside `typed_index_collection`? I'd say it might be useful to expose that functionality directly in the crate, what do you think?

I asked myself the question

src/test_utils.rs
patochectp: Yes, we will keep the duplicates for tests comparison.

src/model.rs
pbougue: Maybe not HashSet for stability.

woshilapin: > Maybe not HashSet for stability.

Sure... but implementing `PartialOrd` on the objects `T` might not always make sense.

We might also take a look at [`Vec::dedup()`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.dedup). The downside seems to be that it would work only for sorted `Vec`... and sorting relies on `PartialOrd` :facepalm:

src/model.rs
```diff
@@ -159,6 +160,24 @@ impl Collections {
...
+            source.retain(|object| {
+                let object_hash = calculate_hash(&object);
+                if objects_hashes.contains(&object_hash) {
+                    false
+                } else {
+                    objects_hashes.push(object_hash);
+                    true
+                }
+            });
+        }
```
woshilapin, line 179: Something was bothering me about this function, but I couldn't find out why... I think I finally found it. Why using `Hash` where we could simply use `PartialEq`?

```rust
fn dedup_collection<T: PartialEq>(source: &mut Collection<T>) {
    let mut dedup = Vec::default();
    for object in source.take() {
        if !dedup.contains(&object) {
            dedup.push(object);
        }
    }
    *source = Collection::new(dedup);
}
```

That also means that we're going to implement `PartialEq` on `Transfer` only on `from_stop_id` and `to_stop_id` but I'd say it is as good as doing it for the `Hash` (I don't see why we would do it for `Hash` but not `PartialEq`).

ArnaudOggy: :+1: 
It should be better as you suggest yes. I also felt a weird thing in this use of hash. I will test this.

ArnaudOggy: Done :heavy_check_mark:

src/test_utils.rs
woshilapin: I'm not sure that I understand the need for the changes in this file? I removed the changes on my machine and am still able to run the entire test suite without any problem.

tests/read_ntfs.rs
```diff
@@ -452,3 +452,19 @@ fn ntfs_with_duplicated_ids() {
         }
     });
 }
+
+#[test]
+// Test the sanitation of duplicated objects in collections without id
+fn ntfs_with_duplicated_objects_without_id() {
```
woshilapin, line 458: For this test, I was thinking you could reuse `tests/fixtures/ntfs` or `tests/fixtures/ntfs_minimal` and then do something like this at the end.

```rust
let model = transit_model::ntfs::read("tests/fixtures/ntfs_complete_with_duplicated_ids").unwrap();
test_in_tmp_dir(|output_dir| {
	transit_model::ntfs::write(&ntm, output_dir, get_test_datetime()).unwrap();
	compare_output_dir_with_expected(
    	&output_dir,
    	None,
    	"tests/fixtures/ntfs",
);
```

woshilapin (approved):

src/test_utils.rs
ArnaudOggy: The problem is not visible in TransitModel tests, but is in a TartareTools test (see ND-1375).
This modification is very relevant.

woshilapin: Yes, as discussed together, it is not directly related to the new feature being exposed in this PR, but it is still a bug in our test toolkit that was hiding some errors. So good for me!

tests/read_ntfs.rs
woshilapin: Maybe not as relevant as I thought, so let's keep it like you did.

src/model.rs
woshilapin: OK, so it seems it was a bad idea afterall :facepalm: (see https://github.com/CanalTP/transit_model/pull/808).

PR #788, opened 2021-06-25
[feature] output in ZIP format if extension is '.zip'

woshilapin:
First commit is only about setting up some tests.
Second commit is about the new feature.

The new feature is to detect if `--output` contains `.zip` in the name. If no, the previous behavior is done: writing files to a folder. If yes, then we write the output files to a temporary folder, then ZIP all these files into the `--output` path.

:mag: Help for reviewing:
1. 8ec61ef29f181f47d17356ee3be7d1dd84de54ab: add tests on the binary for `gtfs2ntfs`
2. 75407635be7d51483de7afb33e12887704f1458b: create a ZIP file in `gtfs2ntfs` if output path has extension `.zip`
3. a6ef370690341e582ab9a11450bca48e0519bed0: (1) add `write()` and `write_to_zip()` functions on the API of `transit_model::netex_france` and (2) create a ZIP file in `gtfs2netexfr` and `ntfs2netexfr` if output path has extension `.zip`
4. e8743b18d69acf443fb8a87cef492c63cc866a2b: create a ZIP file in `ntfs2ntfs` if output path has extension `.zip`
5. f79827b85185c4294d13ca2e77e70ebe9ee55050: create a ZIP file in `ntfs2gtfs` if output path has extension `.zip` and add a `write_to_zip()` function to `transit_model::gtfs`

:building_construction: All the steps that have to be implemented
- [X] `gtfs2ntfs` supports ZIP export (75407635be7d51483de7afb33e12887704f1458b)
- [X] `gtfs2netexfr` supports ZIP export (a6ef370690341e582ab9a11450bca48e0519bed0)
- [X] `ntfs2netexfr` supports ZIP export (a6ef370690341e582ab9a11450bca48e0519bed0)
- [x] `ntfs2ntfs` supports ZIP export (e8743b18d69acf443fb8a87cef492c63cc866a2b)
- [x] `ntfs2gtfs` supports ZIP export (f79827b85185c4294d13ca2e77e70ebe9ee55050)
- [X] Bump versions of `Cargo.toml` everywhere (1e5762005b2bb9895622cad4757de6b37e9a3f3d)

:question: Unresolved questions
- [X] Tests added in `gtfs2netexfr` and `ntfs2netexfr` are covering more than what is in `tests/write_netex_france.rs`. Should I remove them from `tests/write_netex_france.rs`? -> On the binary tests, only test that output files are produced (not the entire output fixture) and extend to all the other similar tests (f2af9699554c60c23213d9148b279d3034a50175, 0ee7806a816618cc5cf3fa4a29501e2c370f0dfd, 0ee7806a816618cc5cf3fa4a29501e2c370f0dfd, b7e7f58a1a5271a1e4054a17c82d8a36cd551eba)
- [X] Should we remove `transit_model::netex_france::Exporter` from the public API? -> Make `Exporter` private and possibly document the migration to the new `write()` function (1e5762005b2bb9895622cad4757de6b37e9a3f3d)
- [X] Should we homogenize the `ntfs::write(model: &Model, ...)` and `gtfs::write(model: Model, ...)` (some with reference and some other with ownership)? -> To be done in another PR
- [X] Add a test for `--output` with an extension that is not `.zip` (585b37e922735eed2d298d72a71ede17ccc69ca3)


gtfs2ntfs/src/main.rs
```diff
@@ -144,7 +144,30 @@ fn run(opt: Opt) -> Result<()> {
         None,
     )?;
 
-    transit_model::ntfs::write(&model, opt.output, opt.current_datetime)?;
+    match opt.output.extension() {
+        Some(ext) if ext == "zip" => {
+            let output_dir = tempfile::TempDir::new()?;
+            transit_model::ntfs::write(&model, output_dir.path(), opt.current_datetime)?;
+            let zip_file = std::fs::File::create(opt.output)?;
```
patochectp, line 151: Why you don't use `pub fn zip_to<P, R>(source_path: P, zip_file: R) -> crate::Result<()>` in utils ?

woshilapin: Mostly because I didn't know its existence... :smile:

woshilapin: But now that I know about it... it's a private method... and I'm wondering if this the job of a library like `transit_model` to expose such a function.

woshilapin: Finally, I'm using `transit_model::ntfs::write_to_zip()` that I didn't know about but @patochectp mentioned to me in private.

patochectp (approved):

antoine-de (approved):
🎉 
it's a pity that the zip file is not done using directly a zip archive writer but it's fine for me to do this in a separate PR 😁

woshilapin:
> tada
> it's a pity that the zip file is not done using directly a zip archive writer but it's fine for me to do this in a separate PR grin

Compared to writing to a temporary folder, then into an archive, it sure would be better. But not sure that it'd be as simple to do because it would force us to call `ZipWriter::start_file()` in code that doesn't know (and doesn't have to know) that it is a ZIP or not. We might need an abstraction here... in a separate PR for sure :stuck_out_tongue:

datanel (approved):
approve but -1 for the "pouet" in the fixtures.

src/gtfs/mod.rs
```diff
@@ -562,3 +562,16 @@ pub fn write<P: AsRef<Path>>(model: Model, path: P) -> Result<()> {
 
     Ok(())
 }
+
+/// Exports a `Model` to [GTFS](https://gtfs.org/reference/static) files
+/// in the given ZIP archive.
+/// see [NTFS to GTFS conversion](https://github.com/CanalTP/transit_model/blob/master/src/documentation/ntfs2gtfs.md)
+pub fn write_to_zip<P: AsRef<std::path::Path>>(model: Model, path: P) -> Result<()> {
+    let path = path.as_ref();
+    info!("Writing NTFS to ZIP File {:?}", path);
```
antoine-de, line 571: if it's ntfs to zip, shoudln't this be in src/ntfs/mod.rs?

woshilapin: Oops, some copy/paste that I didn't change... Done in def0c255.

patochectp (approved):

patochectp (dismissed):

patochectp (approved):

PR #789, opened 2021-06-28
[fix] 'install_proj' in the Makefile now works as expected

woshilapin:
Putting a comment at the end of the line makes the space after the version, part of the variable... which obviously breaks the URL afterwards (trying to download to `https://github.com/OSGeo/proj.4/releases/download/7.2.1 /proj-7.2.1 .tar.gz`). `make` is awesome!


pbougue (approved):
:facepalm: :grinning: sorry about that!

ArnaudOggy (approved):

woshilapin:
> facepalm grinning sorry about that!

How could you even imagine. Note that the space BEFORE the version is not taken into account :exploding_head:

PR #790, opened 2021-06-29
[ci] 'cargo-tarpaulin:v0.18.0' does not accept '--all-targets' anymore

woshilapin:
There was a change of version of `cargo-tarpaulin` on the 28th of June. The new version doesn't accept `--all-targets` anymore so our CI fails. This PR should fix it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -100,7 +100,7 @@ jobs:
       uses: actions-rs/cargo@v1
       with:
         command: install
-        args: cargo-tarpaulin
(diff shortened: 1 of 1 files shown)


ArnaudOggy (approved):

patochectp (approved):

woshilapin:
Removing `--all-targets` for version `0.18.0` seems to bring some other problems. Instead, I propose we force version `0.16.0` which was working just fine.

ArnaudOggy (approved):

PR #795, opened 2021-07-07
[tech] Skip and warn duplicated objects on NTFS reading

ArnaudOggy:
Ref. ND-1541


ArnaudOggy:
@woshilapin 
Most of the code is done. It just remains to check that I haven't forgotten anything, 
and to pollute the ntfs (and add the test lines accordingly).
You can watch if you want.
note: done only on ntfs reading (there was no request for the gtfs or other, and no feedback either)

tests/fixtures/ntfs_complete_with_duplicated_ids/lines.txt
```diff
@@ -0,0 +1,4 @@
+line_id,line_code,line_name,forward_line_name,backward_line_name,line_color,line_text_color,line_sort_order,network_id,commercial_mode_id,geometry_id,line_opening_time,line_closing_time
+ME:line1,Line 1,My Line 1,,,,,,ME:ntw1,Metro,ME:4bf028:geo3,10:00:00,21:47:00
+ME:line2,Line 2,My Line 2,,,,,,ME:ntw2,Metro,,14:40:00,25:57:00
+ME:line2,Line 2,My Line 2,,,,,,ME:ntw2,Metro,,14:40:00,25:57:00
```
woshilapin, line 4: I'd rather put different information (except for the identifier) in the other field, to be sure we always keep the same in the tests. Maybe you can name one "Skipped duplicated line" which help also to understand.

tests/read_ntfs.rs
```diff
@@ -391,3 +392,19 @@ fn sanitize_grid_with_line_external_code() {
...
+        transit_model::ntfs::read("tests/fixtures/ntfs_complete_with_duplicated_ids").unwrap();
+    assert_eq!(2, model.lines.len());
+    testing_logger::validate(|captured_logs| {
+        let warn_logs = captured_logs
+            .iter()
+            .filter(|log| log.level == LogLevel::Warn)
+            .collect::<Vec<_>>();
+        assert_eq!(1, warn_logs.len());
+        assert_eq!("identifier ME:line2 already exists", warn_logs[0].body);
+    });
```
woshilapin, line 409: Since you copied an entire fixture, you might just test for the entire output result too. Because running the entire process only for one collision, you might have been able to mae the test simpler and do a unit test on `make_collection_with_id` and `make_opt_collection_with_id`.

woshilapin (commented):
There isn't so many duplicate identifiers in the fixtures. For example, on `calendar.txt` you could put some. Or `pathways.txt`. If I remember correctly, the feature is requested for lines but since your code does more, you might as well test for all the one that are now supported.

tests/read_ntfs.rs
ArnaudOggy: Oh no I will pollute all the collections and test everything. I just had time to do this for the lines quickly yesterday.

tests/fixtures/ntfs_complete_with_duplicated_ids/lines.txt
ArnaudOggy: It’s planned. I didn’t have enough time last night :+1:

patochectp (approved):

woshilapin (approved):

PR #800, opened 2021-07-23
[fix] RUST_LOG error message only when invalid

woshilapin:
Brings `tracing` to all the binaries without printing an error message when `RUST_LOG` is missing (but will still produce an error message when `RUST_LOG` is invalid).

### DIFF
--- a/gtfs2netexfr/Cargo.toml
+++ b/gtfs2netexfr/Cargo.toml
@@ -15,13 +15,10 @@ keywords = ["gtfs", "netex", "transit"]
 chrono = "0.4"
 failure = "0.1"
 log = "0.4"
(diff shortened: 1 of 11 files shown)


woshilapin:
Mmm, interesting, I just read this in an article.

```rust
if std::env::var("RUST_LOG").is_err() {
    std::env::set_var("RUST_LOG", "info")
}
```

It might be more easy to read and understand than what I'm doing with trying to understand the error... What do you think?

ArnaudOggy (approved):

woshilapin:
A different version has been implemented which is more simple.

ArnaudOggy (approved):

PR #803, opened 2021-07-27
[fix] Try to fix Dockerhub login/push

ArnaudOggy:


ArnaudOggy:
@woshilapin 
It crashes. Apparently it's on `/v2` now. Secrets are goods.
I also remove the warning (like before, with the use of `password-stdin`, as recommended)

PR #808, opened 2021-08-17
[fix] Revert dedup simplification

datanel:
ref: https://jira.kisio.org/browse/ND-1572

The commit  https://github.com/CanalTP/transit_model/pull/786/commits/c93a60d9e0c6614d259a0fa0ea4474c6c778eb28 from https://github.com/CanalTP/transit_model/pull/786 deteriorates the performance. I reverted a part of the commit

ntfs2ntfs on a 132MB NTFS

Before > 23min, 2,1 GB Ram
After ~2 min, 3 GB Ram

### DIFF
--- a/src/model.rs
+++ b/src/model.rs
@@ -26,8 +26,9 @@ use serde::{Deserialize, Serialize};
 use skip_error::skip_error_and_log;
 use std::{
     cmp::{self, Ordering, Reverse},
-    collections::{BTreeMap, HashMap, HashSet},
+    collections::{hash_map::DefaultHasher, BTreeMap, HashMap, HashSet},
     convert::TryFrom,
+    hash::{Hash, Hasher},
     ops,
 };
 use typed_index_collection::{Collection, CollectionWithId, Id, Idx};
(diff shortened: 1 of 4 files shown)


src/model.rs
```diff
@@ -159,14 +160,19 @@ impl Collections {
...
+                s.finish()
+            };
+            let mut set: BTreeMap<u64, T> = BTreeMap::new();
+            let items = source.take();
+            for item in items {
+                set.insert(calculate_hash(&item), item);
             }
-            *source = Collection::new(dedup);
+            let collection: Vec<T> = set.values().cloned().collect();
+            Collection::new(collection)
```
datanel, line 175: or we can use the crate https://docs.rs/itertools/0.10.1/itertools/index.html which uses a `HashMap(I::Item, ()>` where `I::Item` are the differents objects

```rust
use itertools::Itertools;
Collection::new(source.into_iter().unique().collect());
```

patochectp (approved):

prhod (approved):

PR #811, opened 2021-09-06
[feature] support 'ntfs:v0.12.1'

woshilapin:
`ntfs:v0.12.1` (compared to `v0.12.0`) introduce the possible value `3` in the fields `drop_off_type` and `pickup_type` of `stop_times.txt`. There is nothing much to do since these fields are implemented as `u8` (so the value `3` is already supported).


patochectp:
You have to bump the version

woshilapin:
> You have to bump the version

Indeed, done.

patochectp (dismissed):

patochectp (approved):

PR #815, opened 2021-09-13
[feature] remove dummy stop times

patochectp:
ref: ND-1603


src/model.rs
```diff
@@ -119,6 +119,27 @@ pub struct Collections {
 }
 
 impl Collections {
+    /// Remove associated schedules with points of route
+    pub fn remove_points_of_route(&mut self) {
```
woshilapin, line 123: How about `remove_route_points()`?

src/model.rs
```diff
@@ -119,6 +119,27 @@ pub struct Collections {
...
+        let vj_idxs: Vec<Idx<VehicleJourney>> =
+            self.vehicle_journeys.iter().map(|(idx, _)| idx).collect();
+        for vj_idx in vj_idxs {
+            let mut vj = self.vehicle_journeys.index_mut(vj_idx);
+
+            let is_point_of_route = |stop_time: &mut StopTime| -> bool {
+                stop_time.pickup_type == 3 || stop_time.drop_off_type == 3
+            };
```
woshilapin, line 131: 1. This function can be declared outside of the `for` loop (and might be named `is_route_point` if you change the other one too).
2. I don't think this closure needs a mutable reference, since it's only about reading information. A simple reference is probably enough.

src/model.rs
```diff
@@ -119,6 +119,27 @@ pub struct Collections {
...
+            let mut stop_time_idx = 0;
+            while stop_time_idx < vj.stop_times.len() {
+                if is_point_of_route(&mut vj.stop_times[stop_time_idx]) {
+                    vj.stop_times.remove(stop_time_idx);
+                } else {
+                    stop_time_idx += 1
+                }
+            }
```
woshilapin, line 139: You might be able to use something like `.retain()` here instead of the entire `while` loop.

```rust
vj.stop_times.retain(|stop_time| !is_point_of_route(stop_time));
```

src/model.rs
```diff
@@ -2840,5 +2861,37 @@ mod tests {
...
+                    });
+                })
+                .build();
+
+            let mut collections = model.into_collections();
+            collections.remove_points_of_route();
+            let vj = collections
+                .vehicle_journeys
+                .get("vj1")
+                .expect("Failed to find vehicle journey vj1");
```
woshilapin, line 2893: minor: we actually don't expect to "Failed to find vehicle journey vj1", we actually expect to **find** it (but all the other tests are doing the same weirdness, so I'm fine if this is not fixed here).

tests/fixtures/ntfs2ntfs/stops/stop_times.txt
```diff
@@ -19,6 +19,6 @@ DEFR,RERAB1,5,09:24:00,09:25:00,0,0,0,1,1,,,,2
 CDGR,RERAB1,8,09:39:00,09:40:00,0,0,0,0,0,,,,0
 GDLR,RERAB1,13,09:44:00,09:45:00,0,0,0,0,0,,,,0
 NATR,RERAB1,21,09:49:00,09:50:00,0,0,0,0,0,,,,0
-MTPZ,RERAB1,50,19:24:00,19:25:00,0,0,0,0,1,,,,2
+MTPZ,RERAB1,50,19:24:00,19:25:00,0,0,0,0,0,,,,0
```
woshilapin, line 22: I'm not sure that I understand why this is changing?

patochectp: This change is due to another anomaly. Discussed with @prhod, this will be corrected later

patochectp:
done in https://github.com/CanalTP/transit_model/pull/816

PR #816, opened 2021-09-14
[feature] Don't export dummy stop time

patochectp:
Ref. ND-1603


src/gtfs/mod.rs
```diff
@@ -504,34 +504,10 @@ struct Route {
...
-    });
-
-    collections
-}
-
 /// Exports a `Model` to [GTFS](https://gtfs.org/reference/static) files
 /// in the given directory.
 /// see [NTFS to GTFS conversion](https://github.com/CanalTP/transit_model/blob/master/src/documentation/ntfs2gtfs.md)
 pub fn write<P: AsRef<Path>>(model: Model, path: P) -> Result<()> {
-    let collections = remove_stop_zones(model);
```
woshilapin, line 533: I think I understand why you did move this line directly to the connector, but it seems it would be better to keep it here, so anything that is writing GTFS will be impacted. With your solution, we would need to think about adding a call to `remove_stop_zones` on every new binary producing some GTFS.

src/model.rs
```diff
@@ -119,6 +119,34 @@ pub struct Collections {
 }
 
 impl Collections {
+    /// Remove associated schedules with route points
+    pub fn remove_route_points(&mut self) {
+        let vj_idxs: Vec<Idx<VehicleJourney>> =
+            self.vehicle_journeys.iter().map(|(idx, _)| idx).collect();
+        let is_point_of_route = |stop_time: &StopTime| -> bool {
```
woshilapin, line 126: If the function above is named `remove_route_points`, this closure can probably be named `is_route_point` :wink:

src/gtfs/mod.rs
woshilapin: OK, so we discussed it together and the conclusion is: yes, it is true that if we ever have another binary that is writing GTFS, we need to not forget to add a call `remove_stop_zones()`. However:
- it is still very unlikely we have to create a new binary that write GTFS (we usually convert to or from NTFS)
- even in the case of such a binary, it is very unlikely that we will end up with `pickup_type` and/or `drop_off_type` with a value of 3
- this PR probably improved performance by an order of magnitude by avoiding transforming a `Model` back to a `Collections`, then back again to a `Model` (which is the main reason why the call to `remove_stop_zones()` has been moved in the first place).

Therefore, we're ok to lose a bit of a "_maybe more logical_" place to call `remove_stop_zones()` in the profit of performance.

tests/fixtures/netex_france/output/reseau_TheGreatNetwork_1e97c33560621530a594ced114597ea1/offre_f31e1eef20f64733a18c538073e78396.xml
```diff
@@ -106,14 +106,14 @@
 							<ForAlighting>true</ForAlighting>
 							<ForBoarding>true</ForBoarding>
 						</StopPointInJourneyPattern>
-						<StopPointInJourneyPattern id="FR:StopPointInJourneyPattern:M1F1_2:" order="3" version="any">
-							<ScheduledStopPointRef ref="FR:ScheduledStopPoint:M1F1_2:">
+						<StopPointInJourneyPattern id="FR:StopPointInJourneyPattern:M1F1_3:" order="4" version="any">
```
woshilapin, line 109: Interesting (and kind of funny too), if we had chosen to make the filtering of route points at the reading of NTFS (instead of at the writing of NeTEx France), then `order` would not have been modified... This almost means that we're revealing an implementation detail here... But I believe it's fine!

woshilapin (approved):

src/model.rs
patochectp: done in f63f8fd7b66e8aea900bc20513d580610d6e75f8

woshilapin (approved):

ArnaudOggy:
We risk having some data inconsistencies, such as vj-headsigns, routes names or others. 
These data exist in the input ntfs, and therefore will not be recalculated.
But it can be acceptable I think.

ArnaudOggy (approved):

ArnaudOggy:
> @patochectp , @woshilapin
> We risk having some data inconsistencies, such as vj-headsigns, routes names or others.
> These data exist in the input ntfs, and therefore will not be recalculated.
> But it can be acceptable i think.

See with the boss. It's ok.

ArnaudOggy (approved):

PR #817, opened 2021-09-21
[fix] do not enhance pickup and drop-off for route-point

woshilapin:
Ref. ND-1644


patochectp (dismissed):

src/enhancers/enhance_pickup_dropoff.rs
```diff
@@ -0,0 +1,794 @@
...
+    #[test]
+    fn stay_in_different_stop_but_with_route_point() {
+        let mut collections = Collections::default();
+        let stop_config = (
+            "block_id_1".to_string(),
+            1,
+            Time::new(10, 0, 0),
+            Time::new(11, 0, 0),
+        );
+        let next_vj_config_config = (
```
ArnaudOggy, line 441: you should rename your variables (vj1_conf, and vj2_conf, or something like that)

woshilapin: Indeed. It was not part of my refactoring to give such dumb names, but it seems it was me who did that in the past :exploding_head: Fixed in 406f72b.

ArnaudOggy (approved):

PR #825, opened 2021-11-04
[tech] Fix StopLocation reading

ArnaudOggy:
Surely an old mistake. 
A StopLocation (like an entry or an exit) "visible" in an ntfs (csv) goes "invisible" at loading / ntfs reading.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 authors = ["Kisio Digital <team.coretools@kisio.com>", "Guillaume Pinot <texitoi@texitoi.eu>"]
 name = "transit_model"
-version = "0.42.0"
+version = "0.42.1"
 license = "AGPL-3.0-only"
 description = "Transit data management"
 repository = "https://github.com/CanalTP/transit_model"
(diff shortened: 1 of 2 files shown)


woshilapin (commented):
Maybe this was intentional to keep the stop locations invisible from the autocompletion (this `visible` flag seems to be [only useful for autocompletion](https://github.com/CanalTP/ntfs-specification/blob/master/ntfs_fr.md#stopstxt-requis)). That said, I'd prefer to have it as an explicit enhancement process in `Model::new()` if this was the case, instead of where it is right now (`read/ntfs.rs`).

ArnaudOggy:
@patochectp Does this invisibility speak to you? Was it a choice or a mistake ? :thinking:

patochectp:
> @patochectp Does this invisibility speak to you? Was it a choice or a mistake ? thinking

I think like @woshilapin that was intentional to keep the stop locations invisible from the autocompletion

ArnaudOggy:
> I think like @woshilapin that was intentional to keep the stop locations invisible from the autocompletion

Ok, this is weird :/
The locations are stops with type >=3, therefore filterable by Mimir.
I'll close the PR and we'll stay that way. We'll reopen and fix it as needed.

PR #830, opened 2021-12-06
[tech] bump 0.43.0

datanel:
and revert versions of "sub-crates" that were bumped in the previous PR because it's useless...


ArnaudOggy (approved):

woshilapin (approved):
Maybe we can even put `0.0.0` everywhere to make it more obvious that we don't care about these versions.

PR #831, opened 2022-01-03
[fix] gtfs2ntfs: consider pickup / drop off "Must coordinate with driver to arrange pickup" as odt in ntfs

patochectp:
ref: DATENG-30


src/gtfs/mod.rs
```diff
@@ -551,3 +600,24 @@ pub fn write_to_zip<P: AsRef<std::path::Path>>(model: Model, path: P) -> Result<
     input_tmp_dir.close()?;
     Ok(())
 }
+
+/// gtfs pickup or drop off method "Must coordinate with driver to arrange pickup" (type 3)
+/// corresponds to pickup or drop off method "upon reservation only" in ntfs (type 2)
+pub fn convert_gtfs_pickup_drop_off_type_to_ntfs_type(collections: &mut Collections) {
```
woshilapin, line 606: What a long name. How about something a bit shorter like `convert_pickup_dropoff_to_ntfs` (since we already are in `gtfs` module anyway so no need to keep it and the word `type` is probably not that important?).

src/gtfs/mod.rs
```diff
@@ -551,3 +600,24 @@ pub fn write_to_zip<P: AsRef<std::path::Path>>(model: Model, path: P) -> Result<
...
+    let vj_idxs: Vec<Idx<VehicleJourney>> = collections
+        .vehicle_journeys
+        .iter()
+        .map(|(idx, _)| idx)
```
woshilapin, line 610: Maybe one day we should add a method `.keys()` to `CollectionWithId` :angel:

woshilapin: And here it is...
https://github.com/CanalTP/typed_index_collection/pull/15

woshilapin (approved):

src/gtfs/mod.rs
patochectp: At first I had implemented it in model and I wanted an explicit name. But you’re right it’s no longer necessary

src/gtfs/mod.rs
woshilapin: `typed_index_collection:2.2.0` has been released and now contains `CollectionWithId::indexes()` so you should be able to do the following.

```rust
    let vj_idxs: Vec<Idx<VehicleJourney>> = collections
        .vehicle_journeys
        .indexes()
        .collect();
```

patochectp: done in 5902c73400934ccf080e56d2a3b9b0dbb71b20e0

PR #836, opened 2022-01-12
[feature] gtfs : read as collections

patochectp:


src/gtfs/mod.rs
```diff
@@ -413,20 +424,58 @@ impl Reader {
             ))
         }
     }
+    /// Imports `Collections` from the
+    /// [GTFS](https://gtfs.org/reference/static).
+    /// files in the given directory.
+    /// This method will try to detect if the input is a zipped archive or not.
+    /// If the default file type mechanism is not enough, you can use
+    /// [Reader::parse_zip] or [Reader::parse_dir].
```
ArnaudOggy, line 432:
```suggestion
    /// [Reader::parse_zip_collections] or [Reader::parse_dir_collections].
```

ArnaudOggy (dismissed):

src/gtfs/mod.rs
```diff
@@ -413,20 +424,58 @@ impl Reader {
...
     /// Imports a `Model` from the [GTFS](https://gtfs.org/reference/static)
     /// files in the `path` directory.
     pub fn parse_dir(self, path: impl AsRef<Path>) -> Result<Model> {
+        let collections = self.parse_dir_collections(path)?;
+        Model::new(collections)
+    }
+
+    /// Imports `Collections` from the [GTFS](https://gtfs.org/reference/static)
+    /// files in the `path` directory.
+    fn parse_dir_collections(self, path: impl AsRef<Path>) -> Result<Collections> {
```
ArnaudOggy, line 468: We can push the idea further and put them in `pub`

patochectp: This is not necessary because `parse_collections` is `pub`

ArnaudOggy: Yes I saw. It's in case we needed it later. and to remain homogeneous.
But yes for the moment we can stay like that.

ArnaudOggy (approved):

ArnaudOggy (approved):

PR #838, opened 2022-01-17
[tech] make some items public

patochectp:


src/gtfs/mod.rs
```diff
@@ -552,6 +561,35 @@ struct Route {
...
+            .map(|(mut eq, id)| {
+                eq.id = id;
+                eq
+            })
+            .collect();
+
+        eqs.sort_by(|l, r| l.id.cmp(&r.id));
+        eqs
+    }
+    ///insert transit model equipment into EquipmentList
```
woshilapin, line 585:
```suggestion
    /// insert transit model equipment into EquipmentList
```

src/gtfs/mod.rs
```diff
@@ -31,10 +31,21 @@ use anyhow::{anyhow, Context};
...
+use std::{
+    collections::{BTreeMap, HashMap},
+    fmt,
+    path::Path,
+};
+
 use tracing::info;
 use typed_index_collection::CollectionWithId;
 
+#[cfg(feature = "gtfs")]
```
woshilapin, line 43: I would have made them like this because these functions are really specific to parsing.

```suggestion
#[cfg(all(feature = "gtfs", feature = "parser"))]
```

src/gtfs/mod.rs
```diff
@@ -552,6 +563,35 @@ struct Route {
     sort_order: Option<u32>,
 }
 
+#[derive(Default)]
+/// To associate a list of equipment with a stop
+pub struct EquipmentList {
```
woshilapin, line 568: I would also gate that `struct` behind `#[cfg(all(feature = "gtfs", feature = "parser"))]` (and the `impl` block below).

src/lib.rs
```diff
@@ -51,6 +51,10 @@ pub mod calendars;
 #[macro_use]
 pub mod objects;
 mod enhancers;
+#[cfg(not(feature = "parser"))]
+pub(crate) mod file_handler; // to keep backward compatibility, not exposing FileHandler
```
woshilapin, line 55: You probably can remove the comment, I did put a comment in the snippet I've sent on private message to explain why I was doing it.

src/read_utils.rs
```diff
@@ -94,118 +94,8 @@ pub fn read_config<P: AsRef<path::Path>>(
...
-    fn source_name(&self) -> &str {
-        self.archive_path
-            .to_str()
-            .unwrap_or_else(|| panic!("the path '{:?}' should be valid UTF-8", self.archive_path))
-    }
-}
-
 /// Read a vector of objects from a zip in a file_handler
-pub(crate) fn read_objects<H, O>(
+pub fn read_objects<H, O>(
```
woshilapin, line 98: This should be featured-gated with `#[cfg(feature = "parser")]`.

src/read_utils.rs
```diff
@@ -278,11 +168,8 @@ where
     }
 }
 
-/// Read a CollectionId from a zip in a file_handler
-pub(crate) fn read_collection<H, O>(
-    file_handler: &mut H,
-    file_name: &str,
-) -> Result<CollectionWithId<O>>
+/// Read a CollectionId from a required file in a file_handler
+pub fn read_collection<H, O>(file_handler: &mut H, file_name: &str) -> Result<CollectionWithId<O>>
```
woshilapin, line 172: This should be featured-gated with `#[cfg(feature = "parser")]`.

src/read_utils.rs
```diff
@@ -291,7 +178,8 @@ where
     CollectionWithId::new(vec).map_err(|e| anyhow!("{}", e))
 }
 
-pub(crate) fn read_opt_collection<H, O>(
+/// Read a CollectionId from a optional file in a file_handler
+pub fn read_opt_collection<H, O>(
```
woshilapin, line 182: This should be featured-gated with `#[cfg(feature = "parser")]`.

src/lib.rs
patochectp: I thought it was interesting, but I’m going to remove it

src/gtfs/mod.rs
patochectp: I put the `EquipmentList` back in `read.rs`

src/parser/mod.rs
```diff
@@ -0,0 +1,116 @@
+// Copyright (C) 2017 Kisio Digital and/or its affiliates.
```
woshilapin, line 1: minor: You could have named the file `src/parser.rs` instead of `src/parser/mod.rs`?

src/read_utils.rs
```diff
@@ -17,15 +17,11 @@ use crate::{
     objects::{self, Contributor},
```
woshilapin, line 17: Since this file now only contains things about the configuration, how about renaming it `config.rs` or `configuration.rs`?

woshilapin (dismissed):

woshilapin:
I was thinking it would be nice to add some documentation about the 2 new introduced features into the preamble of `src/lib.rs` (see how it's already done for the existing features).

patochectp:
> I was thinking it would be nice to add some documentation about the 2 new introduced features into the preamble of `src/lib.rs` (see how it's already done for the existing features).

I told myself the same thing

src/lib.rs
```diff
@@ -38,6 +38,13 @@
 //! mutate a `Model`. It might not be completely stable at the moment so use
 //! with care (or not at all!).
 //!
+//! ## `gtfs`
+//! This feature is only used to expose some gtfs functions for use in external projects
```
woshilapin, line 42: I would mention that the exposed API are not stable and that using this is experimental. I'm not expecting us to be as careful about the stability here that for the rest of the exposed API. What do you think?

patochectp: done in 6f11bef5115732b36972ed4fe3f9e8477c629b99

woshilapin (approved):

woshilapin (approved):

PR #843, opened 2022-02-08
[feature] check coordinate validity

patochectp:
Ref. https://navitia.atlassian.net/browse/DATENG-34


src/model.rs
```diff
@@ -395,7 +395,7 @@ impl Collections {
                 })
                 .collect(),
         )?;
-        let stop_areas = self
+        let mut stop_areas = self
```
ArnaudOggy, line 398: And struct `StopLocation` too

ArnaudOggy (approved):

PR #845, opened 2022-03-11
[tech] Upgrade packages

datanel:


Cargo.toml
```diff
@@ -54,7 +54,7 @@ minidom_ext = "1"
 minidom_writer = "1"
 num-traits = "0.2"
 pretty_assertions = "1"
-proj = { version = "0.24", optional = true } # libproj version used by 'proj' crate must be propagated to CI and makefile
+proj = { version = "0.25", optional = true } # libproj version used by 'proj' crate must be propagated to CI and makefile
```
woshilapin, line 57: Does it involve a change of Proj C library version?

src/netex_france/exporter.rs
```diff
@@ -175,7 +175,7 @@ impl<'a> Exporter<'a> {
         let from = "+proj=longlat +datum=WGS84 +no_defs"; // https://epsg.io/4326
         let to = "+proj=lcc +lat_1=49 +lat_2=44 +lat_0=46.5 +lon_0=3 +x_0=700000 +y_0=6600000 +ellps=GRS80 +towgs84=0,0,0,0,0,0,0 +units=m +no_defs"; // https://epsg.io/2154
         Proj::new_known_crs(from, to, None)
-            .ok_or_else(|| anyhow!("Proj cannot build a converter from '{}' to '{}'", from, to))
+            .map_err(|_| anyhow!("Proj cannot build a converter from '{}' to '{}'", from, to))
```
woshilapin, line 178: I'm wondering if we should reuse original error too? The [`anyhow!()` macro](https://docs.rs/anyhow/latest/anyhow/macro.anyhow.html) can take an `Error` as argument too. But it doesn't seem possible to customize the message and keep the source error with the macro... But I'm thinking maybe `.with_context()` should be used instead?

src/ntfs/read.rs
```diff
@@ -295,7 +295,7 @@ where
             |v| v != 0,
         );
 
-        let precision = stop_time.precision.or_else(|| {
+        let precision = stop_time.precision.or({
```
woshilapin, line 298: I'm surprised by this change since there is no reason to evaluate the `if` when `stop_time.precision` is `Some`. Why not keeping it lazy?

src/objects.rs
```diff
@@ -796,7 +796,7 @@ impl FromStr for Time {
         };
         let hours: u32 = hours.parse()?;
         let minutes: u32 = minutes.parse()?;
-        let last_second_number = seconds.find('.').unwrap_or_else(|| seconds.len());
+        let last_second_number = seconds.find('.').unwrap_or(seconds.len());
```
woshilapin, line 799: Same here, why not keeping the evaluation of `.len()` lazy?

src/ntfs/read.rs
datanel: ask for CLippy https://rust-lang.github.io/rust-clippy/master/index.html#unnecessary_lazy_evaluations

I think it's not necessary for simple types

Cargo.toml
datanel: It works with proj 8.1 so I suppose we don't need to change Proj C

src/ntfs/read.rs
woshilapin: Weird. Especially since this is not a simple type because there is an evaluation (an `if` in one case, a function in the other one). But hey, if `clippy` say so, since our CI depends on it, fine by me!

datanel: wait for the pull request of `tartare-tools` before being shocked

Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 authors = ["Kisio Digital <team.coretools@kisio.com>", "Guillaume Pinot <texitoi@texitoi.eu>"]
 name = "transit_model"
-version = "0.47.0"
+version = "0.48.0"
```
ArnaudOggy, line 4: Need rebase ? We are already in 0.48.0

datanel: yes thanks

src/netex_france/exporter.rs
datanel: ### test with

```rust
let from = "BAD"; // https://epsg.io/4326
let to = "+proj=lcc +lat_1=49 +lat_2=44 +lat_0=46.5 +lon_0=3 +x_0=700000 +y_0=6600000 
```

### 1. Custom message

```rust
Proj::new_known_crs(from, to, None)
    .map_err(|_| anyhow!("Proj cannot build a converter from '{}' to '{}'", from, to))
```

will display

```
Proj cannot build a converter from 'BAD' to '+proj=lcc +lat_1=49 +lat_2=44 +lat_0=46.5 +lon_0=3 +x_0=700000 +y_0=6600000 +ellps=GRS80 +towgs84=0,0,0,0,0,0,0 +units=m +no_defs'
```

### 2. Original Error

```rust
Proj::new_known_crs(from, to, None).map_err(|err| anyhow!(err))
```

will display

```
The underlying PROJ call failed: Unknown error (code 4096)
```

### 3. with context

```rust
use anyhow::Context;
Proj::new_known_crs(from, to, None)
            .with_context(|| format!("Proj cannot build a converter from '{}' to '{}'", from, to))
```

will display 

```
Proj cannot build a converter from 'BAD' to '+proj=lcc +lat_1=49 +lat_2=44 +lat_0=46.5 +lon_0=3 +x_0=700000 +y_0=6600000 +ellps=GRS80 +towgs84=0,0,0,0,0,0,0 +units=m +no_defs'

Caused by:
    The underlying PROJ call failed: Unknown error (code 4096)', 
```

We know that it will never fail and the error message will never be displayed because `from` and `to` are known and never passed as arguments.

for me

1. is ok but the message is not necessary because never displayed
2. is ok, the message says nothing but we don't care
3. 👎🏻 context does not help

woshilapin: Thanks for testing and indeed, the underlying message is useless. So fine with solution 1 as you did.

woshilapin (approved):

ArnaudOggy (dismissed):

ArnaudOggy (approved):

PR #846, opened 2022-03-11
feat: add memory optimization for 'Model'

woshilapin:
Trying to optimize the memory imprint of the `Model` by shrinking down the `Vec<StopTime>` in `VehicleJourney`.

To improve a bit loading performance too, we pre-allocate `Vec<StopTime>` (with an arbitrary value of 50 (we might improve this value later if needed) in 2 places:
- creation of `VehicleJourney` with `Default` trait
- creation of `VehicleJourney` when reading GTFS (first, `VehicleJourney` are created, then they are populated with `StopTime`)

ref: NAV-1050


src/enhancers/memory_shrink.rs
```diff
@@ -0,0 +1,16 @@
...
+    collections.stop_time_ids.shrink_to_fit();
+    collections.stop_time_comments.shrink_to_fit();
+    let vj_idxs: Vec<_> = collections.vehicle_journeys.indexes().collect();
+    for vj_idx in vj_idxs {
+        collections
+            .vehicle_journeys
+            .index_mut(vj_idx)
+            .stop_times
+            .shrink_to_fit();
+    }
```
pbench, line 15:
```suggestion
    for vehicle_journey in collections.vehicle_journeys.values_mut() {
        vehicle_journey
            .stop_times
            .shrink_to_fit();
    }
```
to avoid the unecesseray allocation of the vj_idxs vector

woshilapin: `.values_mut()` doesn't exist on `CollectionWithId` (but it does exist on `Collection`). It would need to implement `DerefMut` but I believe there is inherent difficulty to actually support that (I tried once :smile:).

For the record, this is the error with the code you're proposing.
```
error[E0596]: cannot borrow data in a dereference of `typed_index_collection::CollectionWithId<objects::VehicleJourney>` as mutable
  --> src/enhancers/memory_shrink.rs:16:15
   |
16 |     for vj in collections.vehicle_journeys.values_mut() {
   |               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ cannot borrow as mutable
   |
   = help: trait `DerefMut` is required to modify through a dereference, but it is not implemented for `typed_index_collection::CollectionWithId<objects::VehicleJourney>`
```

pbench: > It would need to implement DerefMut but I believe there is inherent difficulty to actually support that (I tried once smile).

I am now curious about what is the difficulty ;)

Instead of implementing DerefMut and delegating to the inner collection, is it easier to just impl the function values_mut() on CollectionWithId ?

ArnaudOggy (approved):

src/enhancers/memory_shrink.rs
woshilapin: So, I improved a bit by removing the allocation in 2df97f5d.

Issue #854, opened 2022-06-29
ODT : incorrect behavior ?


christopheblin:
I try the following commands and I expect that gtfs2ntfs would create a comments.txt and a comment_links.txt for ODT as indicated in the docs

To summarize in a human friendly way what I did
- launch latest version (0.51.0) of transit_model through docker
- execute gtfs2ntfs
- list files of the output directory to check that commetns.txt and comment_links.txt are missing
- display the content of the -c option (you can see that both odt:true and odt_comment are set)
- display an extract of the content of stop_times.txt (header + at least one trip where pickup_type and drop_off_type are set to 2)

Am I doing something wrong or is there a bug ?

Dont hesitate to ask more details if needed

**Some details**

https://github.com/hove-io/transit_model/blob/master/documentation/gtfs_to_ntfs_specs.md#prepending-data  

> Two parameters can be specified in the configuration of the converter in order to determine if on demand transport (ODT) data should be considered when reading the input GTFS (in particular, when [reading the stop_times.txt file](https://github.com/hove-io/transit_model/blob/master/documentation/gtfs_to_ntfs_specs.md#reading-stop_timestxt)):
> - a boolean parameter odt, by default set to false, indicating if the GTFS should be considered as containing ODT information
> - a string odt_comment setting the message associated to an ODT comment.

https://github.com/hove-io/transit_model/blob/master/documentation/gtfs_to_ntfs_specs.md#reading-stop_timestxt

| NTFS file | NTFS field | Constraint | Value/Note |
|-----------|------------|------------|------------|
| comments.txt | comment_type | Optional | on_demand_transport                            |
| comments.txt | comment_name | Required | The message set for the parameter odt_comment. |

**Commands**

```
docker run --rm -it -v $PWD:/data --entrypoint /bin/sh navitia/transit_model:v0.51.0
# gtfs2ntfs -i /data/gtfs -c /data/navitia.json -p vss -o /data/ntfs
2022-06-29T07:31:20.033793Z  INFO gtfs2ntfs: Launching gtfs2ntfs...
2022-06-29T07:31:20.033837Z  INFO transit_model::configuration: Reading dataset and contributor from "/data/navitia.json"
2022-06-29T07:31:20.038736Z  INFO transit_model::parser: Reading calendar.txt
2022-06-29T07:31:20.041757Z  INFO transit_model::parser: Reading calendar_dates.txt
2022-06-29T07:31:20.044944Z  INFO transit_model::parser: Reading agency.txt
[...]

PR #855, opened 2022-06-29
docs: improve documentation about ODT parameters

woshilapin:
Follow-up on #854.

image


ArnaudOggy (approved):

prhod (approved):

christopheblin:
@woshilapin As you ask my opinion on the ticket, I provide it here 😄 

1. I think that `read-as-line` has the same ambiguity
2. "Two parameters can be specified as parameters of the converter ..." is not very clear to me, I would suggest "Two parameters can be specified as CLI arguments of the converter ..."

woshilapin:
Thanks for the useful feedback @christopheblin. I made the changes you proposed.

ArnaudOggy (approved):

prhod (approved):

PR #876, opened 2022-12-01
[tech] Github Actions runners - Simplification on release creation and push

ArnaudOggy:
Better to test on release creation, because we don''t have `Cargo.lock`,
and therefore potential crate updates since the last merge master


woshilapin (approved):
Maybe we could add the `Cargo.lock` in the repository, it'd simplify things?

datanel:
We don't upgrade major version it shouldn't be a problem if we don't test at release.
but if it is a problem +1 for `Cargo.lock`

(we should also use workspaces like `tartare-tools`)

PR #883, opened 2023-01-04
feat: remove 'mutable-model' feature

woshilapin:
This feature, that was introduced for an experimentation, didn't prove itself safe and constrained enough. Let's remove it.

Original PR was #761. For details, opening this door of mutability was opening an opportunity to make the `Model` inconsistent. The consequence of that would be disastrous, since relations in the graph might become incorrect, the information incoherent and raise subtle bugs that might be hard to debug and fix.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 authors = ["Hove <team.coretools@kisio.com>", "Guillaume Pinot <texitoi@texitoi.eu>"]
 name = "transit_model"
-version = "0.53.0"
+version = "0.54.0"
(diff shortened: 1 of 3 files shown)


ArnaudOggy (approved):

pbougue (approved):
OK to remove it, I think a bit more sharing on why it was added (maybe just a link to first PR) and more interesting why it's not worth keeping it would be nice.

woshilapin:
I improved a bit the description. But to be honest, I vaguely remember finding an example of using this API and breaking the `Model` but it was a long time ago.

PR #884, opened 2023-01-05
feat: add 'occupancies.txt' support (ntfs:v0.13.0) 

woshilapin:
Implementing support for the new version of NTFS (not released yet, see https://github.com/hove-io/ntfs-specification/pull/38).

`sanitize` is not ideal because it could go a step further by removing `Occupancy` when:
- none of the `vehicle_journey` of the line is traveling on the interval `from_date`-`to_date`
- `from_stop_area` is **after** `to_stop_area` on all `vehicle_journey` of the line


src/model.rs
```diff
@@ -553,6 +554,24 @@ impl Collections {
...
+                    "Occupancy with 'line_id={}' has been removed because stop area is not used",
+                    occupancy.line_id
+                );
+                return false;
+            }
+            if !stop_area_ids_used.contains(&occupancy.from_stop_area) {
+                debug!("Occupancy with 'from_stop_area={}' has been removed because stop area is not used", occupancy.from_stop_area);
+                return false;
+            }
+            if ! stop_area_ids_used.contains(&occupancy.to_stop_area){
```
ArnaudOggy, line 569:
```suggestion
            if !stop_area_ids_used.contains(&occupancy.to_stop_area){
```

pbench (approved):

src/model.rs
woshilapin: Thank you. I do not understand why my editor seems to fail reformating on this specific file, only when adding these lines :shrug: :thinking:

src/objects.rs
```diff
@@ -1900,6 +1900,77 @@ impl AddPrefix for Address {
...
+    pub saturday: Option<bool>,
+    #[serde(
+        deserialize_with = "de_opt_bool_from_str",
+        serialize_with = "ser_from_opt_bool"
+    )]
+    pub sunday: Option<bool>,
+    pub occupancy: OccupancyStatus,
+}
+impl_id!(Occupancy, Line, line_id);
+
```
ArnaudOggy, line 1973: to add, especially for the merge later
```
impl AddPrefix for Occupancy {
    fn prefix(&mut self, prefix_conf: &PrefixConfiguration) {
        self.line_id = prefix_conf.referential_prefix(self.line_id.as_str());
        self.from_stop_area = prefix_conf.referential_prefix(self.from_stop_area.as_str());
        self.to_stop_area = prefix_conf.referential_prefix(self.to_stop_area.as_str());
    }
}
```

src/ntfs/read.rs
```diff
@@ -668,6 +668,19 @@ where
...
+    collections: &mut Collections,
+    file_handler: &mut H,
+) -> Result<()>
+where
+    for<'a> &'a mut H: FileHandler,
+{
+    let file = "occupancies.txt";
+    let occupancies = read_objects_loose::<_, Occupancy>(file_handler, file, false)?;
+    collections.occupancies = Collection::new(occupancies);
+    Ok(())
```
ArnaudOggy, line 681: It might be relevant to check the memory build like here :thinking: 
https://github.com/hove-io/transit_model/blob/a3649520def97797ee3fff92a572aab72ddc6b90/src/ntfs/read.rs#L626
on 3 main fields (line_id, from_stop_area, to_stop_area)

tests/fixtures/transfers/mono_contributor/input/feed_infos.txt
```diff
@@ -1,5 +1,5 @@
 feed_info_param,feed_info_value
-ntfs_version,0.3
+ntfs_version,0.13.0
```
ArnaudOggy, line 2: Like some others, it's input. 
Normally you didn't need to change it. 
Especially since this ntfs may really be in version 0.10.0.

woshilapin: Yes, I realized that, but I'm wondering if we're really testing something with this? If we're not, let's put everything to the same value, easier to maintain (I actually changed all the values at once with a `sed`), what do you think?

src/ntfs/read.rs
woshilapin: I actually started doing exactly what you have in mind, but I realized that we're going to do that on `sanitize` anyway and that it might be better to do it only once. No strong opinion on either solution. Do you prefer me to do it on `manage_occupancies` too?

tests/fixtures/transfers/mono_contributor/input/feed_infos.txt
ArnaudOggy: I would have preferred to leave it as before.
Maybe one day we will have an old version ntfs (with breaking change) in input to guarantee backward compatibility.
But I don't have a definite opinion. We'll see what others say.

src/objects.rs
woshilapin: Good catch. Added in f1c5e4db. I'm wondering why this has not failed any test? Should we add a test doing the following:
- read a NTFS
- add prefix
- compare to an output fixture

src/ntfs/read.rs
ArnaudOggy: If it's not too long, do it. It seems more elegant to me to clean beforehand.
Otherwise let it go.

src/objects.rs
ArnaudOggy: It's certainly historic. These tests are in TartareTools now

tests/fixtures/transfers/mono_contributor/input/feed_infos.txt
woshilapin: I've rolled back all the `input` fixtures to previous NTFS versions in f716e462.

ArnaudOggy (approved):

src/model.rs
```diff
@@ -553,6 +554,30 @@ impl Collections {
             .retain(|address| addresses_used.contains(&address.id));
         self.admin_stations
             .retain(|admin_station| stop_area_ids_used.contains(&admin_station.stop_id));
+        self.occupancies.retain(|occupancy| {
+            if !line_ids_used.contains(&occupancy.line_id) {
+                debug!(
+                    "Occupancy with 'line_id={}' has been removed because stop area is not used",
```
patochectp, line 560:
```suggestion
                    "Occupancy with 'line_id={}' has been removed because line is not used",
```

woshilapin: Fixed in 715b88da. Thanks.

ArnaudOggy (approved):

PR #920, opened 2023-09-20
[tech] Fix opening/closing times on frequency lines

ArnaudOggy:


woshilapin (approved):
Covered by tests, looks good to me. [Here](https://github.com/hove-io/transit_model/pull/921) is some propositions, maybe not everything should be used, not sure of the maintenability of some of what I propose.

woshilapin (approved):

PR #936, opened 2024-01-15
[feature] Add model builder into transit_model

datanel:
1. Commit 1/2: move `ModelBuilder` from `model-builder` crate to `transit_model`. It will easier to maintain and to use from another crate
2. Commit 3: add more functions


woshilapin (approved):
Nice rework, and it's going to be much easier to maintain it I believe... and maybe use it somewhere else 🤷 😄 Thanks for the nice commit splitting, really helped to review the PR.

ArnaudOggy (approved):

PR #942, opened 2024-03-07
impl Display for CommentType

datanel:


woshilapin (approved):
Maybe [`strum`](https://crates.io/crates/strum) can be used? Only for one use, it would probably be too much, but there might be more of these to replace 🤷

datanel:
> Maybe [`strum`](https://crates.io/crates/strum) can be used? Only for one use, it would probably be too much, but there might be more of these to replace 🤷

That's what I saw too but it thought it was overkill

patochectp (dismissed):

patochectp (approved):

PR #953, opened 2024-07-23
[NAV-3235] Feature: odt reservation

datanel:
https://navitia.atlassian.net/browse/NAV-3235


ArnaudOggy:
@datanel 
Can you add this in model ?
```
impl ODTReservation {
    pub fn is_similar(&self, other: &Self) -> bool {
        self.url == other.url && 
        self.phone == other.phone && 
        self.conditions == other.conditions && 
        self.deeplink == other.deeplink
    }
}
```
We will need it, and it will avoid sanitize deduplication

datanel:
> @datanel Can you add this in model ?
> 
> ```
> impl ODTReservation {
>     pub fn is_similar(&self, other: &Self) -> bool {
>         self.url == other.url && 
>         self.phone == other.phone && 
>         self.conditions == other.conditions && 
>         self.deeplink == other.deeplink
>     }
> }
> ```
> 
> We will need it, and it will avoid sanitize deduplication

done

ArnaudOggy (approved):

PR #1031, opened 2026-06-11
refactor: DHUB-2583 ntfs2netexfr parallelize offer file generation

datanel:
## Performance Optimizations for NeTEx France Export

This PR significantly improves the performance of `ntfs2netexfr` when exporting large datasets (tested on a file with 72 networks and 2,715 offer files, previously taking ~20 minutes).

#### 1. Fixed `OfferExporter` recreation bug
- **Before**: `OfferExporter::new()` was called once per network (72 times), recalculating expensive `route_points` and `line_modes` on the entire model each time.
- **After**: Instantiated once globally in `write_offers()` and reused across all networks.
- **Impact**: Eliminates redundant O(N) computations where N is the total number of routes/lines.

#### 2. Added `BufWriter` for all file writes
- Wrapped all `File::create()` calls with `BufWriter::new()` in `write_lines`, `write_stops`, `write_calendars`, `write_transfers`, and `write_network_offers`.
- **Impact**: Reduces syscall overhead by buffering writes.

#### 3. Pre-computed stop point coordinates (enables parallelism)
- **Problem**: `Proj` (the coordinate converter) contains FFI raw pointers and is not `Send+Sync`, preventing parallel access.
- **Solution**: Pre-compute all stop point coordinates to EPSG:2154 format during `OfferExporter::new()`, store in a `HashMap<Idx<StopPoint>, Option<String>>`, then drop the `Proj` instance.
- **Impact**:
  - Makes `OfferExporter` fully `Send+Sync` (required for parallelism)
  - Avoids repeated coordinate conversions (each stop point is converted once, even if used in thousands of vehicle journeys)
  - Removes FFI dependency from the hot path

#### 4. Parallelized offer file generation with Rayon
- Refactored `write_offers()` into two phases:
  - **Phase 1 (sequential)**: Create network directories and collect all work items as `(network_path, line_idx)` pairs.
  - **Phase 2 (parallel)**: Use `rayon::par_iter().try_for_each()` to generate and write each `offre_*.xml` file independently across all CPU cores.
- Added `rayon = "1"` as an optional dependency, enabled via the `proj` feature.
- **Impact**: Theoretical ~8× speedup on 8-core machines for the offer writing phase (dominant cost at ~20 minutes → expected ~2-3 minutes).

### Technical Details

- All file writes are now independent and can run in parallel without locks.
- Error handling in the parallel section uses `Result<()>` with `try_for_each`, aborting on the first error.
- The pre-computation of coordinates also reduces memory allocations during the parallel phase.

### Expected Performance Improvement

On a machine with 12 CPU cores, the offer export phase (previously ~20 minutes) completes in approximately **30 seconds**

### DIFF
--- a/src/netex_france/exporter.rs
+++ b/src/netex_france/exporter.rs
@@ -20,19 +20,21 @@ use crate::{
         CalendarExporter, CompanyExporter, LineExporter, NetworkExporter, OfferExporter,
         StopExporter, TransferExporter,
     },
-    objects::{Date, Line, Network},
+    objects::{Date, Line},
     Result,
 };
 use anyhow::anyhow;
 use chrono::prelude::*;
 use proj::Proj;
+use rayon::prelude::*;
 use relational_types::IdxSet;
 use std::{
     convert::AsRef,
     fmt::{self, Display, Formatter},
     fs::{self, File},
+    io::BufWriter,
     iter,
-    path::Path,
+    path::{Path, PathBuf},
 };
 use tracing::info;
 use typed_index_collection::Idx;
@@ -260,7 +262,7 @@ impl Exporter<'_> {
         P: AsRef<Path>,
     {
         let filepath = path.as_ref().join(NETEX_FRANCE_LINES_FILENAME);
-        let file = File::create(&filepath)?;
+        let file = BufWriter::new(File::create(&filepath)?);
         let network_frames = self.create_networks_frames();
         let lines_frame = self.create_lines_frame()?;
         let companies_frame = self.create_companies_frame();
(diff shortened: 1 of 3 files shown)


src/netex_france/exporter.rs
```diff
@@ -447,46 +449,57 @@ impl Exporter<'_> {
...
+                let line_indexes: IdxSet<Line> = self.model.get_corresponding_from_idx(network_idx);
+                Ok(line_indexes
+                    .into_iter()
+                    .map(|line_idx| (network_path.clone(), line_idx))
+                    .collect())
+            })
+            .collect::<Result<Vec<_>>>()?
+            .into_iter()
+            .flatten()
+            .collect();
```
woshilapin: If I’m not wrong, looks like a case for https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.flatten_ok to avoid collecting twice.

datanel: yes you are right but finally I did it without external crate

ArnaudOggy (approved):

